PR #4, opened 2015-12-19
fix issue #1 read http not failing on bad status code

### LINKED ISSUE #1: http read adapter not throw correct status code error
When running:
```
rlgomes@x230> ./bin/juttle -e "read http -url 'https://api.github.com/repositories?access_token=845c03c47b0398900de1431b9dd27738f01c0025' "
Error: internal error CompileError: Error: Invalid format option value, must be one of the following: csv, json, jsonl { [RuntimeError: Error: internal error CompileError: Error: Invalid format option value, must be one of the following: csv, json, jsonl]
...
```
the expected failure would be to indicate the underlying 403 from github indicating that we forget to set the User-Agent but the current implementation is failing to do this before it tries to process the HTTP response.

rlgomes:
fixes #1 by just ripping out request-promise for now and using the
request library directly so we don't end up mixing stream events and
promise chains and having timing issues.

### DIFF
--- a/lib/adapters/http/read.js
+++ b/lib/adapters/http/read.js
@@ -1,8 +1,8 @@
 var _ = require('underscore');
 var contentType = require('content-type');
-var request = require('request-promise');
 var Juttle = require('../../runtime').Juttle;
 var parsers = require('../parsers');
+var request = require('request');
 
 var Read = Juttle.proc.source.extend({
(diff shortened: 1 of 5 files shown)


rlgomes:
@davidvgalbraith @demmer

davidvgalbraith:
+1

PR #5, opened 2015-12-19
fix travis notifications

rlgomes:
when we opened up the juttle repo we stopped receiving slack
notifications from travis this should fix that


mstemm:
+1

PR #8, opened 2015-12-21
Drop the "code " from the error string.

mstemm:
For error strings like:

<input>:1:17:
   1:read from gmail -from :1 hour ago: | view table
                     ^^^^^
Options must be specified before filter expressions. code (JUTTLE-SYNTAX-ERROR-WITHOUT-EXPECTED)

the "code " is not needed, so remove it.

@dave @mnibecker


davidvgalbraith:
+1

PR #10, opened 2015-12-21
Use more specific error for unknown views.

mstemm:
Use a specific error RT-INVALID-VIEW when a program refers to a view
that isn't supported by the runtime (i.e., for the cli, by the client
sink manager).

@demmer


demmer:
LGTM.

Issue #6, opened 2015-12-19
file adapter runs out of memory on a very large write


rlgomes:
Here's a simple scenario that leads to the OOM:

```
./bin/juttle -e "read stochastic -source 'logs' -from :2014-01-01: -nhosts 100 -lpm 1000 | write file -file 'test.json'"
```

From a quick look at the code its waiting to accumulate all of the data points before writing them out to the underlying file instead of writing out each point or batches of points in order to avoid running out of memory.

rlgomes:
@demmer may interest you.

PR #11, opened 2015-12-21
Fix file adapter to handle large files

rlgomes:
this PR reworks the file adapter so its easier to add more features in
the future by splitting the read and write up into individual modules.

the fix for issue #6 is currently to set a write limit so that we
simply limit writing out no more than 100K JSON points in the currently
supported writing format of JSON.

future work could be to include streamable writing formats such as CSV,
or JSONL (jsonlines: http://jsonlines.org/)

### DIFF
--- a/lib/adapters/file/write.js
+++ b/lib/adapters/file/write.js
@@ -1,11 +1,12 @@
 var _ = require('underscore');
 var Juttle = require('../../runtime/index').Juttle;
-var fs = require('fs');
+var Promise = require('bluebird');
+var fs = Promise.promisifyAll(require('fs'));
 
 var Write = Juttle.proc.sink.extend({
     procName: 'write-file',
     initialize: function(options, params) {
-        var allowed_options = ['file'];
+        var allowed_options = ['file', 'bufferLimit', 'maxFilesize', 'flushFrequency'];
         var unknown = _.difference(_.keys(options), allowed_options);
         if (unknown.length > 0) {
             throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
@@ -21,6 +22,29 @@ var Write = Juttle.proc.sink.extend({
             });
         }
 
+        this.bufferLimit = options.bufferLimit || 100000;
+        if (!_.isNumber(this.bufferLimit) || this.bufferLimit < 0 ) {
+            throw this.compile_error('RT-OPTION-SHOULD-BE-POSITIVE-INTEGER', {
+                option: "bufferLimit"
+            });
+        }
+
+        // 200MB bufferLimit on file
+        this.maxFilesize = options.maxFilesize || 200*1024*1024;
+        if (!_.isNumber(this.maxFilesize) || this.maxFilesize < 0 ) {
+            throw this.compile_error('RT-OPTION-SHOULD-BE-POSITIVE-INTEGER', {
+                option: "maxFilesize"
+            });
+        }
+
+        // flush every 100 points
+        this.flushFrequency = options.flushFrequency ? options.flushFrequency : 100;
+        if (!_.isNumber(this.flushFrequency) || this.flushFrequency < 0 ) {
+            throw this.compile_error('RT-OPTION-SHOULD-BE-POSITIVE-INTEGER', {
+                option: "flushFrequency"
+            });
+        }
+
         this.filename = options.file;
         this.queue = [];
     },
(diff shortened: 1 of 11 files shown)


rlgomes:
@dmehra @demmer @dmajda @mstemm fixing the situation identified by issue #6 and I'll update the documentation here in this PR once the initial review for the change goes through. 

I'm thinking that if we set the limit for the only existing format for write to be 100K right now then what we can do in a follow up PR is to add the ability to `write file` to specify an output format such as csv and jsonlines which would allow for incremental writing and never run out of memory.

rlgomes:
@demmer initial approach of using the scheduler, doing this on a 1s interval seems a little "harsh" so I'm open to better suggestion but if we do things with a larger interval than the program would have to wait up to N additional seconds after finishing to actually complete and have the data flushed.

dmajda:
@rlgomes General note — in adapter-related PRs please tag @bkutil instead of (or in addition to) me. He is working more in the area so he should be able to give better feedback. Thanks.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        if (!_.has(options, 'file')) {
+            throw this.compile_error('RT-MISSING-OPTION-ERROR', {
+                proc: 'write file',
+                option: "file"
+            });
+        }
+
+        this.limit = options.limit ? options.limit : 100000;
```
bkutil: It is also possible to use `this.limit = options.limit || 100000`, which is somewhat shorter.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+            data = JSON.stringify(points, null, 4);
+            fs.writeFileSync(this.filename, data);
+            this.queue = [];
+        }
+
+        if (this.isdone) {
+            this.done();
+        } else {
+            var self = this;
+            this.program.scheduler.schedule(new Date().getTime() + 1000, function() {
```
bkutil: `Date.now()` is equivalent.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+
+    flush: function() {
+        if (this.queue.length !== 0) {
+            var data = [];
+            var points = [];
+
+            try {
+                data = fs.readFileSync(this.filename, 'utf8');
+                points = JSON.parse(data);
```
bkutil: A thing to keep in mind (we don't have to fix it now) is that this introduces a potential race condition - other file sink / other process could do a write while we keep the points in memory, and we'd then overwrite previous contents.

lib/adapters/file/write.js
bkutil: Were missing a check that limit should be non-negative.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+            } catch (err) {
+                if (err.code !== 'ENOENT') {
+                    this.logger.error('error reading file:', err.toString());
+                }
+            }
+
+            var space_left = this.limit - points.length;
```
bkutil: There's an edge case here - if there were already some points in the file, say 200k (e.g. because we're appending points to an existing file), and limit is set to 100k, space_left will be negative. This will cause the last 100k points to be added again. `space_left` minimum should be capped at 0.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+                // should stop the whole pipeline
+                this.trigger('error', this.runtime_error('RT-PROC-LIMIT-EXCEEDED', {
+                    proc: 'write file'
+                }));
+            } else {
+                // lets write out some more points
+                points = points.concat(this.queue);
+            }
+            data = JSON.stringify(points, null, 4);
+            fs.writeFileSync(this.filename, data);
```
bkutil: I know its a pain, but could we use async fs ops here? It won't block the event loop while the (possibly large) amount of data is written.

lib/runtime/procs/sort.js
```diff
@@ -58,8 +58,8 @@ var sort = fanin.extend({
         var k, row;
         var space_left = this.limit - this.n_pts;
         if (space_left < points.length) {
```
bkutil: I see there are similar issues w/ `space_left` and `limit` in the `sort` proc - I'll take care of them.

bkutil:
> the fix for issue #6 is currently to set a write limit so that we
simply limit writing out no more than 100K

Nice! So no more OOMs? IIUC, we still read & parse the whole file...

> future work could be to include streamable writing formats such as CSV,
or JSONL (jsonlines: http://jsonlines.org/)

+10k - I agree with you that this is the way to go.

lib/adapters/file/write.js
rlgomes: I'll add a unit test for this specific scenario.

rlgomes:
@bkutil taking care of the various suggestions all good catches. The issue with OOM'ing because we'd have an existing 1GB JSON file I'm not sure how we'd fix that at this point. I think another simple limitation would be that we'd fail to open files biger than 200MB ? impose some hard limit on the size of the file ?  @demmer @dmajda what do you think of this ? I guess we'd also have to expose some way of overriding the limitation but at least this way we could avoid blowing things up in a nasty way.

lib/adapters/file/write.js
rlgomes: I'll also see about making the read async as yes this would benefit not taking over the event loop while reading/writing large files.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,85 @@
...
+            var self = this;
+            this.program.scheduler.schedule(new Date().getTime() + 1000, function() {
+                self.flush();
+            });
+        }
+    },
+
+    eof: function() {
+        this.logger.debug('eof');
+        this.isdone = true;
```
demmer: It seems better to me to have `eof` trigger an immediate flush instead of just setting `isdone` and waiting.

This should be trivial to handle now since all the fs operations are async, but as @bkutil pointed out, they should be changed to be async, so you'll need to make sure you don't race between the scheduler callback and the immediate eof from flush. I think it's easy to handle by just setting `isdone` immediately after calling flush so that you know a subsequent callback won't race.

demmer:
As far as the limit on the size of the file, I guess you'll need to have two options: `-limit` which is a number of points and `-maxSize` in bytes. Seems icky but I don't see a better option.

rlgomes:
But I'll document that maxSize is the file size while -limit is the number of points in said file.

rlgomes:
@demmer @bkutil have a looksie I think this is much better at this point.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    flush: function(final) {
+        var self = this;
+
+        if (this.isdone) {
+            return;
+        }
+
+        if (final) {
```
demmer: Looks like this won't ever call `done()` if there aren't any points buffered in the queue when the EOF arrives.

You should fix this by always entering the promise chain even if the queue length is empty and have a test case for it.

lib/adapters/file/write.js
```diff
@@ -0,0 +1,135 @@
...
+            this.isdone = true;
+        }
+
+        if (this.queue.length !== 0) {
+            var points = [];
+
+            fs.statAsync(this.filename)
+            .then(function(stats) {
+                if (stats.size > self.maxFilesize) {
+                    throw Error('Exceeded max filesize of ' + self.maxFilesize + ' bytes');
```
demmer: This seems like it should trigger the `RT-PROC-LIMIT-EXCEEDED` error but from my read of the code it's going to show up as an internal error.

rlgomes:
@demmer @bkutil another look at the current solution. I opted for not having a `-flushInterval` and instead doing a `-flushFrequency` (open to suggestions on a better name) where we only flush when we've hit a target amount of points and not just periodically checking if we have more things to flush or not. The `-flushInterval` option was leading to some absurdly complicated code due to the background callbacks having to synchronize their work with the final flush from the `eof` call.

rlgomes:
Things to finish up:
 * [ ] add tests for option validations that aren't covered and showing up as uncovered lines in the coverage report.

demmer:
Overall this looks good to me now since the above comments are just a few minor suggestions.

rlgomes:
Ok merging on green as I've taken care of all @demmer 's suggestions and I think this is ready for prime time.

rlgomes:
@dmehra quick review on documentation additions: https://github.com/juttle/juttle/pull/11/files#diff-8cca8547a0a58a3918b9eb089cdda8a6

dmehra:
Doc review comment: how about a rephrase so defaults go into the "Required?" column, and don't mention `-file` because per https://github.com/juttle/juttle/issues/44 it will be renamed to `-path` anyway. Maybe like this:

```
`-bufferLimit`    | Maximum number of points to buffer in memory before each write to the specified file | No; defaults to 100000
`-maxFilesize`    | Maximum size of the file being written to, limited since the entire JSON needs to fit in memory | No; defaults to 200MB
`-flushFrequency` | How often (in number of points) to flush the points in memory out to the specified file | No; defaults to every 100 points
```

rlgomes:
@dmehra fixed!

Issue #13, opened 2015-12-21
json parser test intermittently fails


demmer:
I've noticed the json parser test for "can parse a single JSON point" has intermittently failed a couple of times in Travis:

```
  2) parsers/json can parse a single JSON point:
      Uncaught AssertionError: expected [] to deeply equal [ Array(1) ]
      + expected - actual
      -[]
      +[
      +  {
      +    "foo": "4"
      +    "time": "2014-01-01T00:00:04.000Z"
      +  }
      +]
```

See https://travis-ci.org/juttle/juttle/jobs/98195108 for an example.

demmer:
Fixed in #16.

PR #16, opened 2015-12-21
test: rework the unit tests for the parsers to be more robust

demmer:
Update the various parser tests to return a promise from each test
case and include the checks as part of the promise chain as opposed
to the previous approach in which the tests would throw expectation
errors from within the context of the emit callback.

This should help localize failures more clearly.

As part of this change, update the tests to properly handle the case
where the parser is given a limit on the number of points per emit
so that it properly expects an empty array at the end of the test.

This was causing intermittent failures in travis as identified in #13.

### DIFF
--- a/test/adapters/parsers/csv.spec.js
+++ b/test/adapters/parsers/csv.spec.js
@@ -15,7 +15,7 @@ describe('parsers/csv', function() {
 
     it('fails when given an invalid CSV stream', function() {
         var csv = parsers.getParser('csv');
-        csv.parseStream(fs.createReadStream(invalidFile))
+        return csv.parseStream(fs.createReadStream(invalidFile))
         .then(function() {
             throw Error('previous statement should have failed');
         })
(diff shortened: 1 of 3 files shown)


demmer:
@rlgomes please take a look.

test/adapters/parsers/csv.spec.js
```diff
@@ -15,7 +15,7 @@ describe('parsers/csv', function() {
 
     it('fails when given an invalid CSV stream', function() {
         var csv = parsers.getParser('csv');
-        csv.parseStream(fs.createReadStream(invalidFile))
+        return csv.parseStream(fs.createReadStream(invalidFile))
```
rlgomes: add the `function() {}` in here for emit since its generating those failures due to the timing of error out vs hitting the emit in the `end` event.

rlgomes:
+1 with the fix for the failing tests.

rlgomes:
Merging this as it makes these tests less flakey and they're failing on a few other branches.

PR #19, opened 2015-12-22
Parser cleanups: Structural changes

dmajda:
A couple of structural changes in the `parser` module which make `parse` and `parseSync` more readable, more DRY, and closer together structurally. Eventually, they should be identical except for the sync/async difference.


dmajda:
@bkutil

bkutil:
Looks good. +1

PR #20, opened 2015-12-22
Add live emitter to handle batch/sort of points

bkutil:
A live emitter is responsible for batching points, then emitting them sorted, while allowing for 'late arrivals' within a given time window.

Internally, it is represented by a time-based priority queue with configurable maximum size and time window. Points arriving outside of it are dropped. When queue size limit is reached, emitter drops oldest points first. A periodic process is responsible for emitting points within the time window.

This is a common logic that will be used across adapters, so it makes sense to keep it in a separate class.

### DIFF
--- a/lib/runtime/live-points-queue.js
+++ b/lib/runtime/live-points-queue.js
@@ -1,0 +1,115 @@
+var _ = require('underscore');
+
+var Base = require('extendable-base');
+var Heap = require('heap');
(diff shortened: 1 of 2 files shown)


bkutil:
cc: @demmer @dmajda

bkutil:
Closing for now as I'm working on the nested points ATM. Will re-open when this will become active again.

PR #22, opened 2015-12-22
don't interleave output from multiple client views

demmer:
Currently the following juttle program:

```
emit -from :2015-01-01: -limit 2
| (put sink=1 | view table;
   put sink=2 | view table)
```

Would result in the following output:

```
┌────────────────────────────────────┬──────────┐
│ time                               │ sink     │
├────────────────────────────────────┼──────────┤
│ 2015-01-01T00:00:00.000Z           │ 1        │
┌────────────────────────────────────┬──────────┐
│ time                               │ sink     │
├────────────────────────────────────┼──────────┤
│ 2015-01-01T00:00:00.000Z           │ 2        │
├────────────────────────────────────┼──────────┤
│ 2015-01-01T00:00:01.000Z           │ 1        │
├────────────────────────────────────┼──────────┤
│ 2015-01-01T00:00:01.000Z           │ 2        │
└────────────────────────────────────┴──────────┘
└────────────────────────────────────┴──────────┘
```

This is because progressive mode defaults to true, so the two views interleave their output.

Change the initialization API so that the client-sink-mgr sets the default progressive mode option to true only if there is exactly one view in the program.

Also add a progressive option to the text view to address the same problem.

### DIFF
--- a/lib/sinks/client-sink-mgr.js
+++ b/lib/sinks/client-sink-mgr.js
@@ -33,7 +33,8 @@ var ClientSinkManager = Base.extend({
     setup: function(program) {
         var self = this;
 
-        self.program.get_client_sinks(program).forEach(function(sink) {
+        var sinks = self.program.get_client_sinks(program);
+        sinks.forEach(function(sink) {
 
             var sink_class;
             if (_.has(self.sink_classes, sink.name)) {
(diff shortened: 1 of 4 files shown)


demmer:
@mstemm please take a look

mstemm:
+1. now the rebase from heck for my pr.

Issue #21, opened 2015-12-22
read file adapter incorrectly parsing JSON payload with nested array


rlgomes:
So with the following file:
```
{
    "time": "2014-01-01T00:00:00.000Z",
    "foo": [
        {"fizz1": "buzz"},
        {"fizz2": "buzz"},
        {"fizz3": "buzz"}
    ]
}
```
We're getting this odd behaviour:
```
rlgomes@x230> ./bin/juttle -e "read file -file 'test/adapters/parsers/input/json/nested_array.json' | view text"
[
{"fizz1":"buzz"},
{"fizz2":"buzz"},
{"fizz3":"buzz"},
{"time":"2014-01-01T00:00:00.000Z","foo":[{"fizz1":"buzz"},{"fizz2":"buzz"},{"fizz3":"buzz"}]}
]
```

PR #23, opened 2015-12-22
fixes json parsing issue identified in issue #21

rlgomes:
this isn't elegant but it does fix the issue while still using the
oboejs library.


rlgomes:
@demmer @mstemm @davidvgalbraith if you have a better idea suggestions are welcome. I think that we could get this in so things aren't broken as they currently are and then I could open an issue to replace oboejs with something else.

lib/adapters/parsers/json.js
```diff
@@ -13,12 +13,16 @@ module.exports = base.extend({
...
             oboe(stream)
-            .node('*', function (pt) {
-                if (!_.isObject(pt) || _.isArray(pt) ) {
+            .node('*', function (point, path) {
+                if (!_.isObject(point) || _.isArray(point) ) {
                     return;
                 }
 
-                buffer.push(pt);
+                if (path.length > 1 && !_.isNumber(path[0])) {
```
demmer: I think you should put a comment explaining what this check actually enforces.

test/adapters/parsers/json.spec.js
```diff
@@ -56,6 +57,27 @@ describe('parsers/json', function() {
         });
     });
 
+    it('can parse a file with nested arrrays correctly', function() {
```
demmer: I also think there should be a test case for an array of objects with nested objects / arrays.

demmer:
Other than the minor comments above, this looks good to me.

rlgomes:
@demmer @davidvgalbraith fixed and squashed. Demmer thanks for the test idea as it showed that second condition needed some fixing.

test/adapters/parsers/json.spec.js
```diff
@@ -56,6 +58,48 @@ describe('parsers/json', function() {
...
+                        {'fizz1': 'buzz'},
+                        {'fizz2': 'buzz'},
+                        {'fizz3': 'buzz'}
+                    ]
+                }
+            ]]);
+        });
+    });
+
+    it('can parse a JSON array with nested arrrays correctly', function() {
```
davidvgalbraith: What if we go crazy and have nested objects in our nested arrays and those nested objects have nested arrays, etc?

rlgomes: don't see how that would fail at this point given the checks we have in place using the `path` argument to ignore all nodes that have a path that has more than 1 parent.

rlgomes:
@demmer @davidvgalbraith shall I merge ?

davidvgalbraith:
+1

PR #24, opened 2015-12-22
Rename sink to view

### LINKED ISSUE #40: Rename client sink to view in the code
We're trying to standardize on the name "view" for the endpoint of a
program that is viewed (either in the cli as a table/text block, or in
the browser). So go through a big rename exercise to clean that up.

mstemm:
We're trying to standardize on the name "view" for the endpoint of a
program that is viewed (either in the cli as a table/text block, or in
the browser). So go through a big rename exercise to clean that up.

Generally anywhere where the term sink was used when the thing was
actually a viewable item, switch to view. The term sink still has
meaning for the union of views and writes.

This change is incompatible with 0.1.0, so once published we'll have
to bump the version to 0.2.0.

This fixes #40.

@demmer @dmajda


lib/program.js
```diff
@@ -142,14 +142,14 @@ var Program = Base.extend({
         });
     },
 
-    // returns a list of the sinks of this program, formatted for use by the app
-    get_client_sinks: function() {
+    // returns a list of the views of this program, formatted for use by the app
+    get_client_views: function() {
```
demmer: I think we should just use "view" instead of "clientview" as the proc name and call this simply `get_views()`.

demmer:
This looks like a great start and covers pretty much all the renaming from the compiler out.

However I think we should go further though and rename the Sink/SinkOption AST nodes along with the rest of the references to "Sink" inside the compiler to be "View" for consistency.

lib/program.js
mstemm: Ok renamed.

mstemm:
How about I merge this today and we address the additional renaming in a separate PR? I'd rather not have this stay on a branch over the break.

mstemm:
bump

lib/compiler/graph-compiler.js
```diff
@@ -462,7 +462,7 @@ var GraphCompiler = CodeGenerator.extend({
     gen_Sink: function(ast) {
         // silly special case for ast.name due to sink's special-case parsing
         // and AST.
-        var proc = this.gen_proc(ast, 'clientsink', '{name: "'+ast.name+'"}');
+        var proc = this.gen_proc(ast, 'clientview', '{name: "'+ast.name+'"}');
```
demmer: Although I think this should simply be 'view' and not 'clientview', that can be done in a subsequent change when we also rename the AST nodes from `Sink` to `View`. At that point I think we can just use 'view' consistently.

lib/views/client-view-mgr.js
```diff
@@ -4,14 +4,14 @@ var Promise = require('bluebird');
 var Base = require('extendable-base');
 var JuttleErrors = require('../errors');
 
-var ClientSinkManager = Base.extend({
+var ClientViewManager = Base.extend({
```
demmer: Why not `ViewManager`?

lib/views/client-view.js
```diff
@@ -3,12 +3,12 @@ var Base = require('extendable-base');
...
 
-// This is a base class used by the other terminal sinks (text,
+// This is a base class used by the other terminal views (text,
 // table). Objects of this class should not be created directly.
 //
-// The Client sink manager creates objects derived from this base
+// The Client view manager creates objects derived from this base
 // class to handle streams of data from juttle programs.
-var ClientSink = Base.extend({
+var ClientView = Base.extend({
```
demmer: Similarly I think this should just be `View`.

lib/compiler/graph-compiler.js
demmer: Actually looking at this further, I see there are a number of places in the tests etc where we now reference `clientview` instead of `clientsink` so my preference would be to change them all to `view` once and for all rather than wait for another PR to rename it again.

lib/views/client-view-mgr.js
mstemm: Sure, I'll rename (and also rename lib/views/client-view-mgr.js to lib/views/view-mgr.js).

lib/views/client-view.js
mstemm: Sure, also renaming lib/views/client-view.js to lib/views/view.js.

dmajda:
@mstemm I got to this only now, but in the future please do 2 things:

  * Split big renaming changes like this into more commits, each dealing with one part (e.g. rename `ClientSink`, rename each view, rename of this or that method, etc.). As part of the review I need to be able to verify that each commit is correct (no instance was missed, etc.) and the ideal state is when the commit is “obviously correct” because it is very simple. I find it impossible to do this verification with your first commit and thus this PR would not get a +1 from me in its current form.

  * Squash commits with review-induced changes into the original commits, which leads to more sensible history (i.e. not a set of zig-zag changes).

Thanks!

PR #25, opened 2015-12-22
adapters: downgrade initialization logging and add timing info

demmer:
Downgrade the adapter init logging message from info to debug and
include some timing info to see how long it takes for the adapter to
load and to be initialized.


demmer:
@mstemm this is a simple one

mstemm:
lgtm

PR #28, opened 2015-12-22
Rendering the project README as part of the docs site

### LINKED ISSUE #27: Include project README into the docs site
i redeployed the docs to http://juttle.github.io/juttle/ (with added CLI reference) and noticed that if you come to the docs, not having read the README in the product itself, the experience is odd, since the README has overview which the doc site does not. Neither "About" nor "Overview" sections in the docs replace the readme.

README.md can be added to the docs as "Project README" article via symlink. The only limitation is that we must now have relative links in the README as they would all break, it should use absolute URLs pointing at gh-pages site.

dmehra:
resolves #27 

Looks like this:

image

Already ok'd by @demmer, will merge and re-publish docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 Juttle is an analytics system for developers built upon
 a stream-processing core and a
-[dataflow language](./docs/concepts/dataflow.md)
+[dataflow language](http://juttle.github.io/juttle/concepts/dataflow)
 that tightly integrates with
 [streaming visualizations](http://github.com/juttle/juttle-viz).
 The Juttle language lets you query,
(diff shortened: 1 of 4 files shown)


demmer:
I think there was a bit of a miscommunication on this front since I don't think it makes sense to have the docs files explicitly link to the currently hosted docs site.

I've found it useful with the relative links to be able to browse the docs files directly from within the github UI or by using an offline markdown viewer like Marked, both of which would be broken if we use the absolute external links like you've done here.

My instinct is that instead of trying to make the whole README a part of the docs site, that we just add explicit links back to the github site from within the docs. Something like https://github.com/juttle/juttle/compare/readme-in-docs-alternative though perhaps with better copy.

dmehra:
My point was that it isn't just installation instructions that README has and main docs don't - it's use case examples and overview text. 

If we have absolute links to github in the README (rather than to docs site), would that be good enough? The offline nav through those few links wouldn't work I suppose, but we'd have a cohesive message in the docs so we can point interested people at that one location.

I'm loath to duplicate the content in two files, but another option is to add "Use Cases" section to the docs, which currently would copy from README but later would grow.

dmehra:
See https://github.com/juttle/juttle/pull/70 for a new approach.

dmajda:
It looks like this PR is obsolete ⇒ closing.

@dmehra If I am mistaken, please reopen.

PR #70, opened 2015-12-28
Bringing usage examples into juttle docs

### LINKED ISSUE #27: Include project README into the docs site

dmehra:
Another attempt to resolve #27. This creates a new docs section "Usage Examples" which currently has a copy of 3 examples from the project readme, plus the sine wave one from the still-private demo. The "Coming soon" list is where this is headed. 

The problem I'm struggling with is where to put the actual juttle code (and dockerfiles, once we're setting these up as self contained demos). The juttle repo doesn't seem like the right place, since we won't be able to syntax check the examples code. Should they live in outrigger? If we do that, inlining the juttle code in the docs will no longer be possible, unless we set up separate docs for outrigger (at the moment it only has a README). Maybe that's the way to go. Then juttle's "Usage Examples" will only include short description for each one, and link out to the right page of outrigger demo docs.

@rlgomes @demmer @dmajda @bkutil thoughts?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,6 +2,16 @@
 
 Welcome to the documentation for [Juttle](https://github.com/juttle/juttle/), a dataflow programming language!
 
+## Project Info
+
+See the [GitHub project README](http://github.com/juttle/juttle) for installation and setup instructions, and if you wish to contribute.
+
+If you discover a bug in Juttle or these docs, please [open a GitHub issue](https://github.com/juttle/juttle/issues/new) or talk to us on [Gitter](https://gitter.im/juttle/juttle).
+
+Our public [Waffle board](https://waffle.io/juttle/juttle) tracks GitHub issues and PRs in a single place.
+
+This documentation was produced with [mkdocs](http://mkdocs.org), see details in [README](README.md).
+
 ## About Juttle
 
 Juttle is an analytics system and language for developers built upon
@@ -66,6 +76,7 @@ Here are some ideas of what you can do with Juttle:
   your data before implementing them in detail and in production using other
   stream-processing systems like Spark or Storm.
 
+Refer to the [Usage Examples](use_cases.md) section for more specific examples with Juttle programs and demos.
 
 ## Stream-processing Systems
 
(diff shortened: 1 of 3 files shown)


dmehra:
(The travis build is failing on some slow unit test, unrelated to docs-only change here)

docs/index.md
```diff
@@ -2,6 +2,16 @@
...
 Welcome to the documentation for [Juttle](https://github.com/juttle/juttle/), a dataflow programming language!
 
+## Project Info
+
+See the [GitHub project README](http://github.com/juttle/juttle) for installation and setup instructions, and if you wish to contribute.
+
+If you discover a bug in Juttle or these docs, please [open a GitHub issue](https://github.com/juttle/juttle/issues/new) or talk to us on [Gitter](https://gitter.im/juttle/juttle).
+
+Our public [Waffle board](https://waffle.io/juttle/juttle) tracks GitHub issues and PRs in a single place.
+
```
dmehra: Should the two lines above go into juttle/README.md instead of here? @dmajda

docs/index.md
```diff
@@ -87,20 +98,3 @@ the Juttle compiler and runtime to generate scala in addition to
...
-Juttle examples are provided as code snippets; copy them to run in your own environment.
-
-:construction: Embedded Juttle examples are coming soon.
-
-## Maintaining Docs
-
-:baby_symbol: These docs are just coming together.
-
-If you find a problem in the documentation, members of juttle GitHub project can edit the articles inline by following "Edit on GitHub" link. Small obvious fixes may be committed directly to master. If discussion is needed, please put up a PR on a branch.
-
```
dmehra: The docs maintenance part is described in docs/README.md.

docs/index.md
```diff
@@ -87,20 +98,3 @@ the Juttle compiler and runtime to generate scala in addition to
 javascript as a target output langauge and leverage the Spark infrastructure
 for Juttle programs.
 
-## How To Navigate
-
```
dmehra: This section no longer seems necessary.

docs/index.md
```diff
@@ -87,20 +98,3 @@ the Juttle compiler and runtime to generate scala in addition to
...
-Use the left sidebar menu for navigating the sections. Getting started with the [Juttle Overview](concepts/overview.md) should orient you reasonably well; the rest of the Concepts section has background information on the language design.
-
-The search feature (powered by lunr.js in mkdocs) can be helpful too, see the search box in upper left corner.
-
-Juttle examples are provided as code snippets; copy them to run in your own environment.
-
-:construction: Embedded Juttle examples are coming soon.
```
dmehra: No they aren't, we left that idea alone, figuring users would run the examples from CLI or in outrigger if they want to.

PR #18, opened 2015-12-22
ClientSink: Remove `file` sink from comment (it no longer exists)

dmajda:

PR #71, opened 2015-12-28
increase the test timeout and also tweak the slow value

rlgomes:
with these settings we'll have better settings for running tests in
general in travis

fixes #18 by simply tweaking the timeouts to better match the expectations of running tests in travis

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -122,7 +122,8 @@ function gulp_test() {
     return gulp.src('test/**/*.spec.js')
         .pipe(mocha({
             log: true,
-            timeout: 5000,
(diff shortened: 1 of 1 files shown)


rlgomes:
@dmehra @davidvgalbraith this should minimize the amount of false positives in travis at this point. i've increased the test runtime timeout from 5s to 10s and then made it so the `slow` setting is at 3s instead of the default mocha of 75ms which is unrealistic for most of the juttle tests. The last `slow` setting will reduce the amount of little red (xxxms) indicating a test took more than 75ms.

davidvgalbraith:
+1

PR #73, opened 2015-12-28
add -from/-to to `read file`

### LINKED ISSUE #65: have `read file` support -from / -to
The file backend does not support time filtering options, for no reason other than it hasn't been implemented yet. See if #64 makes this easier.

rlgomes:
fixes #65

this includes the ability to specify the -from/-to options to `read
file` as well as the ability to detect out of order points and drop
them with a warning.

### DIFF
--- a/lib/adapters/file-adapter.js
+++ b/lib/adapters/file-adapter.js
@@ -11,15 +11,17 @@
 
 var _ = require('underscore');
 var Juttle = require('../runtime/index').Juttle;
+var JuttleMoment = require('../moment').JuttleMoment;
 var parsers = require('./parsers');
 var fs = require('fs');
+var values = require('../runtime/values');
 
 var Read = Juttle.proc.source.extend({
     sourceType: 'batch',
     procName: 'read-file',
 
     initialize: function(options, params, pname, location, program, juttle) {
-        var allowed_options = ['file', 'timeField', 'format'];
+        var allowed_options = ['from', 'to', 'file', 'timeField', 'format'];
         var unknown = _.difference(_.keys(options), allowed_options);
         if (unknown.length > 0) {
             throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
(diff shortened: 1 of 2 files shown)


rlgomes:
@davidvgalbraith @go-oleg since you guys are around what do you think ? I obviously would add the necessary tests before anything else but do you think this would be a "good" way to support `-from/-to` on `read file`. 

@dmehra if the reviewers are good with this I can add the tests later today or tomorrow and then your PR for the tutorial can actually use the `-from/-to` filtering and I can just rework PR #11 since I'm sure that won't get thoroughly reviewed till we're all back from holidays.

rlgomes:
couple of failing tests are due to reading files with out of order data which would obviously be unsupported at this time and result in drops... I'll fix those once I get a review on the existing approach.

lib/adapters/file-adapter.js
```diff
@@ -56,10 +72,28 @@ var Read = Juttle.proc.source.extend({
...
-            self.emit(points);
+            _.each(points, function(point) {
+                if (last && last.gt(point.time)) {
+                    // out of order points are dropped
+                    self.trigger('warning', self.runtime_error('RT-OUT-OF-ORDER'), {
+                        proc: 'read file'
+                    });
+                } else {
+                    // timeless points just pass through at this point
+                    if (point.time && point.time.moment) {
```
go-oleg: You can use `values.typeOf(time) === 'Date'` to do this check.

lib/adapters/file-adapter.js
```diff
@@ -56,10 +72,28 @@ var Read = Juttle.proc.source.extend({
...
+                        if (point.time.gte(self.from) && point.time.lt(self.to)) {
```
go-oleg: Wondering whether we want to emit an EOF as soon as we see a point thats more recent than `self.to`. This would be an optimization and not a semantic change in behavior so not necessary to address in this PR.

lib/adapters/file-adapter.js
```diff
@@ -56,10 +72,28 @@ var Read = Juttle.proc.source.extend({
...
+                            validPoints.push(point);
+                        }
+                    }
+                }
+                last = point.time;
```
go-oleg: This should be moved up into the `else` above so that times of OUT-OF-ORDER points don't move `last` backwards.

lib/adapters/file-adapter.js
rlgomes: time is a moment its type will come back as Date ? @go-oleg

lib/adapters/file-adapter.js
rlgomes: Good catch and I'll write an actual tests that hits this first.

lib/adapters/file-adapter.js
go-oleg: Yea, that `values` is our [library](https://github.com/juttle/juttle/blob/c3f5ab745e2123ac2f040f02d16b0e1f81006e37/lib/runtime/values.js#L48-L56).

go-oleg:
The approach looks good to me.

rlgomes:
@go-oleg @davidvgalbraith @dmehra alright fixed up to not drop points at all and added a few simple unit tests.

davidvgalbraith:
+1

rlgomes:
@dmehra this is merged to master so you can actually use `read file` with `-from/-to`

PR #74, opened 2015-12-28
remove tests using non monotonic data

rlgomes:
these tests ended up testing an unintended set of behaviours of the
`read file` adapter.


rlgomes:
@davidvgalbraith @go-oleg @dmehra

dmehra:
+1 and we can bring back non-monotonic tests later once full complement of reviewers are back online after holidays, if those tests had a point that we're missing.

rlgomes:
Will revisit this next week as removing these isn't the right thing to do....

PR #78, opened 2015-12-29
stdio adapter to easily read/write data from stdin/stdout

### LINKED ISSUE #36: add a stdin backend to read data directly from stdin
The idea being that you'd be able to simply do:

```bash
cat /some/file.log | juttle -e "read stdin | ..."
```

instead of something like:

```bash
juttle -e "read file -file 'cat/some/file.log' | ..."
```

Also this would allow you to do things like tailing a file from a local or remote system:

```bash
tail -f /some/local.log | juttle -e "read stdin | ..."
```

rlgomes:
fixes #36 

Just a simple implementation of an stdin adapter that would allow anyone to cat a JSON or CSV file directly to the juttle binary and process that data without having to stick the name of the file into the juttle code itself.


rlgomes:
@davidvgalbraith @go-oleg

lib/adapters/stdin-adapter.js
```diff
@@ -0,0 +1,57 @@
...
+var Read = Juttle.proc.source.extend({
+    sourceType: 'batch',
+    procName: 'read-stdin',
+
+    initialize: function(options, params, pname, location, program, juttle) {
+        var allowed_options = ['timeField', 'format'];
+        var unknown = _.difference(_.keys(options), allowed_options);
+        if (unknown.length > 0) {
+            throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
+                proc: 'read file',
```
davidvgalbraith: Should be `read-stdin`? Or better yet `this.procName`.

davidvgalbraith:
No tests? Think you could do `process.stdin.write(JSON.stringify(some_points))` to simulate the intended use.

rlgomes:
@davidvgalbraith yup I'll be adding tests + documentation of course. Was thinking about how to easily unit test this but maybe the `process.stdin.write` just might work for unit testing directly without the use of the CLI itself.

go-oleg:
Thoughts on renaming this to `TerminalAdapter` or `IOAdapter` so that it can support both `read` (stdin) and `write` (stdout)?

rlgomes:
@go-oleg I thought about it but then what would the `write stdout` do differently from the `view text` sink ?

rlgomes:
one thought was to support something simple such as `write stdout -template "{{field1}} - {{field2}}"` which means allowing the user to reconstruct the output exactly as they see fit using a template per point ?

go-oleg:
When using the CLI, `write stdout` and `view text` end up having pretty much the same behavior. The difference though is who is actually responsible for writing to `stdout`. In the case of `view text`, its the CLI connecting the output of the juttle runtime to the sink. In the case of `write stdout`, it would be the juttle runtime itself writing to `stdout`. That means if you used `write stdout` and programmatically ran a juttle program via the compiler/runtime API, it would "just work" instead of you having to write the output of the `view text` to `stdout`. Does that make sense?

go-oleg:
Defaulting to `jsonl` but letting a user specify their own templated format for the output snds useful!

rlgomes:
@davidvgalbraith @go-oleg did a bunch of work to make `read stdin` and `write stdout` work and also introduce the notion of `serializers` let me know what you think and after the new years I'll add all of the missing tests + documentation here. 

I'm considering introducing a format for writing (possibly reading) in which you can do something like so:

```
... | write stdout -format "handlebars" -formatOptions { template: "{{time}} : {{somefield}} - {{some_other_field}}" } 
```

What do you think ? is there a nicer way to expose the format specific options ? also this `-formatOptions` would be useful when writing out lets say `csv` so you can change the **delimiter** or **header** line. 

The `handlebars` format would be introduced in a follow up PR as I don't want this PR to get bloated.

rlgomes:
Oh and as for testing the serializers and parsers I'm going to make sure to use the json serializer to verify the json parser works and vice-versa which will make for a good way to verify we can always consume what we produce in juttle.

rlgomes:
@davidvgalbraith @go-oleg @demmer bump!

lib/adapters/index.js
```diff
@@ -3,8 +3,11 @@ var Juttle = require('../runtime/index').Juttle;
 var stochastic = require('./stochastic-adapter');
 var file = require('./file-adapter');
 var http = require('./http');
+var io = require('./io');
 
 // register the built-in adapters
 Juttle.adapters.register('file', file({}));
 Juttle.adapters.register('http', http({}));
 Juttle.adapters.register('stochastic', stochastic({}));
+Juttle.adapters.register('stdin', io.StdinAdapter({}));
```
go-oleg: I was actually envisioning putting the two in one adapter (`io` or `terminal`) that implements both `read` and `write`, so the juttle would look like

```
read io  | write io;
// or
read terminal | write terminal
```

Thoughts?

lib/adapters/serializers/base.js
```diff
@@ -0,0 +1,15 @@
+var Base = require('extendable-base');
+
+module.exports = Base.extend({
+    initialize: function(stream, options) {
+        this.stream = stream;
+    },
+
+    write: function(points) {
```
go-oleg: I think writing to should be the responsibility of the adapter and not the serializer. In that case, this method would be `serialize` and would return a string which the adapter would then write to its stream.

lib/adapters/index.js
rlgomes: humm I guess its not a bad thing since I can't right now see a reason to want to write to stderr and therefore `read io` reads from stdin and `write io` writes to stdout and if someone really wants the option to write to stderr then we'd simply add the option `write io -stderr true` for that very special case.

lib/adapters/serializers/csv.js
```diff
@@ -0,0 +1,16 @@
+var _ = require('underscore');
+var base = require('./base');
+
+module.exports = base.extend({
+
+    write: function(points) {
+        var self = this;
+        _.each(points, function(point) {
+            var line = [];
+            _.each(point, function(value, key) {
```
go-oleg: The csv serializer needs to write a header line with the field names, and then write out the fields in the right order (iterating over the fields of an object isn't guaranteed to always return them in the same order).

go-oleg:
@rlgomes, I like the `serializer` thing. I'm wondering if we should merge the `parsers` and `serializers` so that each class (`json`, `jsonl`, `csv`, etc.) has both a `serialize` and `deserialize` (or `parse`) function. This will help ensure symmetry (i.e. that `deserialize(serialize(someJSObj))` is an _identity_ function.

rlgomes:
@go-oleg have a nother look as I've updated most of the things you asked for except doing anything drastic to the serializer way of handling data.

@demmer @bkutil @davidvgalbraith @mstemm @dmehra 

Decided to experiment with a regex format for input that can be used to do things like so:

```
cat /var/log/auth.log | ./bin/juttle -e "read io -format 'regex' -formatOptions { regex: '(?<time>[^x]+) (?<message>.*)' } | view text" 
[
{"time":"2001-01-03T17:03:54.000Z","message":"x230 sudo:  rlgomes : TTY=unknown ; PWD=/home/rlgomes ; USER=root ; COMMAND=/usr/lib/linuxmint/mintUpdate/checkAPT.py"},
{"time":"2001-01-03T17:03:54.000Z","message":"x230 sudo: pam_unix(sudo:session): session opened for user root by (uid=0)"},
{"time":"2001-01-03T17:05:55.000Z","message":"x230 sudo: pam_unix(sudo:session): session closed for user root"},
{"time":"2001-01-03T17:17:01.000Z","message":"x230 CRON[14287]: pam_unix(cron:session): session opened for user root by (uid=0)"},
{"time":"2001-01-03T17:17:01.000Z","message":"x230 CRON[14287]: pam_unix(cron:session): session closed for user root"},
{"time":"2001-01-03T17:23:10.000Z","message":"x230 cinnamon-screensaver-dialog: gkr-pam: unlocked login keyring"},
{"time":"2001-01-03T17:29:20.000Z","message":"x230 cinnamon-screensaver-dialog: gkr-pam: unlocked login keyring"},
{"time":"2001-01-03T22:05:34.000Z","message":"x230 cinnamon-screensaver-dialog: gkr-pam: unlocked login keyring"},
{"time":"2001-01-03T22:05:47.000Z","message":"x230 sudo:  rlgomes : TTY=unknown ; PWD=/home/rlgomes ; USER=root ; COMMAND=/usr/lib/linuxmint/mintUpdate/checkAPT.py"},
{"time":"2001-01-03T22:05:47.000Z","message":"x230 sudo: pam_unix(sudo:session): session opened for user root by (uid=0)"},
{"time":"2001-01-03T22:06:00.000Z","message":"x230 sudo: pam_unix(sudo:session): session closed for user root"},
{"time":"2001-01-03T22:17:01.000Z","message":"x230 CRON[16029]: pam_unix(cron:session): session opened for user root by (uid=0)"},
{"time":"2001-01-03T22:17:01.000Z","message":"x230 CRON[16029]: pam_unix(cron:session): session closed for user root"},
{"time":"2001-01-03T22:35:48.000Z","message":"x230 sudo:  rlgomes : TTY=unknown ; PWD=/home/rlgomes ; USER=root ; COMMAND=/usr/lib/linuxmint/mintUpdate/checkAPT.py"},
{"time":"2001-01-03T22:35:48.000Z","message":"x230 sudo: pam_unix(sudo:session): session opened for user root by (uid=0)"},
{"time":"2001-01-03T22:36:00.000Z","message":"x230 sudo: pam_unix(sudo:session): session closed for user root"},
{"time":"2001-01-03T22:56:13.000Z","message":"x230 cinnamon-screensaver-dialog: gkr-pam: unlocked login keyring"},
{"time":"2001-01-03T22:56:16.000Z","message":"x230 dbus[858]: [system] Rejected send message, 7 matched rules; type=\"method_return\", sender=\":1.17\" (uid=0 pid=1927 comm=\"/usr/sbin/dnsmasq --no-resolv --keep-in-foreground\") interface=\"(unset)\" member=\"(unset)\" error name=\"(unset)\" requested_reply=\"0\" destination=\":1.5\" (uid=0 pid=984 comm=\"NetworkManager \")"},
...
```

I can definitely move that one commit to a separate follow up PR but just wanted to further expand the ideas around reading logs from stdin and immediately making use of them in juttle.

rlgomes:
Oh and I'm not done with the tests as I'm going to rejig them so that the parser and serializer tests use each other to test that they can produce and consume each others output.

rlgomes:
Ignore the regex commit since  I'm redoing that using grok instead

rlgomes:
Ok so I quickly put together an example of how we can expose grok as a format option to for parsing data and allow users to quickly parse existing log data. I can elaborate in a separate PR exactly how to use this to do your own custom parsing as well as using built in `patterns` that node-grok already handles for you.

lib/adapters/serializers/csv.js
```diff
@@ -0,0 +1,40 @@
+var _ = require('underscore');
+var base = require('./base');
+var csv = require('csv-write-stream');
+
+module.exports = base.extend({
+
+    initialize: function() {
+        this.header;
```
go-oleg: accidentally left over from something?

lib/adapters/serializers/csv.js
```diff
@@ -0,0 +1,40 @@
...
+    },
+
+    write: function(points) {
+        var self = this;
+
+        _.each(points, function(point) {
```
go-oleg: just an fyi, you can pass `this` as a third argument to `_.each` to maintain context (instead of using `self`)

go-oleg:
@rlgomes, lookin pretty good to me. Its getting a bit big though, so perhaps someone else should take a look too.

lib/adapters/serializers/csv.js
rlgomes: I knew that but hadn't seen us doing that elsewhere so thought maybe it wasn't the way we handled this kind of situation ?

rlgomes:
@go-oleg I'm going to clean up the PR (squash & rebase) and then add the documentation for review of the io adapter.

rlgomes:
@dmehra review my documentation please :)

rlgomes:
@davidvgalbraith @demmer does this look good to go ?

docs/adapters/io.md
```diff
@@ -0,0 +1,78 @@
...
+
+Supported formats are JSON array, JSON lines and CSV; see examples below.
+
+```text
+read io [-format <format>] [-timeField <fieldname>]
+```
+
+Parameter         |             Description          | Required?
+----------------- | -------------------------------- | ---------:
+`-file`           | File path on the local filesystem, absolute or relative to the current working directory  | Yes
```
dmehra: what's that doing in here

docs/adapters/io.md
```diff
@@ -0,0 +1,78 @@
...
+`-format`         | Input file format: `csv` for [CSV](https://tools.ietf.org/html/rfc4180) data, `json` for [JSON](https://tools.ietf.org/html/rfc7159) data, or `jsonl` for [JSON lines](http://jsonlines.org/) data | No; defaults to `json`
```
dmehra: just "Input format", not "file format"

docs/adapters/io.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+cat myfile.json | juttle -e "read io | filter hostname = 'lemoncake' | view table"
+```
+
+_Example: Process a remote log file using logstash and redirect it to be processed in juttle
+
+Similar to the previous example but with use of logstash to process logs into json data:
+
+```bash
+logstash -f myconfig.conf  | juttle -e "read io | reduce avg_resp_ms=avg(resp_ms) by hostname | sort avg_resp_ms -desc"
```
dmehra: missing "cat log.log | logstash"

docs/adapters/io.md
```diff
@@ -0,0 +1,78 @@
...
+
+Supported formats are JSON array, JSON lines and CSV; see examples below.
+
+```text
+... | write io [-format <format>]
+```
+
+Parameter         |             Description          | Required?
+----------------- | -------------------------------- | ---------:
+`-format`         | Input file format: `csv` for [CSV](https://tools.ietf.org/html/rfc4180) data, `json` for [JSON](https://tools.ietf.org/html/rfc7159) data, or `jsonl` for [JSON lines](http://jsonlines.org/) data | No; defaults to `json`
```
dmehra: s/Input file/Output

docs/adapters/io.md
```diff
@@ -0,0 +1,78 @@
...
+
+If the file already exists and contains a valid JSON array, the write will append new data rather than overwrite.
```
dmehra: That whole statement does not apply

rlgomes:
Documentation fixed but I still have to figure out what is the best way to warn in this situation:

https://github.com/juttle/juttle/pull/78/files#diff-18b8cd2be0c1fcfa28d90f3736dc5f9fR24

ideas @go-oleg @demmer @davidvgalbraith  ?

lib/adapters/serializers/csv.js
davidvgalbraith: That strategy is all over the UI, but you rarely see it on the backend for whatever reason.

go-oleg: Yea, whatever works :-).

davidvgalbraith:
+1. I have full confidence in your ability to come up with something reasonable to do for that warning.

docs/adapters/index.md
```diff
@@ -12,6 +12,7 @@ These adapters are included in the Juttle runtime and can be used in Juttle prog
 
 * [file](../adapters/file.md)
 * [http](../adapters/http.md)
+* [io](../adapters/io.md)
```
demmer: I'd find `io` to be too generic of a name for stdin / stdout. Every adapter is technically some kind of input/output.

I know this is contrary to what @go-oleg suggested, but I find `read stdin` and `write stdout` to be more intuitive than `read io`.

go-oleg: Yea, my other thought was using `terminal` (which may not actually be technically accurate) for the adapter name. Wasn't particularly fond of `io` or `terminal`, but thought both the `read` and `write` should be in the same adapter...perhaps not.

Could the adapter be called `stream`?
```
read stream -name "STDIN"
| (
  filter error != null | write stream -name "STDERR";
  filter error == null | write stream -name "STDOUT"
)
```
Meh...a bit more verbose.

rlgomes: `read stdio` @demmer ?

demmer: I could live with `read stdio` and `write stdio`. That would enable `write stdio -stream 'stderr'` as @go-oleg suggested above.

rlgomes:
@demmer the serializer emits an `error` but the `stdio` adapter actually fires off a warning (https://github.com/juttle/juttle/pull/78/files#diff-e3e7402bcd2b6916803d9d79817be916R26) as that has been the approach so far with handling errors at the end of a pipeline.

demmer:
Ah ok. That seems reasonable 

-m

rlgomes:
@demmer @go-oleg @davidvgalbraith unless there are any objections in the morning I'll merge this as I've updated things with the last requested changes.

demmer:
This looks good to me.

For future reference It would have been cleaner to have separate commits to add the serializers from adding the adapter itself.

rlgomes:
@demmer Agreed as I could have fixed the file adapter to use the serializers (still missing and will file an issue for that).

PR #87, opened 2016-01-03
cli: add debug logging to show juttle errors with stack traces

demmer:
The normal processing of any errors in the juttle just prints the
error message but doesn't include the full stack trace. This is
appropriate for problems in the user programs, but makes it a pain
to track down errors when working on the compiler core.

To help this, add a debug log for any errors that come out of the execution or
compiler that includes the full stack trace.


demmer:
@dmajda @mstemm

mstemm:
lgtm.

PR #88, opened 2016-01-03
use external moment parser library

### LINKED ISSUE #26: use external moment-parser
Juttle should use the external moment-parser project instead of having its own internal copy of the code.

demmer:
Remove the moment-parser implementation and update to use the external moment-parser library from NPM to translate moment expressions into moment or duration values.

This makes it so that Juttle no longer knows anything about the AST node types used internally within moment-parser.

Closes #26.


demmer:
Oops -- there are still a couple of failing tests so this isn't quite ready yet.

demmer:
OK this is finally ready for review. @dmajda / @bkutil please take a look.

@welch FYI in case you're interested.

lib/compiler/semantic.js
```diff
@@ -21,6 +23,8 @@ var SemanticPass = Base.extend({
 
     initialize: function(options) {
         this.now = options.now;
+        this.beginning = moment(-Infinity);
```
welch: et tu juttle?
per today's discussions over on moment-parser, this creates an invalid moment with NaN as its value.

lib/moment/juttle-moment.js
```diff
@@ -844,6 +788,66 @@ var JuttleMoment = Base.extend({
...
         return operand.valueOf();
+    },
+    momentString: function(moment) {
+        if (isNaN(moment.valueOf())) {
+            return moment._i.toString();
+        } else {
+            return moment.toDate().toISOString();
+        }
+    },
+    simpleDurationString: function(d) {
```
welch: around the time of shuttering, there was talk of removing this short form of duration string and only using the ASP.NET "TimeSpan" format (the thing that looks like a digital clock readout). People were often confused by this shorter form appearing in output.

dmajda: This got pre-empted by open sourcing work, I hope to revisit that topic soonish.

dmajda:
The code looks good overall. My +1 depends on the summary of grammar changes and (assuming these will be uncontroversial) related doc changes.

Also, the commits should be squashed (only 0.3.0 update, only one commit with test changes).

welch:
It was a narrow grammar change -- eliminating "last <duration>" as a synonym of "<duration> ago", because it sounded like one was specifying a range of time rather than a single moment.

demmer:
@welch thanks for clarifying -- I had a chat with @dmajda over on the juttle gitter room a while ago and he agreed this is a sensible thing to do.

lib/compiler/semantic.js
demmer: This is consistent with the current behavior in which we process `beginning` by calling `new JuttleMoment(-Infinity)` which in turn calls `moment(-Infinity)`.

As we discussed separately, there's a potential argument to make that we shouldn't use infinite moments for these expressions, but my goal for this PR was not to make semantic changes to the language but simply to clean up the code and reference the separate library, so I don't want to make the behavior change yet.

lib/moment/juttle-moment.js
demmer: That seems totally sensible to me as well -- do you know if we have an issue tracking it in the old Jira? I don't see anything in github.

dmajda: PROD-8585, there is also a PR by @welch in our old repo referenced from there (could be used as a base but it has a couple of un-addressed comments).

demmer: OK cool -- once we get this PR merged we can revisit that.

demmer:
@dmajda I squashed / rebased / cleaned up as you suggested and then pushed another commit a326f16
that moves the moment stuff into the parser.

I agree that looks better - please take a look. Assuming it's ok then I would squash it with 619265b

dmajda:

Issue #31, opened 2015-12-23
CLI can’t be interrupted by Ctrl+C in some situations


dmehra:
Put the following into .juttle-config.json in your checkout:

```
{
    "backends": {
        "demo": {
            "path": "./lib/backends/demo-backend"
        }
    }
}
```

Run this command:
```
$ bin/juttle -e 'readx demo -from :beginning: -to :end:'
```
The program is running in an infinite loop now. Try to interrupt it using Ctrl+C and observe you can’t.

The expected result is that Ctrl+C works all the time.

When a job is running, ctrl-c should stop that job and return to the prompt. when a job is not running, ctrl-c should exit.

dmehra:
From @mstemm - this code reproduces the problem:
```
process.on('SIGINT', function() {
process.stderr.write("got signal SIGINT");
});

var readline = require('readline');

var rl = readline.createInterface({
input: process.stdin,
output: process.stdout
});

rl.on('line', function(data) {
console.log("GOT LINE", data);
var i;
for(;;) {
i++;
}
});

rl.on('SIGINT', function(data) {
console.log("GOT SIGINT", data);
});

rl.prompt("foo>");
```

dmehra:
From @mstemm - This is caused by a few things:

1. It looks like the stochastic source has an infinite loop somewhere. I see it entering but not leaving the call to step() at https://github.com/juttle/juttle/blob/master/lib/adapters/stochastic-adapter.js#L158.

2. When bin/juttle runs in repl mode (i.e. without -e or without any filename), the underlying node readline module sets standard input and standard output to "raw" mode so it can do things like read a single character at a time for looping through command history, redraw the screen, etc.

2a. When standard input is in raw mode, the readline module will not be sent SIGINT immediately when you ctrl-c. Instead, it reads standard input itself and when it reads a ctrl-c sends itself SIGINT.

Because of 1, 2a never gets a chance to read the ctrl-c, and can't convert the ctrl-c to a SIGINT.

One thing that got a little bit better since this card was originally filed (end of Nov) is that the readline use (and raw stdin change) only occurs when you're in repl mode now.

mstemm:
Aside from that, there's an additional problem for well-behaved programs. Calling program.deactivate does not properly tear down the procs in the function. The necessary changes are the following (at least):
- Juttle.proc.clientview (or Juttle.proc.clientsink, depending on when you look) needs to implement teardown() and call eof()
- Juttle.proc.source also needs to implement teardown().

[...]

PR #89, opened 2016-01-03
stochastic adapter: fix infinite loop when run with -from :beginning:

demmer:
Since :beginning: is actually an infinite moment, stepping forward ended up
causing an infinite loop.

Add a fix for this by explicitly checking for `-from :beginning:` and
implicitly converting it to start at the epoch instead.

Partially addresses #31.

### DIFF
--- a/lib/adapters/stochastic-adapter.js
+++ b/lib/adapters/stochastic-adapter.js
@@ -45,6 +45,9 @@ var Read = Juttle.proc.source.extend({
                 value: values.inspect(this.from)
             });
         }
+        if (this.from.isBeginning()) {
+            this.from = new JuttleMoment(0);
+        }
         this.historic = this.from.lt(NOW);
 
         this.to = options.to || FOREVER;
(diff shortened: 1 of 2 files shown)


demmer:
@dmehra or @mstemm thoughts on this?

mstemm:
lgtm!

dmajda:
Late to the party but wondering whether it wouldn’t be better to disallow `:beginning:` here.

@bkutil What do you think?

bkutil:
Considering that `JuttleMoment` accepts negative arguments and `0` doesn't actually correspond to ':beginning:' (only to a certain interpretation of it), the choice seems kind of arbitrary.

If we'd want this to emit points, I'd suggest putting here a `JuttleMoment` w/ the smallest date parseable by `Date` - a beginning in the JS sense. Maybe `JuttleMoment` should itself define that - `-8640000000000000` as a constant.

Otherwise, like @dmajda suggests, disallow `:beginning:` and just throw an error.

PR #91, opened 2016-01-04
table: look at all points in the batch to determine column widths

### LINKED ISSUE #33: CLI table columns are often too narrow
The CLI sets column width for `view table` based on the name/value of the first field it sees, meaning it crops values in the later rows that are wider. 

The workaround is to set `-progressive=false` option, but that seems too onerous for the user; perhaps the fix is to make that the default, if we know the program is historical not live?

Simple example:

```
juttle -e "read file -file '/tmp/github_data_from_es.json' | split | reduce by name"
```
Output sets width from "name" and "type", severely trimming later values like "actor_login":
```

┌──────────┐
│ name     │
├──────────┤
│ type     │
├──────────┤
│ actor_l… │
├──────────┤
│ repo_na… │
├──────────┤
│ n_commi… │
└──────────┘
```
The data in the file is like this and we should be able to show full values like actor_login in the table:

```
[
    {
        "type": "IssuesEvent",
        "actor_login": "smarterclayton",
        "repo_name": "GoogleCloudPlatform/kubernetes",
        "time": "2015-01-04T03:03:37.000Z"
    },
    {
        "type": "ForkEvent",
        "actor_login": "miksmile",
        "repo_name": "puppetlabs/puppet",
        "time": "2015-01-04T03:18:18.000Z"
    }
]
```

demmer:
In progressive mode, look at all the points in the first call to
consume when determining the set of columns and widths instead of
only looking at the first point.

Fixes #33.

### DIFF
--- a/lib/views/table.js
+++ b/lib/views/table.js
@@ -111,57 +111,64 @@ var TableView = View.extend({
     consume: function(data) {
         var self = this;
 
-        // If in progressive output, print a tablelet for the column
-        // headers and first point, or a tablelet for the points.
+        if (self.options.limit && self.items_written >= self.options.limit) {
+            return;
+        }
+
+        data = utils.fromNative(data.map(_.clone));
+
+        // In progressive mode, look at the first batch of points to detect the
+        // columns and determine their widths based on the length of the fields.
+        // Then print a tablelet for each batch of points, using the same widths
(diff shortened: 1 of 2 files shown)


demmer:
@mstemm

mstemm:
lgtm.

PR #94, opened 2016-01-04
stochastic: bump up the syslog max limit

### LINKED ISSUE #39: `read stochastic` emits throttling warning for too many logs
```
juttle docs/examples/backends/demo_cdn_historic.jutlle
```
used to result in:
```
{ [RuntimeError: Warning: too many logs from sea.0, throttling]
  message: 'Warning: too many logs from sea.0, throttling',
  code: 'RT-TOO-MANY-LOGS-WARNING',
  info:
   { host: 'sea.0',
     location: { filename: 'main', start: [Object], end: [Object] },
     procName: 'readx-demo' } } 'Warning: too many logs from sea.0, throttling'
```

I redid the docs example to avoid this, but without specifying a filter, I had to squeeze the time interval to 30 minutes to avoid the warning, above that it's "too many logs".

The workaround is to specify filter expressions in the read:

```
read stochastic -source 'cdn' -from :1 week ago: -to :now: -source_type 'metric'
name = 'requests' OR name = 'cpu' OR name = 'response_ms' OR name = 'disk'
```

This also avoids the big data too big issue. Stlil, it's not obvious to the user how to do that, I wish the source was generating enough data but not so much that naive use of it lands you in BDTB territory.

demmer:
A naive invocation of `read stochastic -source 'cdn' -last :1w:` would
trigger the limit of 1000 syslog messages.

Bump up the limit to 100000 to make this less likely to occur.

Fixes #39.

### DIFF
--- a/lib/runtime/procs/stoke/cdn.js
+++ b/lib/runtime/procs/stoke/cdn.js
@@ -74,7 +74,7 @@ var _config = {
     syslog_lpm: 0.1, // avg syslog error lines per minute per host
     syslog_thresh: 0.7, // demand at which message rate increases
     syslog_max_lpm: 20, // additional syslog error lines above thresh
-    syslog_max_lines: 1000, // max loglines produced before throttling
+    syslog_max_lines: 100000, // max loglines produced before throttling
     response_ms_normal_mean: 80,
     response_ms_heavy_mean: 2000,
     response_ms_thresh: 0.5,


demmer:
@dmehra @dmajda @bkutil

bkutil:
Looks good. Are we still safe w/ 100k points fitting into memory?

demmer:
Yeah I think 100k points is fine.

PR #104, opened 2016-01-04
Adding contributing guidelines

### LINKED ISSUE #99: Add CONTRIBUTING.md
Add guidelines for contributing to our project, starting with instructions for using the right github labels so the Waffle workflow is sane.

dmehra:
resolves #99

I used [puppet](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md) and [atom](https://github.com/atom/atom/blob/master/CONTRIBUTING.md) docs as examples. The filename [CONTRIBUTING.md](https://github.com/blog/1184-contributing-guidelines) is standard.

We are missing something like a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) or Certificate of origin like [nodejs](https://nodejs.org/en/get-involved/development/#developers-certificate-of-origin-10). It's not clear to me how critical is to have one, and must it be explicitly signed (using a tool like [clahub](https://www.clahub.com/) perhaps?) or is it enough to mention that by contributing, you allow to abide by the CLA.

@mccanne @demmer @dmajda

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,178 @@
+# How To Contribute
+
+Thank you for taking the time to contribute to project Juttle!
+
+The aim of this document is to make the process easy. These guidelines apply to new contributors as well as the regular project maintainers.
+
+If you have questions, feel free to ask in our [Gitter chat room](https://gitter.im/juttle/juttle).
+
+[TOC]
+
+## Project Organization
+
+We use GitHub issues to track bugs and enhancement requests. All you need to submit an issue is a GitHub user login. 
+
+We accept GitHub Pull Requests (PRs) with code contributions.
+
+Our project workflow is organized as a public [Waffle dashboard](https://waffle.io/juttle/juttle) that displays both GitHub issues and PRs.
+
+Juttle is composed of multiple repositories. To submit bugs, enhancement requests, and code contributions in the right place, use this logical grouping:
+
+- juttle - core language and CLI
+- outrigger - dev environment with visualizations in the browser
+- juttle-viz - charting library
+- moment-parser - time handling library
+- backend adapters:
+  - juttle-elastic-adapter
+  - juttle-influx-adapter
+  - juttle-gmail-adapter
+  - juttle-mysql-adapter
+  - juttle-postgres-adapter
+  - juttle-sqlite-adapter
+  - juttle-twitter-adapter
+  - etc.
+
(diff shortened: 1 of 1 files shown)


demmer:
Looks like a great start to me.

dmajda:
TL;DR. Seriously.

I totally appreciate the effort that went into this, but the contribution guidelines need to be really short, otherwise they will discourage contributions and people wouldn’t read them before opening an issue or a PR, which defeats their purpose.

I think the main goals of the guidelines are:

  * Help users to create meaningful bug reports and feature requests.
  * Help developers to create good PRs.
  * Direct anyone who needs a question answered or needs to get in touch with developers to the right place.

Ideal guidelines would be a minimal document satisfying these goals. But the suggested guidelines are more of a comprehensive overview of our organization, development process, etc. Writing these things down is important, but IMO they should not be part of the guidelines. I’d much rather see something short & sweet, like [Nette guidelines](https://github.com/nette/http/blob/master/contributing.md). Another example is [YaST contribution guidelines](https://github.com/yast/yast-yast2/blob/master/CONTRIBUTING.md) I wrote some time ago (but there I would do several things differently today).

So, what about splitting some parts out of the guidelines as separate documents, putting them into the wiki, and just linking them? Beside shortening the guidelines, this would lead to less duplication (and later divergence) between `CONTRIBUTING.md` files in our various repos.

I’d separate at least these documents:

  * Project organization (mainly the repo descriptions)
  * Development workflow (Waffle, issue lifecycle and labels used)

And I’d structure the remaining content as:

  * Asking questions
  * Reporting bugs
  * Requesting enhacements
  * Contributing code

I’d make these as short as possible (by moving all that can be moved to Development workflow and writing the content from the “minimum the external user needs to know to do X” perspective). Than decide whether these should stay in the guidelines or also be separated.

In any case, the result should be a document where the user who wants e.g. to file a bug quickly finds the relevant section, can read it quickly, and so the right thing. If interested, he can follow the links and read further on how we do things more generally.

@dmehra @demmer Thoughts?

dmehra:
Simplified the doc and took out project organization info to the wiki, https://github.com/juttle/juttle/wiki

I'm going to commit this, otherwise the links back from wiki to CONTRIBUTING.md would 404, but happy to do another pass or have you @dmajda make changes as you see fit.

dmehra:
Eek `[TOC]` is not a thing in Github markdown... i guess that only works with mkdocs-published documentation. Will fix.

dmajda:
Thanks! I’ll have another look but not an immediate priority.

PR #105, opened 2016-01-05
Fix multiline inputs and history.

### LINKED ISSUE #80: CLI in multiline mode adds `juttle>` on backspace
When using juttle CLI in multiline mode, if you use backspace to delete a wrong character, the prompt `juttle>` shows up at the front of that line. It is harmless in the sense that the program will execute normally, but it's confusing to have it suddenly appear, and makes it hard for copy/pasting of the program (you then have to edit it to get rid of the prompt in the middle).

To reproduce, try:
```
juttle> <
Starting multi-line input. End input with a line containing a single ".":
read file -file '/tmp/github_data_from_es.json'
| reduce by
```
Oops didn't mean to type "by" just yet, backspace to delete it, and look:
```
juttle> <
Starting multi-line input. End input with a line containing a single ".":
read file -file '/tmp/github_data_from_es.json'
juttle> | reduce b
```
Ok let's keep editing and run the program:
```
juttle> <
Starting multi-line input. End input with a line containing a single ".":
read file -file '/tmp/github_data_from_es.json'
juttle> | reduce count()
.
┌──────────┐
│ count    │
├──────────┤
│ 1597     │
└──────────┘
```
See it executed fine. But of course copying that code and trying to re-run it will not work, I'd have to edit out the prompt, otherwise it's an error:
```
juttle> <
Starting multi-line input. End input with a line containing a single ".":
read file -file '/tmp/github_data_from_es.json'
juttle> | reduce count()
.
<input>:2:7: 
   1:read file -file '/tmp/github_data_from_es.json'
   2:juttle> | reduce count()
           ^
Expected ";" or "|" but ">" found. (JUTTLE-SYNTAX-ERROR-WITH-EXPECTED)
```
Most amusingly, trying to edit it still in the juttle CLI mode will let me delete this prompt, but at the cost of having a new prompt appear, after which the code in the shell looks just like the above, but runs without error.

### LINKED ISSUE #81: CLI does not re-run multiline programs from history
When using juttle CLI in multiline mode, after you've run a program, if you wish to run it again by using the up arrow to bring it up from history, it won't run because the "start of multiline mode" character `<` is not retrieved. (It doesn't matter if the multiline program is last in history, or further up.)

```
juttle> <
Starting multi-line input. End input with a line containing a single ".":
read file -file '/tmp/github_data_from_es.json'
| reduce count()
.
┌──────────┐
│ count    │
├──────────┤
│ 1597     │
└──────────┘
juttle> read file -file '/tmp/github_data_from_es.json'
| reduce count()
.

juttle> 
```
After you enter the `.` to run the multiline program, nothing happens, it just sits there. When you hit Enter, it returns to the `juttle>` prompt. 

Can we have it bring up the multiline program including the `<` so it will run, like this:

```
juttle> <
read file -file '/tmp/github_data_from_es.json'
| reduce count()
.
```
(The visual appearance of `<` is optional, as long as the program executes after `.`).

mstemm:
To fix #81, include 2 changes that both address the problem in different
ways. I think I'll keep both fixes just to allow for flexibility in
case we want to change the behavior later.

The first fix involves collapsing multi-line programs to a single line
when saving to history. That way, scrolling through history doesn't
result in an ugly multi-line item, and also allows for editing of the
item. When the multi-line program is run for the first time, however,
it's still run as a multi-line program. That allows any errors to be
shown in the context of the original multi-line program.

The second fix is to allow multiline inputs to work with history. The
problem was that the '.*' regex wasn't matching multiple lines
separated by newlines. Javascript regex apparently doesn't have a
multi-line option when compiling the regex, so just look for newlines
explicitly.

To fix #80, while in the middle of a multi-line input, set the prompt to an empty
string so it won't appear if you use editing keys.

This fixes #80 and #81.

@dmehra

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -335,7 +335,7 @@ function setupCline() {
         trim_history();
     });
 
-    cli.command('{juttle}', 'evaluate juttle program', {juttle: '(?!(help|clear|exit)|\\.).*'}, function(input, args) {
+    cli.command('{juttle}', 'evaluate juttle program', {juttle: '(?!(help|clear|exit)|\\.).*(?:\n.*)*'}, function(input, args) {
         if (in_multi_line) {
             multi_line_src += input + '\n';
         } else {
@@ -351,7 +351,14 @@ function setupCline() {
             cli.prompt(prompt);
         } else {
             in_multi_line = false;
-            add_history(multi_line_src);
+
+            // Remove the newlines from the command before adding it
+            // to history. This will make it easier to view and edit
+            // after scrolling back through history. When running the
+            // command, however, still keep the multi-line version so
+            // any initial errors are printed in context.
+
+            add_history(multi_line_src.replace(/\n/g, ' '));
             perform_mode({
                 action: mode,
                 juttle_src: multi_line_src
--- a/bin/juttle
+++ b/bin/juttle
@@ -332,6 +332,7 @@ function setupCline() {
         console.log('Starting multi-line input. End input with a line containing a single ".":');
         multi_line_src = '';
         in_multi_line = true;
+        cli.prompt('');
         trim_history();
     });
 


mstemm:
bump. Also adding @demmer @rlgomes to increase my chances of a lgtm :)

dmehra:
Tried this out, #80 is certainly fixed. 

The history feature is not ideal because you cannot bring back a multiline program from history and still edit it in a reasonable way. This is a fairly small program:

```
(
emit -points [ { user: "u1", dept: "d1" }, {user: "u2", dept: "d2"}, {user: "u3", dept: "d3"} ];
emit -points [ { dept: "d1", mgr: "m1" }, { dept: "d3", mgr: "m3" }, {dept: "d4", mgr: "m4" } ];
)
| join -outer 1 dept
```

When retrieved from history it's like this:

```
juttle> ( emit -points [ { user: "u1", dept: "d1" }, {user: "u2", dept: "d2"}, {user: "u3", dept: "d3"} ]; emit -points [ { dept: "d1", mgr: "m1" }, { dept: "d3", mgr: "m3" }, {dept: "d4", mgr: "m4" } ]; ) | join -outer 1 dept 
```

so while you can still run it, editing it would be difficult. 

If showing programs in multiline mode from history is hard to implement, this is perhaps an acceptable limitation.

mstemm:
You're right that you can't edit a multi-line program from history, but you couldn't do that before at all, other than possibly editing the last line of the program. With this change you have some ability to edit the program if you really need to.

If you have a multi-line program and want to edit it, you should switch to using an editor and the ``src`` command.

demmer:
lgtm.

PR #106, opened 2016-01-05
Scheduler support for periodic tasks

bkutil:
Most of our backends need to repeatedly execute a periodic fetch for the pseudo-live mode. Update the scheduler to support periodic task execution.

### DIFF
--- a/lib/runtime/scheduler.js
+++ b/lib/runtime/scheduler.js
@@ -1,11 +1,14 @@
 var Heap = require('heap');
 var Base = require('extendable-base');
+var errors = require('../errors');
 
 var Scheduler = Base.extend({
     initialize: function(options) {
         this.timer = null;
         this.queue = new Heap(function(a, b) { return a.ms - b.ms; });
+
         this.running = false;
+        this.stopped = false;
     },
 
     now: function() {
@@ -13,6 +16,10 @@ var Scheduler = Base.extend({
     },
 
     schedule: function(time, callback) {
+        if (this.stopped === true) {
+            throw errors.runtimeError('RT-CANNOT-SCHEDULE-AFTER-STOP');
+        }
+
         this.queue.push({ms:time, callback:callback});
 
         if (this.running === true) {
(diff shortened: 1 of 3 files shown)


bkutil:
cc: @demmer @dmajda @rlgomes

rlgomes:
+1 changes look good.

lib/runtime/scheduler.js
```diff
@@ -1,10 +1,14 @@
...
 var Base = require('extendable-base');
+var errors = require('../errors');
 
 var Scheduler = Base.extend({
     initialize: function(options) {
         this.timer = null;
         this.queue = new Heap(function(a, b) { return a.ms - b.ms; });
+
+        this.running = false;
+        this.stopped = false;
```
demmer: It's not immediately obvious why you have two booleans that at first glance imply the same thing -- upon reading the code it's clear that the thing you're trying to effect is that someone can schedule before start is called but not after stop is called, i.e. there is a difference between `running` and `! stopped`.

Perhaps this would be clearer if you used a `state` variable that could be one of `INIT`, `RUNNING` or `STOPPED`?

lib/runtime/scheduler.js
```diff
@@ -12,33 +16,72 @@ var Scheduler = Base.extend({
...
+        var self = this;
+
+        if (this.stopped === true) {
+            throw errors.runtimeError('RT-CANNOT-SCHEDULE-AFTER-STOP');
+        }
+
+        var job = function() {
+            if (self.running === true) {
+                callback();
+                self.schedule(Date.now() + every, job);
```
demmer: Seems like there would be an error if the execution of the callback caused stop to be called. Perhaps we want to have a second check for `self.running === true` before calling schedule again?

demmer:
Looks good to me.

bkutil:
Introduced a state variable as @demmer suggested and rebased.

bkutil:
I'd like to merge this when emit/a non-trivial backend uses this code. I'm working now on the nested points, so I'll close this for now, then re-open when the other task is finished.

PR #108, opened 2016-01-05
Add error for no -from/-to/-last.

mstemm:
Add a new error RT-MISSING-TIMERANGE-ERROR to handle the case where no
-from, -to, or -last is specified. This reflects the first example in
https://github.com/juttle/juttle/wiki/Time-Range-Semantics.

@dmajda @demmer @go-oleg

### DIFF
--- a/lib/strings/juttle-error-strings-en-US.json
+++ b/lib/strings/juttle-error-strings-en-US.json
@@ -33,6 +33,7 @@
     "RT-REDUCE-DELTA-ERROR": "Error: delta cannot be used with reduce (use reduce last | put delta=delta())",
     "RT-FROM-TO-MOMENT-ERROR": "Error: -from/-to/-last require moments: {{info.value}}",
     "RT-LAST-FROM-TO-ERROR": "Error: -last option should not be combined with -from or -to",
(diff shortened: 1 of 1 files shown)


dmajda:
+1 with the comment addressed

dmehra:
Wait up @dmajda what about `read file` where intuitively, without any time limits, you should read the entire file? That's how it currently works and it seems to me, this new check will break that behavior

demmer:
More generally what about sources that may not have a time field like sql, http, stdin, etc?

dmajda:
Good point. This probably needs more thought but it seems to me we really have two classes of backends:

  * **Static, small-data ones (`file`, `sql`, `http`,…).** For them, it makes sense to allow no `-from`/`-to`/`-last` and have `-from :beginning: -to :end:` as defaults.
  * **Dynamic, big-data ones (`graphite`, `influxdb`,…).** For them, the [time range semantics](https://github.com/juttle/juttle/wiki/Time-Range-Semantics) I wrote down yesterday makes sense.

The main question is whether we should really have different semantics for different adapters (I’d say it’s inevitable) and also whether some don’t fall into both classes depending on the usage (hard to tell for me).

As for timeless points, I think these can generally be treated as always falling into the range defined by `-from` & `-to`.

Issue #41, opened 2015-12-23
Sink undefined on emitting points to table


dmehra:
From @bkutil:
```
juttle> emit -points [{"key":"val"},{"key2":"val2"}] | view table
┌────────────────────────────────────┬──────────┐
│ time                               │ key      │
├────────────────────────────────────┼──────────┤
│ 2015-12-23T18:17:51.195Z           │ val      │
undefined 'Sink undefined: Ignoring new point key(s) "key2"'
├────────────────────────────────────┼──────────┤
│ 2015-12-23T18:17:52.195Z           │          │
└────────────────────────────────────┴──────────┘
```

The problem here is that 2nd point has a different key than 1st point, and our table does not accommodate that by default. The workaround is

```
juttle> emit -points [{"key":"val"},{"key2":"val2"}] | view table -progressive false
┌──────────────────────────┬─────┬──────┐
│ time                     │ key │ key2 │
├──────────────────────────┼─────┼──────┤
│ 2015-12-23T18:18:47.958Z │ val │      │
├──────────────────────────┼─────┼──────┤
│ 2015-12-23T18:18:48.958Z │     │ val2 │
└──────────────────────────┴─────┴──────┘
```

This is similar to https://github.com/juttle/juttle/issues/33 in that making -progressive false the default (for historical programs) would remove the error, and allow the user to see all data in the output.

demmer:
The changes in #91 are a first step towards addressing this, but as I mentioned in the comment on #33, we'd need to add a bit more logic to delay the initial table rendering in order for more than one batch of points to arrive.

demmer:
There's also the other obvious issue that we shouldn't be printing `undefined` in the error messages.

dmehra:
Both referenced issues #91 and #33 are closed, and we no longer emit "undefined" in this case, but the main issue of not seeing all fields remains:

```
juttle> emit -points [{"key":"val"},{"key2":"val2"}] | view table
┌──────────┐
│ key      │
├──────────┤
│ val      │
Warning: view table: Ignoring new point key(s) "key2"
├──────────┤
│          │
└──────────┘
```

@demmer were your recommendations to look at the whole batch not implemented? or did not cover `emit -points`?

demmer:
This is fixed now.

PR #112, opened 2016-01-05
cli: fix 'undefined' in view warning output

demmer:
Previously if a warning was emitted by a view, the message would be
output as something like:

    undefined 'Sink undefined: Ignoring new point key(s) "key2"'

Fix the two `undefined` above by tiding the way the CLI prints out
warning events and by adding a static name property to each view
instead of expecting it to be passed into the options in the constructor.

Partially addresses #41

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -175,7 +175,7 @@ function perform_run(options) {
         });
 
         program.events.on('warning', function(msg, warn) {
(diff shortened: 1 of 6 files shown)


demmer:
@mstemm

mstemm:
lgtm, and thanks for adding the test!

PR #115, opened 2016-01-06
finish renaming sink to view

demmer:
Another pass through the compiler to replace instances of 'sink' with 'view' as appropriate.
Details in the specific commits.

Closes #24 

Note that outrigger will need to be updated once this merges.

### DIFF
--- a/lib/compiler/ast-visitor.js
+++ b/lib/compiler/ast-visitor.js
@@ -84,8 +84,8 @@ var NODE_CHILDREN = {
     ByList:                   ['elements'],
     SortByList:               ['elements'],
     SortField:                ['expr'],
-    Sink:                     ['options'],
-    SinkOption:               ['expr'],
+    View:                     ['options'],
+    ViewOption:               ['expr'],
     InputOption:              ['expr'],
 
     /* Statements */
--- a/lib/compiler/autocompleter.js
+++ b/lib/compiler/autocompleter.js
@@ -38,8 +38,8 @@ var Autocompleter = Base.extend({
     initialize: function(options) {
         options = options || {};
 
-        var sinks = options.sinks || {};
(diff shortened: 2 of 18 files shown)


demmer:
@dmajda @bkutil @mstemm

mstemm:
These changes look fine but I think @dmajda/@bkutil should also take a look.

My grep for sink is still finding lots of instances, but it looks like those are for the base class and as a result are appropriate.

demmer:
Yep. As far as I can tell all remaining instances of sink are proper. 

-m

dmajda:
Grepping revealed there is a `sinks` variable [in `build.js`](https://github.com/juttle/juttle/blob/afd5031dee1a3246258f51adc460d4e66a5a3fd1/lib/compiler/build.js#L77) which looks unused. I’m fine with both removing it now or later (together with flattening the object returned from the builder function so that there is no `graph` property).

dmajda:
I think [this comment](https://github.com/juttle/juttle/blob/afd5031dee1a3246258f51adc460d4e66a5a3fd1/lib/compiler/graph-compiler.js#L463) should be updated too.

dmajda:
Hmm, [this comment](https://github.com/juttle/juttle/blob/afd5031dee1a3246258f51adc460d4e66a5a3fd1/lib/parser/parser.pegjs#L1182) is no longer true (will fix separately).

dmajda:
Other than this, I think all remaining uses of “sink” are legit. There is a couple of pieces of Juttle code like `... | view sink1` in the tests, which I think could be changed too, but not necessarily. Similarly with few “sink” instances [here](https://github.com/juttle/juttle/blob/afd5031dee1a3246258f51adc460d4e66a5a3fd1/docs/doc_templates/usage_snippets.md).

+1

dmajda:
OK, I fixed small semi-related issues separately.

PR #116, opened 2016-01-06
enable filter to operate on multiple fields

demmer:
This just a work-in-progress to enable comparisons between fields
as requested by @mccanne.

Remove the semantic pass checks that ensure that filter expressions
only ever compare a field to a compile-time expression.

Add an option `allowFieldComparisons` to the filter-checker that
allows an expression of the form `field1 == field2`.


demmer:
Properly implemented in #122

PR #119, opened 2016-01-06
Build: Do not return anything from generated build function

dmajda:
The result was unused. Everything important is collected in `GraphBuilder`.

Discovered during review of #115.


dmajda:
@demmer

Issue #55, opened 2015-12-23
Support nested points in Juttle


dmehra:
- [x] correctly handle nested points when grouping
- [x] correctly handle nested points in `sort`
- [x] correctly handle nested points in all backends
- [x] correctly handle nested points in views
- [x] correctly handle nested points in all remaining procs

dmehra:
Details for each task:

### correctly handle nested points when grouping
Support correct grouping by fields containing arrays or objects. This means that:
- Group keys should be tested for equality using the relation defined in values.js.
- Group keys which are arrays or objects should not be stringified when emitted as part of emitted points.

### correctly handle nested points in sort
Support sorting by fields containing arrays or objects. This means field values should be compared using relations defined in {{ops.js}}.

### correctly handle nested points in all backends

### correctly handle nested points in views
Make sure that views handle points with fields containing arrays or objects correctly. More specifically, this means:
- Field values should be stringified using the function defined in values.js.
- Wherever a field value is used for equality testing, comparison, sorting, etc. it should be either handled using functions defined in values.js/ops.js or unhandled value types should be rejected with understandable error message. This includes e.g. values passed in -keyField of incoming points.

### correctly handle nested points in all remaining procs
In no proc should a field containing array or object cause an internal error or incorrect behavior. In general, values should be handled using functions defined in values.js/ops.js (which may throw type errors in certain situations — this is fine). Where this is impossible, unhandled value types should be rejected with understandable error message.
This is a "catch-all" issue for all cases not covered by the above.

dmajda:
> @bkutil does this mean we will support writes of nested points too?

Yes, where possible.

---

I think it makes sense to split this issue according to the tasks (they are themselves pretty big) and maybe keep this one for overall tracking (as an “epic” in scrum terminology). Leaving this to @bkutil.

bkutil:
- [ ] nested point with array value
- [ ] nested point with list of objects value (that's supported right?)
- [ ] reduce by two non-scalar fields
- [ ] other "by" invocations such as head by
- [ ] sort on non-scalar field
- [ ] join on non-scalar field
- [ ] pluck non-scalar field
- [ ] unique on non-scalar field
- [ ] the four above called without parameters on data containing non-scalar

bkutil:
Handle nested points in core views:

- [x] file
- [x] text
- [x] table

Views for juttle-viz are tracked in a separate issue: https://github.com/juttle/juttle-viz/issues/17

bkutil:
[...]

PR #121, opened 2016-01-06
Reduce on nested

bkutil:
Add support for grouping on and unique-counting of 'nested points', i.e. points whose fields can be objects or arrays.

This relies on serializing such fields using `values.inspect` into an unique string representation. This is not optimal from the memory consumption perspective, but should be OK as a first version.

Also, as the `values.valueOf` is now unused, remove it from the codebase.

refs: https://github.com/juttle/juttle/issues/55

### DIFF
--- a/lib/runtime/groups.js
+++ b/lib/runtime/groups.js
@@ -144,7 +144,7 @@ var Groups = Base.extend({
         for (var colno = 0; colno < by.length - 1; ++colno) {
             var currentField = pt[by[colno]];
 
-            index = index[values.valueOf(currentField)];
+            index = index[values.inspect(currentField)];
 
             if (index === undefined) {
                 return undefined;
@@ -152,8 +152,9 @@ var Groups = Base.extend({
         }
 
         var lastField = pt[by[by.length - 1]];
-        return index[values.valueOf(lastField)];
+        return index[values.inspect(lastField)];
     },
+
     _insert_key: function(pt) {
         var index = this.index;
         var next_index;
@@ -163,16 +164,16 @@ var Groups = Base.extend({
         for (colno = 0; colno < n - 1; ++colno) {
             val = pt[by[colno]];
             vals[colno] = val;
-            next_index = index[values.valueOf(val)];
+            next_index = index[values.inspect(val)];
             if (next_index === undefined) {
-                index[val] = next_index = {};
+                index[values.inspect(val)] = next_index = {};
             }
             index = next_index;
         }
         keyID = this.nkeys++;
         val = pt[by[n - 1]];
         vals[n - 1] = val;
-        index[values.valueOf(val)] = keyID;
+        index[values.inspect(val)] = keyID;
         this.key_vals[keyID] = vals;
         return keyID;
     },
--- a/lib/runtime/reducers.js
+++ b/lib/runtime/reducers.js
@@ -418,7 +418,7 @@ reducers.count_unique = {
             },
             update: function(pt) {
                 if (pt[field] !== undefined) {
-                    uniques[values.valueOf(pt[field])] = true;
+                    uniques[values.inspect(pt[field])] = true;
                 }
             }
         };
--- a/lib/runtime/values.js
+++ b/lib/runtime/values.js
@@ -68,10 +68,6 @@ var values = {
         throw new Error('Invalid Juttle value: ' + value + '.');
     },
 
-    valueOf: function(v) {
-        return (values.isDate(v) || values.isDuration(v)) ? v.valueOf() : v;
-    },
-
     // Returns a display name for a Juttle type name (for use in error messages,
     // etc.).
     typeDisplayName: function(name) {
--- a/test/runtime/procs.spec.js
+++ b/test/runtime/procs.spec.js
@@ -1179,6 +1179,43 @@ describe('Juttle procs tests', function() {
                 });
         });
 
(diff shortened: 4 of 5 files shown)


bkutil:
cc: @dmajda @demmer

dmehra:
Great enhancement that deserves more test cases, these ones come to mind:

- nested point with array value 
- nested point with list of objects value (that's supported right?)
- reduce by two non-scalar fields
- other "by" invocations such as head by
- sort on non-scalar field
- join on non-scalar field
- pluck non-scalar field
- unique on non-scalar field
- the four above called without parameters on data containing non-scalar 

How does it work to reference a key inside the nested point? Is it legal to `reduce by a[b]`? If so, also needs tests.

Is multi-level nesting supported? If yes, test please. If no, negative test (where would it fail - as the invalid point enters the flow graph?)

bkutil:
- [ ] nested point with array value
- [ ] nested point with list of objects value (that's supported right?)
- [ ] reduce by two non-scalar fields
- [ ] other "by" invocations such as head by
- [ ] sort on non-scalar field
- [ ] join on non-scalar field
- [ ] pluck non-scalar field
- [ ] unique on non-scalar field
- [ ] the four above called without parameters on data containing non-scalar

dmajda:
+1 for the code, but see below about tests

@dmehra As for the tests, I think we should generally agree on the approach on proc testing. So far I assumed that our tests are white-box unit tests, that is, they “see” into how things are implemented and test individual components. This approach applied here would mean:

  * There should be comprehensive unit test of the serialization mechanism used (`values.inspect`).

  * There should be exactly one test for each proc in which we deal with nested points that verifies we handle them (we use the serialization mechanism). No need to test individual cases (object, array) or their combinations.

    We may actually do this test not at the proc level, but as a unit test of the `Groups` class. The proc tests would then only verify that by-grouping works (we use the `Groups` class) using some trivial case.

Another approach would be to treat proc testing as a black-box. Then we should not assume anything about the implementation, which in this particular case means exercising at least these cases for each proc dealing with nested points:

  * array
  * object
  * some more complex nesting (e.g. array in object + object in array)

I personally can work with both approaches, but I’d like us to choose one consciously. I think the white-box approach will result in better test economy and less duplication, but it will be harder to maintain the tests and it requires more thought about what to test where. The black-box approach is the opposite.

@bkutil, @dmehra: Thoughts?

> - nested point with array value 
> - nested point with list of objects value (that's supported right?)

These should definitely be tested somewhere (where exactly depends on the approach).

> - reduce by two non-scalar fields

Doesn’t add any new coverage and thus not needed IMO.

> - other "by" invocations such as head by

Depends on approach.

> - sort on non-scalar field
> - join on non-scalar field
> - pluck non-scalar field
> - unique on non-scalar field
> - the four above called without parameters on data containing non-scalar 

Out of scope of this PR, but in scope of #55. Maybe move the check-list there, @bkutil?

> How does it work to reference a key inside the nested point?

You mean e.g. reducing by `x` in points like this?

```juttle
{
  coords: {
    x: 5
    y: 6
  }
}
```

Not doable right now. We may allow something like `reduce by coords.x` in the future, but I don’t think this is a high-priority.

> Is it legal to `reduce by a[b]`? If so, also needs tests.

It is legal but it does something different than you’d think :-) And it’s actually buggy. See #127 for details.

> Is multi-level nesting supported?

Yes.

bkutil:
I'd slightly favor the white-box testing approach - to test `groups` and `values.inspect` thoroughly, as most of the complexity is there, then add a trivial check to each of the procs using them.

From briefly looking at procs, it seems they don't add much extra logic on top of groups (which can be tested separately) and use them in an uniform way, so black box unit tests would just repeatedly exercise the same code in groups and not add much value beyond that, I think.

bkutil:
Merging based on +1 from David

rlgomes:
I do see that in some circumstances if module A is well tested and module A is used by B then do you really have to tests all of modules B's uses of A ? I would say its a judgement call on exactly how module B uses A and if it introduces anything in that path that needs to be tested. Now with that said the `values.js` file here being used has no spec file associated as far as I can see and has 90% coverage by what I can assume is indirect testing from other modules being called: 



So we are missing a `values.spec.js`  which would be the white box approach mentioned (although I don't like the use of the whitebox vs blackbox here as it feels out of place).

This situation is similar to the way I've done testing of the `lib/adapters/parsers` and `lib/adapter/serializers` as I've thoroughly tested each of those separately but I haven't written yet another layer of tests for the other modules (ie http, file and stdio) which use them since they're just using them in the same manner that was already tested by the tests dedicated to those modules. This was of course a judgement call given I knew how I was using the "module B" of this scenario.

Just my 2 cents.

dmajda:
> So we are missing a `values.spec.js`  which would be the white box approach mentioned (although I don't like the use of the whitebox vs blackbox here as it feels out of place).

Just to clarify what I meant here:

  * *white-box* — proc tests are structured with the knowledge that various `by` clauses behave the same because they share some underlying code (`values.js`, `Groups`), which allows to verify a lot of their behavior on that level (just once for all procs)
  * *black-box* — proc tests are structured without that knowledge, which means they have to re-test the behavior for each proc separately

Doesn’t feel out of place to me.

---

Looks like we all converge to the first approach (white-box), so let’s use that in general. I filed #139 for adding unit tests of `values.js`.

bkutil:
Ok, assigned myself to #139.

PR #126, opened 2016-01-07
grok parser

### LINKED ISSUE #37: create new grok based parser for log lines
Once we have the stdin adapter (#36) we should introduce an input format "grok" to parse raw log lines.

rlgomes:
fixes #37 

with this we can now parse log lines using grok from the file and stdio
adapter

### DIFF
--- a/lib/adapters/file/read.js
+++ b/lib/adapters/file/read.js
@@ -9,7 +9,7 @@ var Read = Juttle.proc.source.extend({
     procName: 'read-file',
 
     initialize: function(options, params, location, program) {
-        var allowed_options = ['from', 'to', 'file', 'timeField', 'format'];
+        var allowed_options = ['from', 'to', 'file', 'timeField', 'format', 'pattern'];
         var unknown = _.difference(_.keys(options), allowed_options);
         if (unknown.length > 0) {
             throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
@@ -28,7 +28,7 @@ var Read = Juttle.proc.source.extend({
         this.filename = options.file;
         this.timeField = options.timeField;
         this.format = options.format ? options.format : 'json'; // default to json
-        this.parser = parsers.getParser(this.format);
+        this.parser = parsers.getParser(this.format, options);
 
         this.from = options.from || new JuttleMoment(0);
         if (!this.from.moment) {
--- a/lib/adapters/parsers/base.js
+++ b/lib/adapters/parsers/base.js
@@ -12,6 +12,6 @@ module.exports = Base.inherits(events.EventEmitter, {
     },
 
     parseStream: function(stream, emit) {
-        // implement me
+        throw new Error('parseStream must be implemented');
     }
 });
--- a/lib/adapters/parsers/base.js
+++ b/lib/adapters/parsers/base.js
@@ -1,7 +1,17 @@
 var Base = require('extendable-base');
+var events = require('events');
 
-module.exports = Base.extend({
+module.exports = Base.inherits(events.EventEmitter, {
+    /*
+     * A parser must implement the parseStream function below and should fire
+     * errors by using the
+     * `this.emit('error', ...)` EvenEmitter method.
+     */
     initialize: function(options) {
         this.limit = (options && options.limit) ? options.limit : 1024;
+    },
+
+    parseStream: function(stream, emit) {
+        // implement me
     }
 });
--- a/lib/adapters/parsers/grok.js
+++ b/lib/adapters/parsers/grok.js
@@ -1,0 +1,73 @@
+var base = require('./base');
+var errors = require('../../errors');
+var grok = require('node-grok');
+var Promise = require('bluebird');
+var readline = require('readline');
+
+module.exports = base.extend({
+
+    initialize: function(options) {
+        this.pattern = options.pattern || '%{GREEDYDATA:message}';
+    },
+
+    parseStream: function(stream, emit) {
+        var self = this;
+
+        return new Promise(function(resolve, reject) {
+            var buffer = [];
+            var streamEnd = false;
+            var pending = 0;
+            var currentPromise = Promise.resolve();
+
+            function checkAndFlush() {
+                if (buffer.length >= self.limit) {
+                    emit(buffer);
+                    buffer = [];
+                }
+
+                if (streamEnd && pending === 0) {
+                    emit(buffer);
+                    buffer = [];
+                    resolve();
+                }
+            }
+
+            grok.loadDefault(function(patterns) {
+                var pattern = patterns.createPattern(self.pattern);
+                var parse = Promise.promisify(pattern.parse);
+
+                var liner = readline.createInterface({
+                    input: stream,
+                    terminal: false
+                });
+
+                liner.on('line', function(line) {
+                    pending++;
+                    currentPromise = currentPromise.then(function() {
+                        return parse(line)
+                        .then(function(result) {
+                            if (result[1]) {
+                                buffer.push(result[0]);
+                            } else {
+                                self.emit('error', errors.runtimeError('RT-INTERNAL-ERROR',{
+                                    error: 'Skipping line "' + line + '", due to: not matching pattern'
+                                }));
+                            }
+                            pending--;
+                            checkAndFlush();
+                        })
+                        .catch(function(err) { self.emit('error', errors.runtimeError('RT-INTERNAL-ERROR',{
+                                error: 'Skipping line "' + line + '", due to: ' + err.toString()
+                            }));
+                        });
+                    });
+                });
+
+                liner.on('close', function() {
+                    streamEnd = true;
+                    checkAndFlush();
+                });
+            });
+        });
+    }
+});
--- a/lib/adapters/parsers/grok.js
+++ b/lib/adapters/parsers/grok.js
@@ -15,20 +15,12 @@ module.exports = base.extend({
 
         return new Promise(function(resolve, reject) {
             var buffer = [];
-            var streamEnd = false;
-            var pending = 0;
             var currentPromise = Promise.resolve();
 
-            function checkAndFlush() {
+            function maybeFlush() {
                 if (buffer.length >= self.limit) {
                     emit(buffer);
                     buffer = [];
-                }
-
-                if (streamEnd && pending === 0) {
-                    emit(buffer);
-                    buffer = [];
-                    resolve();
                 }
             }
 
@@ -42,7 +34,6 @@ module.exports = base.extend({
                 });
 
                 liner.on('line', function(line) {
-                    pending++;
                     currentPromise = currentPromise.then(function() {
                         return parse(line)
                         .then(function(result) {
@@ -53,10 +44,10 @@ module.exports = base.extend({
                                     error: 'Skipping line "' + line + '", due to: not matching pattern'
                                 }));
                             }
-                            pending--;
-                            checkAndFlush();
+                            maybeFlush();
                         })
-                        .catch(function(err) { self.emit('error', errors.runtimeError('RT-INTERNAL-ERROR',{
+                        .catch(function(err) {
+                            self.emit('error', errors.runtimeError('RT-INTERNAL-ERROR',{
                                 error: 'Skipping line "' + line + '", due to: ' + err.toString()
                             }));
                         });
@@ -64,8 +55,10 @@ module.exports = base.extend({
                 });
 
                 liner.on('close', function() {
-                    streamEnd = true;
-                    checkAndFlush();
+                    currentPromise.then(function() {
+                        emit(buffer);
+                        resolve();
+                    });
                 });
             });
         });
--- a/lib/adapters/stdio/read.js
+++ b/lib/adapters/stdio/read.js
@@ -8,7 +8,7 @@ var Read = Juttle.proc.source.extend({
     procName: 'read stdio',
 
     initialize: function(options, params, location, program) {
-        var allowed_options = ['timeField', 'format', 'formatOptions'];
+        var allowed_options = ['timeField', 'format', 'pattern'];
         var unknown = _.difference(_.keys(options), allowed_options);
         if (unknown.length > 0) {
             throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
@@ -20,8 +20,7 @@ var Read = Juttle.proc.source.extend({
         this.timeField = options.timeField;
 
         this.format = options.format ? options.format : 'json'; // default to json
-        var formatOptions = options.formatOptions || {};
-        this.parser = parsers.getParser(this.format, formatOptions);
+        this.parser = parsers.getParser(this.format, options);
 
         if (params.filter_ast) {
             throw this.compile_error('RT-ADAPTER-UNSUPPORTED-FILTER', {
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "moment-duration-format": "^1.3.0",
     "moment-parser": "^0.3.0",
     "moment-timezone": "^0.4.0",
+    "node-grok": "github:beh01der/node-grok#b76dae6dc4efa745e6bf75a3cb8d113aa47bc16e",
     "oboe": "^2.1.2",
     "pegjs": "^0.9.0",
     "request": "^2.67.0",
--- a/test/adapters/file/file.spec.js
+++ b/test/adapters/file/file.spec.js
@@ -10,7 +10,7 @@ var expect_to_fail = juttle_test_utils.expect_to_fail;
 
 var tmp_file = tmp.tmpNameSync();
 
-var validFormats = {
+var symmetricalFormats = {
     json: 'json',
     jsonl: 'jsonl',
     csv: 'csv'
@@ -42,7 +42,7 @@ describe('file adapter tests', function () {
             });
         });
 
-        _.each(validFormats, function(format) {
+        _.each(symmetricalFormats, function(format) {
             describe(format, function() {
                 var file_name = file + '.' + format;
 
--- a/test/adapters/file/file.spec.js
+++ b/test/adapters/file/file.spec.js
@@ -31,6 +31,8 @@ describe('file adapter tests', function () {
     var file = path.resolve(__dirname, 'input/simple');
     var json_file = file + '.json';
 
+    var syslog = path.resolve(__dirname, '../parsers/input/logs/syslog');
+
     describe('read file', function() {
         it('fails when you provide an unknown option', function() {
             return run_read_file_juttle(tmp_file, {foo: 'bar'})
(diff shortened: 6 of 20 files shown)


rlgomes:
@demmer @dmehra @davidvgalbraith @go-oleg grok! so just seeking early feedback on the approach but basically you can do things like:

```
rlgomes@x230> cat /var/log/syslog | ./bin/juttle -e "read stdio -format 'grok' -formatOptions { pattern: '%{SYSLOGLINE}' } | view text"
[
{"timestamp":"Jan  6 10:54:52","logsource":"x230","program":"anacron","pid":"15134","message":"Job `cron.daily' terminated"},
{"timestamp":"Jan  6 10:54:52","logsource":"x230","program":"anacron","pid":"15134","message":"Normal exit (1 job run)"},
{"timestamp":"Jan  6 11:17:01","logsource":"x230","program":"CRON","pid":"17219","message":"(root) CMD (   cd / && run-parts --report /etc/cron.hourly)"},
{"timestamp":"Jan  6 11:17:01","logsource":"x230","program":"CRON","pid":"17218","message":"(root) MAIL (mailed 1 byte of output; but got status 0x00ff, #012)"},
{"timestamp":"Jan  6 12:17:01","logsource":"x230","program":"CRON","pid":"23971","message":"(root) CMD (   cd / && run-parts --report /etc/cron.hourly)"},
{"timestamp":"Jan  6 12:17:01","logsource":"x230","program":"CRON","pid":"23970","message":"(root) MAIL (mailed 1 byte of output; but got status 0x00ff, #012)"},
{"timestamp":"Jan  6 13:09:54","logsource":"x230","program":"kernel","message":"[548997.882923] i915 0000:00:02.0: BAR 6: [??? 0x00000000 flags 0x2] has bogus alignment"},
...
```

So let me know what you think and then I can follow up with documentation around the feature which cna further use examples to show how to parse well known logs: apache, java, etc.

dmehra:
Finally looked at this! It's awesome. I'd want the same support in `read file` because there is no reason why only stdio should handle grok; how would that work - do we need Yet Another Repo <super>TM</super>? 
Also, how complex do the rules tend to get? Would we need to support `-formatFile` so the grok rule can be edited in the file, as opposed to in the command line?

rlgomes:
@dmehra don't know if we want to expose a `-formatFile` or some other way of referencing rules from our `juttle-config.json`

rlgomes:
@demmer reading from a file already works https://github.com/juttle/juttle/pull/126/files#diff-c4f45babea35a0b332f90b9bbf245e82R181 the name of that test is wrong and I shall fix

rlgomes:
@dmehra still need to create the documentation section for this `grok` stuff and add a few good examples of using grok to quickly parse things like apache, cassandra, etc. logs

dmehra:
Ignore my question about supporting this in other adapters, didn't realize all of them (file, stdio) were built ins. We also just talked about adding grok support to http adapter so we can use grok rules on files downloaded from s3. This seems useful.

rlgomes:
@dmehra the only "issue" with `read http` is that till now rely on the 'content-type' header to dictate the format and not like there's an official `application/grok` content-type nor would we want to impose that. So I'm thinking we'd expose the `-format` and `-formatOptions` to the `read http` to override the default `content-type` detection.

rlgomes:
Bump! @demmer @davidvgalbraith @go-oleg opinions ?

package.json
```diff
@@ -39,8 +39,10 @@
     "moment-duration-format": "^1.3.0",
     "moment-parser": "^0.3.0",
     "moment-timezone": "^0.4.0",
+    "node-grok": "github:beh01der/node-grok#b76dae6dc4efa745e6bf75a3cb8d113aa47bc16e",
```
demmer: Is there no valid node-grok implementation published to npm?

lib/adapters/file/read.js
```diff
@@ -9,7 +9,7 @@ var Read = Juttle.proc.source.extend({
     procName: 'read-file',
 
     initialize: function(options, params, location, program) {
-        var allowed_options = ['from', 'to', 'file', 'timeField', 'format'];
+        var allowed_options = ['from', 'to', 'file', 'timeField', 'format', 'formatOptions'];
```
demmer: Using `-formatOptions` seems opaque. I'd lean more towards adding a `-pattern` option that's only valid when reading with format grok.

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,76 @@
...
+        this.pattern = options.pattern || '%{GREEDYDATA:message}';
+    },
+
+    parseStream: function(stream, emit) {
+        var self = this;
+
+        return new Promise(function(resolve, reject) {
+            grok.loadDefault(function(patterns) {
+                var buffer = [];
+                var liner = readlines.readlineStream();
```
demmer: Seems to me like we should use the builtin node.js implementation of reading a file line by line instead of pulling in this half-baked module:
https://nodejs.org/api/readline.html#readline_example_read_file_stream_line_by_line

package.json
rlgomes: They haven't published the 1.0.5 version yet and it has the recent and working copy of oniguruma which actually builds fine locally for me and on travis (unlike the one in 1.0.4 which is broken, ie oniguruma@4.2.1 ), here's the specific commit: 

https://github.com/Beh01der/node-grok/commit/bf961bc3b8301f491314a8af62fa69f3927bec48

I could poke at the owner to publish the latest 1.0.5 build.

lib/adapters/file/read.js
rlgomes: Agreed as that would make things look simpler after all but its just odd that `-pattern` then applies to the format code and not the `read xyz` code.

lib/adapters/parsers/grok.js
rlgomes: Sure, for some reason came across readlines before the built in readline.

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,76 @@
...
+                var streamEnd = false;
+
+                liner.on('readable', function(){
+                    var line;
+                    while ((line = liner.read())) {
+                        line = line.toString();
+                        pending++;
+                        pattern.parse(line, function (err, point) {
+                            if (err) {
+                                reject(errors.runtimeError('RT-INTERNAL-ERROR',{
```
demmer: This doesn't stop the reader or guard against future errors, so this may cause the reject to happen mulitple times which seems like a bad thing.

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,76 @@
...
+                            } else {
+                                buffer.push(point);
+
+                                if (buffer.length >= self.limit) {
+                                    emit(buffer);
+                                    buffer = [];
+                                }
+                            }
+
+                            if (streamEnd) {
```
demmer: This seems incorrect -- what if pending is > 0 at this point but you've gotten an eof?

Overall it seems like you should DRY up the code by adding a `checkDone(pending, streamEnd)` function that you could call both here and in the 'end' event handler. In fact that could also do the check to emit when the buffer exceeds the limit as well which should tighten things up as well.

test/adapters/stdio/read.stdio.spec.js
```diff
@@ -5,7 +5,7 @@ var juttle_test_utils = require('../../runtime/specs/juttle-test-utils');
 var check_juttle = juttle_test_utils.check_juttle;
 var path = require('path');
 
-var validFormats = {
+var symmetricalFormats = {
```
demmer: This naming change seems superfluous but more importantly it seems you should be consistent and change the file test if you're gonna change this one.

demmer:
Overall I like the idea but there's some work to do on cleaning up the implementation as mentioned above.

Also the docs need to be updated as part of this change as well.

Finally, I would once again prefer to see this broken up into individual commits that separate out the changes to add the grok parser to the core vs updating the adapters themselves to use it.

lib/adapters/parsers/grok.js
rlgomes: @demmer so we have two options here: 

1. reject and then do a stream.close() to stop everything 
2. trigger a warning like so: https://github.com/juttle/juttle/blob/master/lib/adapters/serializers/csv.js#L25 with the details of the line, skip this line and keep chugging along 
3. trigger the warning and actually place the line as a point { time: current_time, message: line } into the pipeline ? 

What do you think ?

demmer: I think skip the bad point and log a warning makes the most sense. 

-m

rlgomes: So now that I'm fixing this I've figured out that the asynchronous parsing is leading to some really buggy code since with grok.parse being asynchronous we actually have no gaurantee on the order of the calling back... thinking about how to correctly fix this.

rlgomes:
@demmer reworked things to separate the work of this PR into the desired commits. I had to use `sync` library to remove the asynchronous behavior of the node-grok parsing which was just really complicated to get it to parse chain the asynchronous parsing callbacks so that the order of the lines were preserved I think the current solution sin't ugly and the library used is well used (https://www.npmjs.com/package/sync).

See here for the exact implementation: https://github.com/juttle/juttle/commit/1ae62e44d0261c90f5290e8b85b57fefa3ff1394#diff-a5dfa897a7bab1ae503ab3f8c82c9862R44

Things to finish up:
 - [ ] waiting on the `node-grok` maintainer on pushing out the latest `node-grok` build
 - [ ] documentation for the new grok format and make sure to update each of the adapters it is used by including detailed examples of parsing popular log formats.

@go-oleg @davidvgalbraith your input is always welcome

rlgomes:
Failure on last was due to code coverage off by 1 line between two different node versions ? That is highly suspicious:

4.2.4:
```
Statements   : 92.38% ( 6429/6959 )
Branches     : 86.94% ( 3048/3506 )
Functions    : 91.97% ( 1351/1469 )
Lines        : 92.4% ( 6386/6911 )
```

vs 

5.0
```
Statements   : 92.4% ( 6430/6959 )
Branches     : 86.94% ( 3048/3506 )
Functions    : 92.04% ( 1352/1469 )
Lines        : 92.4% ( 6386/6911 )
```

I'll see where I'm missing test coverage and fix that instead of tweaking our numbers.

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,71 @@
...
+
+    parseStream: function(stream, emit) {
+        var self = this;
+
+        return new Promise(function(resolve, reject) {
+            var buffer = [];
+            var pending = 0;
+            var streamEnd = false;
+
+            function checkDone(streamEnd) {
```
davidvgalbraith: You don't need to pass `streamEnd` as a parameter, it's in scope already.

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,71 @@
...
+            grok.loadDefault(function(patterns) {
+                var pattern = patterns.createPattern(self.pattern);
+                var liner = readline.createInterface({
+                    input: stream,
+                    terminal: false
+                });
+
+                liner.on('line', function(line) {
+                    pending++;
+                    sync(function() {
```
davidvgalbraith: `sync` makes this run synchronously, right, so do you need all that `pending` tracking/logic?

davidvgalbraith: Actually I don't understand why you need `sync` at all, looks like it would run synchronously either way.

lib/adapters/parsers/grok.js
rlgomes: True.

lib/adapters/parsers/grok.js
rlgomes: Yes with `sync` we don't need the pending logic but without it nothing here is synchronous as the node-grok library calls back with the parsed result asynchronously (due node-grok itself using the oniguruma parser which is asynchronous as well, I even went digging into the code myself to see if there were other sync methods, but no such luck: https://github.com/Beh01der/node-grok/blob/master/lib/index.js#L15)  and that was the reason I had to use the pending logic which was super buggy. I'll fix up things by removing the silly pending logic which is no longer in use.

davidvgalbraith: So, `pattern.parse.sync` is asynchronous, you're saying?

rlgomes: `pattern.parse.sync` isn't from the `node-grok` package (which had the asynchronous function `pattern.parse`), its added to the prototype of all functions through the [sync](https://www.npmjs.com/package/sync) library which then allows you to synchronize an asynchronous function but only when executed within the closure of `sync(function() { ... })`. I may have miss read their examples and how to use the lib so give their README a look and correct me if I'm wrong.

The only reason for this is that if I keep things using the asynchronous method its just a mess to figure out how to reorder things. I even tried promisifying the whole thing and chaining promises (to guarantee the order of that the lines are read from are maintained when emitting points) but things just get even harder to follow at that point.

rlgomes: I'll look at how to solve this in a nicer non crappy sync way.

rlgomes:
@davidvgalbraith please give my promise voodoo a look. At least now I don't try to pull in the sync library and just did things using promises and chaining them the right way.

davidvgalbraith:
Ok @rlgomes I was going to make a suggestion but I wasn't sure it was going to work, so I ended up hacking it out & just pushed up the result. If that looks OK to you I am +1 on this.

rlgomes:
@davidvgalbraith thanks! and looks great. I was thinking about chaining the last flush like that by chaining it at the end of the current promise when I was away from my keyboard and you've proven it works well and simplifies the code greatly. 

I'll be adding the missing documentation bits and then will check with @demmer on having the unofficial build of `node-grok` being used vs waiting for the maintainer to publish the latest changes: https://github.com/Beh01der/node-grok/issues/10 which may not happen soon enough.

lib/adapters/parsers/base.js
```diff
@@ -1,7 +1,17 @@
...
+     * A parser must implement the parseStream function below and should fire
+     * errors by using the
+     * `this.emit('error', ...)` EvenEmitter method.
+     */
     initialize: function(options) {
         this.limit = (options && options.limit) ? options.limit : 1024;
+    },
+
+    parseStream: function(stream, emit) {
+        // implement me
```
demmer: Instead of just a comment perhaps this should throw an `Error`?

lib/adapters/parsers/grok.js
```diff
@@ -0,0 +1,66 @@
...
+                            self.emit('error', errors.runtimeError('RT-INTERNAL-ERROR',{
+                                error: 'Skipping line "' + line + '", due to: ' + err.toString()
+                            }));
+                        });
+                    });
+                });
+
+                liner.on('close', function() {
+                    currentPromise.then(function() {
+                        emit(buffer);
```
demmer: Seems like this should have guard of `if (buffer.length !== 0)`

lib/adapters/parsers/base.js
rlgomes: @demmer `throw Error('implement me')` ?

lib/adapters/stdio/read.js
```diff
@@ -8,7 +8,7 @@ var Read = Juttle.proc.source.extend({
     procName: 'read stdio',
 
     initialize: function(options, params, location, program) {
-        var allowed_options = ['timeField', 'format', 'formatOptions'];
+        var allowed_options = ['timeField', 'format', 'pattern'];
```
demmer: We should probably throw an error if the user specifies a `-pattern` option when the format isn't `grok`.

lib/adapters/parsers/grok.js
rlgomes: We wanted to push out the empty buffer at the end of parsing as we do with all buffers and should keep that behavior consistent right ?

test/adapters/file/file.spec.js
```diff
@@ -10,7 +10,7 @@ var expect_to_fail = juttle_test_utils.expect_to_fail;
 
 var tmp_file = tmp.tmpNameSync();
 
-var validFormats = {
+var symmetricalFormats = {
```
demmer: Why this naming change?

lib/adapters/stdio/read.js
rlgomes: Sure

test/adapters/file/file.spec.js
rlgomes: because grok isn't "symmetrical" and yet doesn't mean its not part of the valid formats ? I can move this back to the original name.

demmer: Oh right -- now I remember commenting on this in the other file.

In context the change makes some sense, but then I would think you'd want a test for grok format in the file adapter here to make it consistent and to actually test that grok is usable in files.

lib/adapters/parsers/base.js
demmer: Sure -- maybe `throw new Error('parseStream must be implemented')` to be more boring and serious

lib/adapters/parsers/grok.js
demmer: I didn't look carefully at the others, but if this is the expected contract then I'm ok with it.

rlgomes:
@dmehra documentation additions:

https://github.com/juttle/juttle/commit/0789f008095ee468dfb52a71eaba2bdab417beaa

may need better names and some restructuring but I think its time we document the "parsers" or "input formats" in a separate doc that can be linked form all of the adapters and each adapter can document the supported list.

@demmer made the fixes you asked for in:

https://github.com/juttle/juttle/commit/8eba2340351edd80f5aa440014539c82451be244

rlgomes:
I'm opening a ticket to expose additional pattern files to the grok parser but don't want to delay this PR any further for that: https://github.com/juttle/juttle/issues/205.

rlgomes:
@dmehra another quick look: 

https://github.com/juttle/juttle/commit/19c2eae4b52783de3ba6301439b674b5a913d068

rlgomes:
@demmer @davidvgalbraith will wait for both your +1's on the final changes here.

davidvgalbraith:
+1 on the final changes here

rlgomes:
@demmer I never got a response from the node-grok maintainer so we'd go with the current non official build directly from the github repo.

demmer:
@rlgomes OK. Let's file a task to remember to update the package.json once a new version gets published.

rlgomes:
@demmer opened https://github.com/juttle/juttle/issues/213 and I linked it to the issue on `node-grok` side so I'll get notified once there's something done on that issue. Meanwhile good to merge this ?

demmer:
Yep!

On Tue, Jan 19, 2016 at 12:08 PM, Rodney Lopes Gomes <
notifications@github.com> wrote:

> @demmer <https://github.com/demmer> opened #213
> <https://github.com/juttle/juttle/issues/213> and I linked it to the
> issue on node-grok side so I'll get notified once there's something done
> on that issue. Meanwhile good to merge this ?
>
> —
> <https://github.com/juttle/juttle/pull/126#issuecomment-172971361>.
>



-- 
Michael Demmer
Jut - CTO / VP Engineering
demmer@jut.io

PR #128, opened 2016-01-07
Add tests for non-scalar fields group by for put, head and tail

bkutil:
refs: https://github.com/juttle/juttle/issues/55


bkutil:
cc: @dmehra @dmajda

dmehra:
+1 assuming that objects with lists are covered elsewhere so it's sufficient to use a simple `"o": { "a": 1 }` here. What I mean is that other test generically covers handling of objects like `"o": { "a": 1, "b": 2 }` so we don't need to use them in each proc testcase.

bkutil:
@dmehra - added tests that exercise grouping by object, array, object containing arrays and arrays containing objects...

bkutil:
Merging based on +1 from @dmehra

PR #129, opened 2016-01-07
Optimize tail

davidvgalbraith:
@demmer @dmajda @bkutil


dmajda:
+1

bkutil:
Maybe we could add an explicit test showing that `tail x by y` optimisation is not supported?

PR #133, opened 2016-01-08
delay loading adapters

### LINKED ISSUE #83: shouldn't try to load adapters that aren't needed
Juttle crashes when I run `juttle -e "emit -limit 2 | put message='hello'"` because it can't find an adapter that's listed in my config.json (but not currently installed).

It would be nicer if it only tried to lookup adapter(s) that are actually needed in the script that is being run.

demmer:
Instead of eagerly loading all of the adapters when the runtime is
initialized, stash the configuration and defer loading the module
until the adapter is actually used.

This avoids irrelevant errors in loading adapters that aren't needed for the
given program and speeds up initialization in cases where not all of the
configured adapters are used.

This required a change to how the adapters are configured -- instead of
specifying the module to load as the key in the config file, it now expects the
name of the adapter, i.e. `elastic` instead of `juttle-elastic-adapter`.

By default, for an adapter named `xyz`, the system tries to load
`juttle-xyz-adapter`, though the path can still be overridden in the adapter
config.

The example in the docs actually erroneously used the new syntax in the
example, so there was nothing to update there, though I also added a bit
more description of what actually happens when the adapters are configured.

Rename the API function from `adapters.load` to `adapters.configure`
since the latter is a better expression of what the operation actually does.

Change how the builtin adapters are registered to be done as a side-effect of
calling `adapters.configure()` as opposed to always loading them whenever the
runtime is required. This is both more consistent and means that the CLI adapters
won't be included in a browserify build of the runtime.

Fixes #83.


demmer:
@dmajda @bkutil please take a look

test/runtime/test-adapter-clone.js
```diff
@@ -0,0 +1,10 @@
+var TestAdapter = require('./test-adapter');
```
bkutil: The test adapter clone is related only to that single adapter test, maybe we should keep it inside of `test/runtime/adapter.spec.js`...

lib/runtime/adapters.js
```diff
@@ -17,40 +18,69 @@ function register(type, module) {
...
-            logger.error('error loading adapter ' + module + ': ' + err.message);
-            throw err;
+// Add configuration for the specified adapters but don't initialize them until
+// they are actually used.
+function load(config) {
+    logger.debug('configuring adapters', _.keys(config).join(','));
+    _.each(config, function(options, type) {
+        // To be backwards-compatible with the old configuration format, if the
+        // adapter name is of the form `juttle-xyz-adapter`, register it as type
+        // `xyz`. This should be deprecated at some point in the future.
```
bkutil: IMHO we should just the switch now, while we don't have many external users and make the corresponding change in all our adapters and their docs now, otherwise, this comment will never go away :)

test/runtime/test-adapter.js
```diff
@@ -8,8 +8,9 @@
 
 var _ = require('underscore');
 var store = {};
+var Juttle = require('../../lib/runtime/index').Juttle;
 
-function TestAdapter(config, Juttle) {
+function TestAdapter(config) {
```
bkutil: Good catch! Thanks for fixing this :)

bkutil:
Overall, looks good. What about fixing the configs now, so we don't have to deal with it later, then merge those changes and this PR together?

test/runtime/test-adapter-clone.js
dmajda: I’m personally find with both.

dmajda:
+1

I agree with @bkutil that it would be better to change adapters now too, or at least clearly limit the compatibility code life span (by declaring when this will go away and filing an issue for that so it’s not forgotter).

test/runtime/test-adapter-clone.js
demmer: It actually needs to be in a separate file since it is lazily loaded using a deferred require.

demmer:
@dmajda / @bkutil I removed the support for legacy config like you suggested and also renamed `adapters.load` to `adapters.configure` since that seems more appropriate.

This will require an outrigger change to match the updated API.

I also updated the docs to include a bit more detail on how the adapters work, added a semantic pass check to ensure that any referenced adapters are actually configured, and reworked how the builtin adapters are loaded so it is done as part of the call to `adapters.configure` and not implicitly when the runtime is required.

FYI @go-oleg this last bit is what we discussed earlier so it shouldn't block your browserify problem.

go-oleg:
thx!

demmer:
@dmajda I added minor commits to address your various PR comments. I will squash these before merging.

I also removed the semantic pass check for valid adapters from this PR. I went back and forth about whether or not it belonged in the semantic pass or in the graph compilation -- it's true that it seems odd to pull files from runtime into the compilation phase but at the same time adapters aren't really _part_ of the runtime and it seems like a good check to have that the program is semantically valid in the current compilation context (which to me seems ok if it includes all configured adapters).

In any event I think it makes sense to separate that out from this change since they're actually unrelated. We can continue the discussion on another PR.

dmajda:
+1

PR #134, opened 2016-01-08
adapters: avoid loading multiple copies of juttle

### LINKED ISSUE #130: adapters should not load a second copy of juttle
Currently, adapters are loaded using a vanilla node `require()` which can result in a problem if the version of juttle being used is not exactly the same code that ends up being loaded via a call to `require()` in the context of the adapter.

For example, consider the case in which a developer has a copy of juttle and the elastic adapter installed in `~/node_modules` but a second copy checked out in a development directory.

In this case, running `bin/juttle -e 'read elastic...`' will result in the following error:

```
Error: elastic_read cannot be used at the start of a flowgraph. (RT-PROC-CANNOT-START-FLOWGRAPH)
```

The reason for this is that the development copy of juttle was able to resolve `require('juttle-elastic-adapter')` to the code in ~/node_modules, but when the adapter itself calls `require('juttle/lib/runtime')` it ends up loading a second copy of juttle from ~/node_modules.

This means that the instantiated javascript runtime ends up with two copies of `Juttle.proc.source` and the one used by the compiler core is not the same one that was used as the base class for the adapter.

To fix this, we should override the node implementation of `Module._load` using the same technique as https://github.com/mfncooper/mockery and rewrite any require calls so that they reference the juttle runtime that's actually in use and not load a second copy.

demmer:
Because adapters pull in the Juttle runtime via require('juttle/lib/xyz'),
there's a chance that loading an adapter may load another copy of the runtime
code, depending on the specifics of how the user installed the various node
modules.

This causes problems if identical code is loaded from two different paths,
since `instanceof` will fail to match an object created from one copy to the
class definition in another copy.

To avoid these issues, before loading the adapter, intercept the node.js
module loader and if there is a conflict for juttle, log an informative
error and exit the process.

If the environment variable JUTTLE_RESOLVE_ADAPTER_CONFLICT exists, then
log a warning and rewrite any require requests for `juttle/*` to be relative
paths using the _current_ module (i.e.  lib/runtime/adapters.js) as the
context.

This ensures that we don't have strange errors when multiple copies of the
runtime are loaded into the same process.

Fixes #130

### DIFF
--- a/lib/runtime/adapters.js
+++ b/lib/runtime/adapters.js
@@ -5,6 +5,7 @@
 var _ = require('underscore');
 var path = require('path');
 var errors = require('../errors');
+var Module = require('module');
 var logger = require('../logger').getLogger('juttle-adapter');
 
 var BUILTIN_ADAPTERS = ['file', 'http', 'stdio', 'stochastic'];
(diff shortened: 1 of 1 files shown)


demmer:
@dmajda @bkutil

demmer:
In response to @dmajda's comments on #130, I just rewrote the implementation so that it now (a) only rewrites the module path if it detects there is an actual conflict and (b) logs a warning the first time it does a rewrite so that the whole thing is more transparent to the user.

demmer:
@dmajda I must have accidentally clicked the close PR button earlier.

This is actually ready for review. See the updated PR description (which mirrors the commit message) for the new behavior.

dmajda:
+1

PR #137, opened 2016-01-08
Replace checks for isSink with inheritance test

bkutil:
All adapters and sinks descend from the Juttle.procs.sink class,
so tests whether a certain proc can rely on that, instead of explicitly
testing an `isSink` boolean flag.

Also fix circular dependency problem between base.js using Juttle.procs
and procs.js requiring base by assigning Juttle to module.exports early.


bkutil:
cc: @dmajda

dmajda:
+1 with the comment addressed

bkutil:
Cool, thanks. Will address the comment and merge :)

PR #144, opened 2016-01-11
Lexicographic ordering on arrays

bkutil:
Define lexicographic ordering on arrays and objects.

Assume that the LHS and RHS are of the same 'shape'. Comparing
different data types at any level of nesting results in an error.

For arrays, define the lexicographic ordering naturally, shorter
arrays are smaller than longer ones, and the elements are compared by a
recursive call in the order of their indices.

For objects, sort the keys, then lexicographically compare them using
the algorithm for arrays. If the lists of keys are equal, compare the
corresponding values recursively in the sorted key order. Sorting of
keys ensures that comparing objects is independent on key ordering
(particularly on insertion-order) and is consistent with implementation
of equality in `values.js`.

For primitive data types (integers, strings, JuttleMoments, etc...),
fall back to the existing implementation.


bkutil:
cc: @dmajda

bkutil:
WIP: whether to support sorting for objects at all and whether to use lexicographic (view keys as another dimension) or inclusion (view keys as a set) sorting is still open - will discuss with David tomorrow.

bkutil:
After a discussion with @dmajda, we decided to drop the support for sorting on objects for now. The reason being that it is not clear which sorting order would be natural. The implemented lexicographic sort did not give 'intuitive' results. David might want to add more details.

lib/runtime/ops.js
```diff
@@ -219,6 +219,20 @@ var ops = {
             return JuttleMoment.lt(left, right);
         } else if (values.isDuration(left) && values.isDuration(right)) {
             return JuttleMoment.lt(left, right);
+        } else if (values.isArray(left) && values.isArray(right)) {
+            if (left.length !== right.length) {
+                return (left.length < right.length);
```
dmajda: Nit: No parens.

lib/runtime/ops.js
```diff
@@ -247,6 +275,8 @@ var ops = {
             return JuttleMoment.lte(left, right);
         } else if (values.isDuration(left) && values.isDuration(right)) {
             return JuttleMoment.lte(left, right);
+        } else if (values.isArray(left) && values.isArray(right)) {
+            return values.equal(left, right) || this.lt(left, right);
```
dmajda: I don’t like the asymmetry between the definitions of `lt`/`gt` and `lte`/`gte`. At the same time I don’t like the big chunk of code in `lt`/`gt` (it feels like something which should be in a separate function) and the duplication of code in all the implementation of comparison operators.

What about implementing the comparison as a separate function in `values.js`, named `values.compare`? It would return the usual `strcmp`-like result and the operators would than be implemented like this:

```javascript
lt: function(left, right) {
    return values.compare(left, right) < 0;
}
```

This solution has some performance overhead, but I don’t think it will matter much in practice and we can always inline `values.compare` and revert to duplicate code if needed.

test/runtime/ops.spec.js
```diff
@@ -0,0 +1,81 @@
...
+                { left: [[1]], right: [[1]], result: false },
+                { left: [[2]], right: [[1]], result: false },
+            ];
+
+            _.each(tests, function(test) {
+                expect(ops.lt(test.left, test.right)).to.equal(test.result, JSON.stringify(test.left) + ' lt ' + JSON.stringify(test.right));
+            });
+        });
+    });
+});
```
dmajda: The `ops` module is generally tested on the language level by JuttleSpec tests such as `test/runtime/specs/juttle-spec/expressions/less-than-operator.spec.md`. I think we should continue that.

test/runtime/specs/juttle-spec/juttle-procs-sort.spec.md
```diff
@@ -126,3 +126,37 @@ Timestamps: batched flow, with grouping
...
+        { arr: [1], value: 1},
+        { arr: [3, 2, 3], value: 3},
+        { arr: [2, 2], value: 2},
+    ] | sort arr | view result
+
+### Output
+
+    {"arr": [1], "value": 1}
+    {"arr": [2,2], "value": 2}
+    {"arr": [3,2,3], "value": 3}
```
dmajda: Assuming the sorting itself is tested I think just one test is enough here.

test/runtime/ops.spec.js
```diff
@@ -0,0 +1,81 @@
...
+                { left: [1, 1], right: [1, 2], result: false },
+                { left: [2, 1], right: [1, 1], result: true },
+                { left: [1, 1], right: [2, 1], result: false },
+                { left: [[2]], right: [[2]], result: false },
+                { left: [[1]], right: [[1]], result: false },
+                { left: [[2]], right: [[1]], result: true },
+            ];
+
+            _.each(tests, function(test) {
+                expect(ops.gt(test.left, test.right)).to.equal(test.result, JSON.stringify(test.left) + ' gt ' + JSON.stringify(test.right));
```
dmajda: Split to multiple lines please.

test/runtime/ops.spec.js
```diff
@@ -0,0 +1,81 @@
...
+                { left: [1, 1], right: [1, 2], result: true },
+                { left: [2, 1], right: [1, 1], result: false },
+                { left: [1, 1], right: [2, 1], result: true },
+                { left: [[1]], right: [[2]], result: true },
+                { left: [[1]], right: [[1]], result: false },
+                { left: [[2]], right: [[1]], result: false },
+            ];
+
+            _.each(tests, function(test) {
+                expect(ops.lt(test.left, test.right)).to.equal(test.result, JSON.stringify(test.left) + ' lt ' + JSON.stringify(test.right));
```
dmajda: Split to multiple lines please.

lib/runtime/ops.js
bkutil: I just noticed that the errors coming from the `lt`, `gt`, and friends explicitly mention the operator, so either we have to pass it down to compare (meh), or change the error (weird, as it would be the same for all operators), or catch and re-throw (slow)... so the implementation won't be the nice oneliner here - any other ideas?

test/runtime/specs/juttle-spec/juttle-procs-sort.spec.md
bkutil: I decided to keep the test in there because it documents how array length affects the order.

dmajda:
+1 with the suggested changes and appropriate squashing

(BTW I may be a bit more nitpicky than usual here because this part of code defines core Juttle semantics so I want the code to be really clear, DRY, etc.)

bkutil:
No worries :) I addressed your comments and squashed the commits. Would you mind doing another pass, please?

dmajda:
+1 with the nit fixed

bkutil:
Nit fixed, will merge on green.

bkutil:
Refs: #55

PR #147, opened 2016-01-12
http adapter -rootPath option

### LINKED ISSUE #146: feature: read http option to access a nested field
To support HTTP API endpoints that return an array of objects inside a top-level wrapper object, the http adapter should support an optional `-root` argument to read.

For example, the npm download counts API at https://api.npmjs.org/downloads/range/last-month/juttle returns an object like:
```
{
    "downloads": [
        {
            "day": "2015-12-12",
            "downloads": 3
        },
        {
            "day": "2015-12-13",
            "downloads": 1
        },
        ...
    ],
    "start": "2015-12-12",
    "end": "2016-01-10",
    "package": "juttle"
}
```

Rather than emitting a single point, it would be better to get download counts for each day as separate juttle points.

To address this, the `root` option could contain a `.` separated string that indicates the path to a field in the result object that would be treated as the root element. Note that this syntax is a subset of JSON path (e.g. https://github.com/jayway/JsonPath). 

When specified, all other options (e.g. `timeField) would then apply to the contents of the object.

demmer:
Add a `-rootPath` option to `read http` that can specify a path into the retrieved object from which the points should be processed.

Fixes #146.

### DIFF
--- a/docs/examples/adapters/http_npm_download_counts.juttle
+++ b/docs/examples/adapters/http_npm_download_counts.juttle
@@ -1,0 +1,13 @@
+/*
+ * Query daily download counts from npm
+ *
+ * Input parameters:
+ *
+ *  package: package name (i.e. 'juttle')
+ */
+
+input package: text -default 'juttle';
+
+read http -url 'https://api.npmjs.org/downloads/range/last-month/${package}'
+     -rootPath 'downloads' -timeField 'day'
+| view table
--- a/docs/adapters/http.md
+++ b/docs/adapters/http.md
@@ -18,6 +18,7 @@ read http -url url
           -headers headers
           -body body
           -timeField timeField
+          -rootPath rootPath
           -includeHeaders true/false
 ```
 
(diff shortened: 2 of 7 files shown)


demmer:
@mstemm

docs/examples/adapters/http_npm_download_counts.juttle
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * Input parameters:
+ *
+ *  package: package name (i.e. 'juttle')
+ */
+
+input package: text -default 'juttle';
+
+read http -url 'https://api.npmjs.org/downloads/range/last-month/${package}'
+     -rootPath 'downloads' -timeField 'day'
```
mstemm: When we were looking at the output earlier, you noticed a bug where the day was sticking around as a column. Did that get fixed?

mstemm:
Other than the above question, which won't really affect this code, lgtm.

bkutil:
Looks good. +1

docs/examples/adapters/http_npm_download_counts.juttle
demmer: Nope -- thanks for reminding me. Just filed #159

PR #150, opened 2016-01-12
values: Improve `values.inspect` comment

dmajda:
Explicitly state that the representation of each Juttle value is unique.


dmajda:
@bkutil

bkutil:
+1

Issue #131, opened 2016-01-08
Do write procs have to call both emit_eof() and done()?


mstemm:
From looking at the juttle-elastic-adapter, which I'm using as a basis for the gmail adapter. When the elastic adapter has completed all writes, it calls both emit_eof() and done().

However, it doesn't look like the sql adapter calls emit_eof(). It only calls done().

Which are required? Both emit_eof() and done() or just done()?

dmajda:
Calling `emit_eof` is a noop because sink has no output node. I went through all our adapters and eliminated such calls from their writers.

Calling `done` is needed so that the program knows when sinks finished processing.

PR #151, opened 2016-01-12
HTTP adapter: Eliminate useless `emit_eof` calls in `Write`

dmajda:
These calls are noops (because `Write` is a sink).

See #131.


dmajda:
@bkutil

bkutil:
+1

PR #156, opened 2016-01-12
adapters: check for valid adapters in the build stage

demmer:
Add a check to the flowgraph build stage that makes sure an adapter with the
specified name is either registered or configured to be dynamically loaded.

As part of this change, modify the ast for read and write in order to preserve
the location information for the adapter name.

Relates to #83

### DIFF
--- a/lib/compiler/ast-visitor.js
+++ b/lib/compiler/ast-visitor.js
@@ -57,12 +57,12 @@ var NODE_CHILDREN = {
     BuiltinProc:              ['options', 'arg'],
     OptionOnlyProc:           ['options'],
     PutProc:                  ['options', 'exprs', 'groupby'],
-    ReadProc:                 ['options', 'filter'],
+    ReadProc:                 ['options', 'filter', 'adapter'],
     ReduceProc:               ['options', 'exprs', 'groupby'],
     SingleArgProc:            ['options', 'arg', 'groupby'],
     SequenceProc:             ['filters'],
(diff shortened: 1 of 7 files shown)


demmer:
@dmajda

demmer:
@dmajda This is the commit that I extracted from #133 and reworked to apply to the graph building stage and not the semantic pass. It still _feels_ a bit more like a semantic time check to me than a graph building step but I can live with it either way.

dmajda:
The change looks OK technically.

My objection wasn’t against the check being in the semantic pass, but against it being in the compiler as a whole (as opposed to the runtime, i.e. generated program). While this is technically fine, it unnecessary conflates the runtime and compile-time and spreads knowledge about adapters into parts of the code which don’t need to know about them. One thing this will likely hurt in the future is “generification” of AST node processing.

On my scale of architectural “sins” this is far from the top, so I’ll stop advocating now and leave the decision to you. If compile-time check stays, semantic pass seems like a better place, but I can live with a build-time check too.

demmer:
@dmajda Since the only user benefit to the change was to include location information in the error, I reworked it so that the locations are propagated into the generated code so they're available in the runtime.

To my taste, it seems like the original solution that did this all in the semantic pass felt like the right architectural choice but I am ok with it this way.

dmajda:
+1

(I imagined a tiny helper above `Juttle.adapters.get` so that this function doesn’t have to take the location, but whatever.)

Issue #152, opened 2016-01-12
JSON Lines serializer tests are unreliable


dmajda:
I often get this test failure locally when running `gulp test` with Node.js 4.2.4:

```
  1) serializers/jsonl can write no points to a provided stream:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

Deeper in the log, there is this error:

```
  serializers/jsonl
    ✓ can instantiate a jsonl serializer
[16:22:51] 'test' errored after 44 s
[16:22:51] Error in plugin 'gulp-mocha'
Message:
    ENOENT: no such file or directory, open '/var/folders/zp/7fg2ysbj4g9drhgxdd32dk2c0000gn/T/tmp-31317JQgtQ5Zfaqxi'
Details:
    errno: -2
    code: ENOENT
    syscall: open
    path: /var/folders/zp/7fg2ysbj4g9drhgxdd32dk2c0000gn/T/tmp-31317JQgtQ5Zfaqxi
    domain: [object Object]
    domainThrown: true
Stack:
Error: ENOENT: no such file or directory, open '/var/folders/zp/7fg2ysbj4g9drhgxdd32dk2c0000gn/T/tmp-31317JQgtQ5Zfaqxi'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.readFileSync (fs.js:397:15)
    at /Users/dmajda/Jut/Code/juttle/test/adapters/serializers/jsonl.spec.js:25:23
    at doNTCallback0 (node.js:419:9)
    at process._tickDomainCallback (node.js:389:13)
    1) can write no points to a provided stream
    ✓ can write out a few points correctly
```

dmajda:
This also happens on Travis CI, see e.g. [this job](https://travis-ci.org/juttle/juttle/jobs/101880046).

@rlgomes Could you have a look?

rlgomes:
I'll have a look at this today

dmajda:
Fixed by #157.

PR #157, opened 2016-01-12
fix timing bug in serializer tests

rlgomes:
Fixes issue #152

The issue was not waiting for the `open` event on the new writable
stream before attempting to verify the file was there and contained
anything that was recently written to it.

### DIFF
--- a/test/adapters/serializers/csv.spec.js
+++ b/test/adapters/serializers/csv.spec.js
@@ -16,14 +16,17 @@ describe('serializers/csv', function() {
     it('can write no points to a provided stream', function(done) {
         var tmpFilename = tmp.tmpNameSync();
         var stream = fs.createWriteStream(tmpFilename);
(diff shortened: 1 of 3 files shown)


rlgomes:
@dmajda this should fix the problem from now on, wasn't easy since the `open` event business is not documented anywhere in the node fs module documentation. I figured things out when I read a stackoverflow article over here: http://stackoverflow.com/a/12906805/1364656 

I was hitting this when running in a loop like so:

```bash
while [ $? == 0 ]; do mocha test/adapters/serializers; done
```

within a minute and now ran for over 5 minutes without any issues so it was just simply the issue of not waiting for the stream to `open` (ie exist) before proceeding.

demmer:
Looks good to me.

dmajda:
@rlgomes Thanks!

Issue #140, opened 2016-01-11
Clean up parser module internals


dmajda:
* [x] Rewrite `parse` and `parseSync` internals to be more readable and consistent
* [x] De-duplicate `parse` and `parseSync`
* [x] Formatting cleanup

dmajda:
Resolved by #141, #149, and #161.

PR #161, opened 2016-01-13
Parser cleanups: Final round

dmajda:
Part of work on #140.


dmajda:
@bkutil

bkutil:
Just the typo & the question (depending on the answer, either the code is fine as is, or the throw should be changed to use `errors`). Otherwise +1.

dmajda:
Fixed the typo, will merge on green.

PR #170, opened 2016-01-14
 source: add timeUnit option

### LINKED ISSUE #166: read -timeField option doesn't work with unix timestamps
I have a json file that has a field called timestamp with unix timestamp values that look like: 1449012170.2838

When I specify the -timeField 'timestamp' option to read, the time field in the output is "Sat Jan 17 18:30:12 1970 UTC" and does not change. The same thing happens if the values in the timestamp field are integers.

I would love it if unix timestamp fields could be used with -timeField so I wouldn't have to convert them myself.

davidvgalbraith:
if the times on points from a source are numeric, add a -timeUnit
option that tells what the numbers represent. Default to seconds.

fixes #166 

@demmer @dmajda @bkutil

### DIFF
--- a/docs/adapters/file.md
+++ b/docs/adapters/file.md
@@ -13,7 +13,7 @@ The `file` adapter allows reading points from, or writing points to, a file on t
 Supported file formats are JSON array, and JSON lines; see examples below.
 
 ```text
-read file -file <path> [-format <format>] [-timeField <fieldname>]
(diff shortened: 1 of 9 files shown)


davidvgalbraith:
Thanks @bkutil, updated.

dmajda:
See my [comment](https://github.com/juttle/juttle/issues/166#issuecomment-171679282) in #166.

PR #171, opened 2016-01-14
Add tests for nested points in join proc

bkutil:
Join proc works without modifications for nested points (as it relies on groups, which have been already updated). Just add tests (tm).

### DIFF
--- a/test/runtime/specs/juttle-spec/juttle-procs-join.spec.md
+++ b/test/runtime/specs/juttle-spec/juttle-procs-join.spec.md
@@ -173,6 +173,16 @@ joins 3 1-point streams
 ### Output
     { "time":"2010-01-01T00:00:00.000Z", "foo":1, "bar":2, "baz":3 }
 
+joins 3 1-point streams with nested points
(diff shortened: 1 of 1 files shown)


bkutil:
cc: @dmajda

bkutil:
refs: #55

dmajda:
Is there a specific reason why these 3 tests?

bkutil:
Yes - I noticed three major classes of tests for joins and added one of the tests for each of those:

1. a simple case (join of 3 streams into 1)
2. an outer join
3. a join by

I'll fix the commit message to state this. Do you think some other class of tests is missing?

dmajda:
> I'll fix the commit message to state this. Do you think some other class of tests is missing?

No. There are many other kinds of joins, but I don’t think they treat points differently, so the coverage you added is sufficient.

(I was thinking that maybe you modeled this after the `join` code, trying to cover different code paths.)

+1

bkutil:
Yeah, it looks like I missed `nearest`, `zip` and `once`...

PR #173, opened 2016-01-14
Update views to handle nested points

bkutil:
To output nested points, use `toString` serialization defined in `values.js`.

For csv text view, output arrays and objects in 'inspect' format, as the
brackets give a little bit more context while parsing the output
visually.

### DIFF
--- a/lib/views/text.js
+++ b/lib/views/text.js
@@ -3,6 +3,7 @@ var _ = require('underscore');
 var babyparse = require('babyparse');
 var View = require('./view.js');
 var utils = require('../runtime').utils;
+var values = require('../runtime/values');
 
 var TextView = View.extend({
     name: 'text',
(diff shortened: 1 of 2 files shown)


bkutil:
cc: @dmajda 

refs: #55

lib/views/text.js
```diff
@@ -66,12 +67,15 @@ var TextView = View.extend({
...
-    format_values: function(values) {
-        var val2 = _.map(values, function(value) {
-            if (_.isDate(value)) {
-                value = value.toISOString();
+    format_values: function(values_) {
+        var val2 = _.map(values_, function(value) {
+            if (typeof(value) === 'undefined') {
+                return '';
+            } else if (values.isArray(value) || values.isObject(value)) {
+                return values.inspect(value);
```
dmajda: I’m not sure about this bit, we should rather make `values.toString` work in such a way that this isn’t necessary.

I started a [discussion on Gitter](https://gitter.im/juttle/juttle?at=569783b7a03e28ad1adea550).

lib/views/text.js
```diff
@@ -98,10 +102,10 @@ var TextView = View.extend({
                 self.warn("Ignoring new point key(s) \"" + new_keys.join(",") + "\"");
                 self.latest_keys = self.latest_keys.concat(new_keys).sort();
             }
-            var values = _.map(self.initial_keys, function(key) {
-                return (_.has(point, key) ? point[key] : '');
+            var values_ = _.map(self.initial_keys, function(key) {
```
dmajda: Why the trailing `_`?

dmajda:
+1

The `values.toString` thing is not a blocker.

lib/views/text.js
bkutil: `values` is already defined on the top level scope.

Issue #172, opened 2016-01-14
fromNative should properly serialize nested points


bkutil:
`fromNative` currently serializes `JuttleMoment`s only at the top level of a point. It should do that recursively to handle nested points as well.

PR #176, opened 2016-01-14
FromNative handles nested time fields

bkutil:
Nested fields can contain juttle moments, which fromNative should
serialize to a string. Modify the conversion code to recursively
serialize object/array elements and add test.

Fixes: #172


bkutil:
cc: @dmajda 

refs: #55

bkutil:
bump :)

dmajda:
+1

PR #178, opened 2016-01-14
Kill lib/runtime/runtime-filter.js

dmajda:
Right now it is dead code.

If we ever want to provide the convenience of catching filter errors and dropping points to other code (e.g. backends), we should build it into whatever is returned from `FilterJSCompiler`.


dmajda:
@bkutil

bkutil:
+1

Issue #97, opened 2016-01-04
Consolidate Juttle runtime


dmajda:
Right now, a compiled Juttle program depends on `_`, `JuttleMoment`, `Filter`, `Juttle`, and `juttle` variables which contain various bits of functionality it needs. This is ugly and leads to repetitive requires with `// jshint ignore:line` scattered around the code base ([example](https://github.com/juttle/juttle/blob/master/lib/compiler/index.js#L14-L16)), which historically lead to various bugs when some of the dependencies were omitted somewhere.

Let’s consolidate this. Ideally, the executed program would be a function which would get passed a `juttle` variable (a “runtime”) containing everything it needs as its properties. In principle, the runtime can be different for various programs and environments, making embedding and testing easier.

Tasks:

  * [x] Independence on `juttle`
  * [x] Independence on `JuttleMoment`
  * [x] Independence on `Filter`
  * [x] Independence on `Juttle`
    * [x] `Juttle.FilterJSCompiler`
    * [x] `Juttle.reducers`
    * [x] `Juttle.proc`
    * [x] `Juttle.extend_options`
    * [x] `Juttle.channels`
    * [x] `Juttle.visitGen`
    * [x] `Juttle.teardown`
    * [x] `Juttle.adapters`
  * [x] Independence on `_`

PR #179, opened 2016-01-14
Consolidate Juttle runtime: Pass Juttle runtime as a dependency

dmajda:
Pass Juttle runtime as an explicit dependency to code evaluated in the `flowgraph` and `eval` stages of the compiler. This means this code does not depend on the `juttle` variable being present in the evaluation context. The change will allow making other dependencies such as `JuttleMoment` explicit too in the future.

Part of work on #97.


dmajda:
@bkutil @demmer

demmer:
Looks good to me. +1

PR #182, opened 2016-01-14
test: tweak test to be more resilient to timing

### LINKED ISSUE #181: Flaky procs test
As [spotted in Travis CI](https://travis-ci.org/juttle/juttle/jobs/102364387):

```

  1) Juttle procs tests outputs results with batch end timestamp:

      AssertionError: expected [ '00:00:00.000', '00:00:01.000' ] to deeply equal [ Array(3) ]
      + expected - actual

       [
         "00:00:00.000"
         "00:00:01.000"
      +  "00:00:02.000"
       ]
      
      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:427:19)
      at Assertion.ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/addMethod.js:40:25)
      at test/runtime/procs.spec.js:1422:58
      at bound (domain.js:280:14)
      at runBound (domain.js:293:12)
      at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
      at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
      at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
      at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
      at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/main/async.js:15:14)
```

demmer:
Previously the test could fail if :now: ended up aligned exactly on
a second boundary since the emitter would not send out the same
number of points as it did when :now: was not an exact second.

Change the test so that both start and end are quantized to the even
second so this isn't a problem.

Fixes #181

### DIFF
--- a/test/runtime/procs.spec.js
+++ b/test/runtime/procs.spec.js
@@ -1413,8 +1413,9 @@ describe('Juttle procs tests', function() {
     it('outputs results with batch end timestamp', function() {
         return check_juttle({
             program:
-            ('const start = Date.quantize(:2 seconds ago:, :second:);'+
-             'emit -from start -to :now: | batch 1 '+
+            ('const start = Date.quantize(:3 seconds ago:, :second:);'+
+             'const end = Date.quantize(:now:, :second:);'+
+             'emit -from start -to end | batch 1 '+
              '| put dt = time - start | (view before -times 1 ; pass) '+
              '| reduce count() '+
              '| put dt = time - start | view after -times 1')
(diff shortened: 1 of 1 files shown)


demmer:
@dmajda

dmajda:
I think it makes sense to also harden the test below which is prone to the same issue. Beside that, +1.

demmer:
@dmajda I fixed the other test too. Will merge on green.

PR #183, opened 2016-01-14
cli: strip locations when in flowgraph mode

demmer:
When running with --mode flowgraph, the resulting data structure
contains propagated location information from the compilation. Since
this isn't an AST we can't use the existing location stripper, so
add some simple logic to the cli to remove the location from all
graph nodes.


demmer:
@dmajda

bin/juttle
```diff
@@ -248,6 +248,13 @@ function perform_mode(options)
             stripper.strip(ast);
             return ast;
         });
+    } else if (options.action === 'flowgraph' && !show_locations) {
+        promise = promise.then(function(graph) {
+            _.each(graph.built_graph.nodes, function(node) {
+                node.location = undefined;
```
dmajda: Why not `delete`?

dmajda:
I’d prefer to move the code out of `bin/juttle`. As I argued earlier somewhere else, `bin/juttle` should mostly just wire together logic implemented & tested elsewhere.

demmer:
@dmajda Good point -- I moved the code inside LocationStripper and modified the contract slightly so the module exports a simple function as opposed to a class.

dmajda:
I’d somewhat prefer to model this as two distinct strippers, but can live with this implementation too.

+1

demmer:
@dmajda Do you mean it would be the job of the CLI to call into either an `ASTLocationStripper` or `GraphLocationStripper` based on the mode?

I went back and forth on which one to do.

dmajda:
I don’t like it much, but I like it more than putting two relatively distinct algorithms operating on different data structures into one function.

dmajda:
@demmer Can we either finish or close this? (I can take it across the finish line if you don’t have time.)

PR #186, opened 2016-01-14
JuttleMoment: new constructor option: rawDurationString

go-oleg:
This new option is used for creating a JuttleMoment duration from the native string representation of a momentjs moment.duration.

Will be used by outrigger for instantiating JuttleMoment durations from momentjs moment.duration.

@demmer


demmer:
lgtm

Issue #185, opened 2016-01-14
cli view has confusing output for points with nested objects


demmer:
The support for nested fields in CLI views that was added in #173 has confusing output due to the lack of {} when rendering objects.

For example:

```
# bin/juttle -e 'emit | put d={nest: {xyz: 123, nested: {nested2: :1m:}}} | view table'
┌────────────────────────────────────┬────────────────────────────────────────────────────────────────────┐
│ time                               │ d                                                                  │
├────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ 2016-01-14T19:15:27.241Z           │ nest: xyz: 123, nested: nested2: 00:01:00.000                      │
└────────────────────────────────────┴────────────────────────────────────────────────────────────────────┘
```

I think this would be much more clear if we simply rendered nested objects by calling the equvalent of `JSON.stringify` on the object itself. We might need some special logic to make it so that moments are properly formatted, but the current approach doesn't seem ideal.

demmer:
@bkutil please take a look

Issue #177, opened 2016-01-14
Update values.toString to output brackets


bkutil:
Based on the outcome of the discussion on gitter about `toString` output of arrays/objects, either update the code or close this issue.

bkutil:
As discussed on gitter, `values.inspect` and `values.toString` should be modified so that

```
const c1 = null;
const c2 = 5;
const c3 = "abcd";
const c4 =  :2016-01-01 12:34:56:;

const s = "Interpolated: ${c1} ${c2} ${c3} ${c4}";
```
should interpolate to:

```
"Interpolated: null 5 abcd 2016-01-01T01:01T12:34:56.000Z"
```

and:

```
const c1 = [ null, 5, "abcd",  :2016-01-01 12:34:56: ];
const c2 = { a: null, b: 5, c: "abcd", d: :2016-01-01 12:34:56: };
const c3 = [];
const c4 = {};

const s = "Interpolated: ${c1} ${c2} ${c3} ${c4}";
```

to:

```
"Interpolated: [ null, 5, "abcd", "2016-01-01T01:01T12:34:56.000Z" ] { a: null, b: 5, c: "abcd", d: "2016-01-01T01:01T12:34:56.000Z" } [] {}"
```

bkutil:
At the moment, we seem to have 3 different ways how to accomplish the same task - serializing a point into a string:

Je trochu nejasny, jak postupovat ohledne vystupu nested points:
1. `toString`
2. `inspect` - if we implement the above proposal from Rodney for `toString`, `inspect` can be dropped
3. `fromNative` - Mike suggested to keep the output structure of the point 'flat' - basically just keys + JSON for nested fields. This looks like a duplicate, too.

I'd propose folding those three implementations into a single one - `toString()`, which has:

1. human-readable output for flat and nested fields
2. represents the value uniquely, so it can be used as a group identifier

bkutil:
This was codified in a spec https://github.com/juttle/juttle/wiki/String-Representation-of-Juttle-Values and implemented in #200

PR #187, opened 2016-01-15
Fix displaying of objects and arrays in the table view

bkutil:
With the #200 in, remove the previously used workaround and simply rely on `toString`.

Fixes: #185


bkutil:
cc: @demmer @dmajda

bkutil:
This is a quick workaround before #177 lands

bkutil:
The issue was fixed after merging #200 - this pull request just reverts to the code w/o a workaround.

dmajda:
+1

bkutil:
Thanks :) Will merge on green.

Issue #184, opened 2016-01-14
Document nested fields


bkutil:
Add documentation of nested fields to http://juttle.github.io/juttle/concepts/fields/

Mention them also in  http://juttle.github.io/juttle/reference/data_types/ and link back to the fields article.

Also do a quick check and verify that `Array.length, Array.indexOf, Object.keys` work as expected for nested fields.

This is part of #55

PR #190, opened 2016-01-15
Add documentation for nested fields

bkutil:
Fixes: #184

### DIFF
--- a/docs/concepts/fields.md
+++ b/docs/concepts/fields.md
@@ -5,6 +5,9 @@ Each Juttle data point consists of a set of named fields. The name of a field is
 
 :construction: This needs to be updated to clarify stream context vs function context.
 
+[TOC]
+
+***
 
 Referencing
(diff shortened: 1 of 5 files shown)


bkutil:
cc: @dmehra

dmajda:
Looks good beside the terminology thing. I think @dmehra should still have a look.

bkutil:
Changed wording to 'fields with array/object values' or 'object/array fields', fixed spacing in example.

dmehra:
lgtm but @demmer and i agreed on inlining code examples if they are wrapped in text that discusses them. With that logic, we should put code snippets decorated with ```juttle into fields.md. They will get syntax checked.

bkutil:
@dmehra - I wasn't aware of that - I understood you aimed for having the examples in separate juttle files, so they are executable. Good to know - thanks.

PR #191, opened 2016-01-15
JuttleMoment compare returns -1, 0, 1

bkutil:
To be consistent with `values.js` compare and simplify the
implementation a bit, make `compare` for JuttleMoments return -1, 0, 1
for smaller, equal, or greater values.

Update the helper functions (`lt` / `gt`...) to handle new return values.

### DIFF
--- a/lib/moment/juttle-moment.js
+++ b/lib/moment/juttle-moment.js
@@ -173,27 +173,27 @@ var JuttleMoment = Base.extend({
     // Comparison convenience methods
 
     eq: function(operand) {
-        return JuttleMoment.compare('=', this, operand);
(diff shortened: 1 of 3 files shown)


bkutil:
cc: @dmajda

davidvgalbraith:
+1, this is long-needed functionality.

dmajda:
I’d combine the first two commits as changing `compare` without changing its caller isn’t atomic.

dmajda:

bkutil:
Thanks :) Squashed two non-atomic commits, removed the dead code. Will merge on green.

PR #192, opened 2016-01-15
Document example that extracts and converts time

bkutil:
Refs #166


bkutil:
This example converts time from milliseconds to seconds and extracting the time field. Also, while testing this approach, I noticed that handles fractional seconds just fine. So perhaps it is a problem with `timeField` ?

PR #197, opened 2016-01-15
Update emacs make cmd to use outrigger-client.

### LINKED ISSUE #198: Update emacs .el file to use outrigger-client push
Update the emacs juttle-derived-mode.el to suggest using outrigger-client push instead of the very old demo program.

mstemm:
Now that we have rendezvous, update the emacs "build" command to use
outrigger-client push, which is our preferred way of pushing programs
into juttled.

This fixes #198.

@mnibecker

### DIFF
--- a/misc/emacs/juttle-derived-mode.el
+++ b/misc/emacs/juttle-derived-mode.el
@@ -14,7 +14,7 @@
 ;; (setq juttle-run-dir "/home/<user>/src/juttle/bin")
 
 (defvar juttle-run-dir "")
-(defvar juttle-run-prog "demo")
+(defvar juttle-run-prog "outrigger-client push --topic mydev --path")
 
 (defvar juttle-derived-mode-hook nil "hooks")
 


misc/emacs/juttle-derived-mode.el
```diff
@@ -14,7 +14,7 @@
 ;; (setq juttle-run-dir "/home/<user>/src/juttle/bin")
 
 (defvar juttle-run-dir "")
-(defvar juttle-run-prog "demo")
+(defvar juttle-run-prog "outrigger-client push --topic mydev --path")
```
mnibecker: what's the deal with the empty `--path` arg?

mstemm: This is a prefix--the name of the program gets added afterward, and it's based on whatever buffer you're editing.

mnibecker: How will users know how view the output of the push?

mstemm: You mean the topic? They can override these in their own config file. If you mean what link they should go to afterward, the output of the program appears in a *compilation* buffer.

mnibecker:
cool +1

PR #199, opened 2016-01-16
single quote enforcement

### LINKED ISSUE #193: single quote enforcement
we've agreed upon internally that we should use a single set of quotes throughout our files to avoid ending up with files with mixed usage.

rlgomes:
fixes #193 

Initial PR to make our single quote usage consistent across our source.


rlgomes:
@demmer @davidvgalbraith @go-oleg @mnibecker @mstemm @dmajda @bkutil @mccanne @dmehra I know this is going to be somewhat controversial but at least the huge indentation part of tihs PR is in a separate commit that we can simply remove.

eslint made it easy to fix the single quote vs double quote inconsistency and also found a few real problems such as:

duplicate cases in a switch statement:

https://github.com/juttle/juttle/commit/823c32aca9896d8480c86ee51cbd039106d6d732#diff-2269d404e4061ee7837307476612c747R40

completely unused variables:

https://github.com/juttle/juttle/commit/823c32aca9896d8480c86ee51cbd039106d6d732#diff-beeabffc8a76322c8d367b80d8db1215L226

davidvgalbraith:
This is entirely in line with my aesthetic biases, so I give it a hearty +1.

bkutil:
Why not :) +1

dmajda:
So, this PR mixes up few things:

  1. Tooling change.
  2. Various kinds of changes in code related to that tooling change (removing of JSHint comments, fixes on now non-compliant code, etc.).
  3. Enforcing single quotes everywhere.

While 3 is a mechanical change which needs just skimming over, 1. and 2. require bit more careful review. However here these changes are mixed, sometimes even in the same commit, so I find a proper review of 1. and 2. impossible to do.

@rlgomes Could you please split the single quote change from the rest and put it into a separate PR (doesn’t matter whether before or after the switch to ESLint)? And could you arrange the other changes in more logical fashion? For example, I can’t really tell what the logical lines dividing the last 4 commits are — and this is not helped by things like having 2 commits with almost the same commit message doing vastly different things.

rlgomes:
@dmajda I'll split things up like you asked. I did this in a hurry but the reality is that only the commit to move from jshint + jscs to eslint was actual work as the rest was 90% done by `eslint --fix` with a few tweaks.

rlgomes:
Closing in favor of upcoming separate PR's to move to eslint and enforce single quotes.

PR #200, opened 2016-01-18
values.js - implement toString/inspect spec

bkutil:
Change the output of toString and inspect for Juttle values to conform
to the specification outlined here:

https://github.com/juttle/juttle/wiki/String-Representation-of-Juttle-Values

The spec is encoded into a set of accompanying tests (taken from the
specfication table).


bkutil:
cc: @dmajda

test/runtime/values.spec.js
```diff
@@ -133,4 +134,175 @@ describe('Values tests', function () {
...
+                {
+                    value: '',
+                    expected: ''
+                },
+                {
+                    value: 'abcd',
+                    expected: 'abcd'
+                },
+                {
+                    value: new RegExp(),
```
dmajda: I’d be slightly more explicit and do `RegExp('')`.

test/runtime/values.spec.js
```diff
@@ -133,4 +134,175 @@ describe('Values tests', function () {
...
+                {
+                    value: new JuttleMoment.duration('5', 'seconds'),
+                    expected: '00:00:05.000'
+                },
+                {
+                    value: new JuttleMoment('2015-01-01T00:00:05'),
+                    expected: '2015-01-01T00:00:05.000Z'
+                },
+                {
+                    value: new Filter({}, 'a < 5'),
```
dmajda: I’d slightly prefer to pass the AST too (see e.g. [here](https://github.com/juttle/juttle/blob/master/test/compiler/inputs.spec.js#L92-L103)).

test/runtime/values.spec.js
```diff
@@ -133,4 +134,175 @@ describe('Values tests', function () {
             });
         });
     });
+
+    describe('string representation', function() {
+        it('toString', function() {
```
dmajda: This `describe`/`it` combination is weird. What about something like this:

```javascript
describe('toString', function() {
    it('returns correct string representation', function() {
        // ...
```

dmajda:
All notes for `toString` tests also apply to `inspect`.

+1 with those addressed and tests fixed

bkutil:
Fixed tests, added the AST to filter, renamed tests.

PR #201, opened 2016-01-18
Purge utils.fromNative from views

dmajda:
Views should properly handle all Juttle data types by themselves, there is no need for the conversion.

### DIFF
--- a/lib/views/table.js
+++ b/lib/views/table.js
@@ -2,7 +2,6 @@
 var _ = require('underscore');
 var Table = require('cli-table');
 var View = require('./view.js');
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
It looks like `fromNative` is now dead code?

dmajda:
No, it is still used in `TestView`. The biggest blocker of its removal is changing JuttleSpec output, which I just filed an issue for (#208).

bkutil:
Cool, thanks :) +1 for this one.

PR #203, opened 2016-01-18
move to eslint

### LINKED ISSUE #202: move jshint + jscs to eslint
simplifiy our tooling by using just eslint to get the job done.

rlgomes:
fixes #202


rlgomes:
@dmajda let me know if this fits the bill for simply moving to eslint as there are still a few changes in here just to fix indentation issues that jshint + jscs were not detecting. There's also a few other things that eslint did find in terms of unused variables and duplicated switch cases.

The args:none for unused variables is needed since there are a few places we have unused function arguments and there's no reason to complain as the function is a "skeleteton" for others to implement.

rlgomes:
Also the fact I had to rerun `eslint --fix` was due to things that eslint coudlnt' fix blocking it from doing the auto fixing of other items in the same file. I'll squish all of those changes down to a single commit once this has been reviewed.

dmajda:
Overall, I think the changes look good except where I noted. And thanks for splitting up the changes.

My biggest gripe is the commit messages. In general, they should describe what the commit does — a good guide is to start them with a verb. Also, you occasionally use quite long lines In general, commit messages should be limited to 72 characters in order to work well with all the tooling. See [here](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for a formatting guide (which we probably should standardize on).

> I'll squish all of those changes down to a single commit once this has been reviewed.

Makes sense to me.

rlgomes:
@dmajda if you can give things another look and also should I further squash ? I thought keeping the commits that were done with the `eslint --fix` separate from the manual interventions.

dmajda:
The commits make lot of sense as they are now, thanks! I’ll just merge so the code doesn’t diverge form master.

Issue #168, opened 2016-01-13
add source base class support for reading


demmer:
To ease the implementation burden on adapters to support the flow control and coordinated scheduling proposal in #167, add support to the source base class for adapters for a simple producer / consumer model that supports both historical and pseudo-live queries.

The high level idea is that the `read` base class would have support for:

- processing and validating the `-from` and `-to` time options
- periodically calling into the adapter implementation to generate points into a buffer
- periodically emitting points out of this buffer based on interactions with the scheduler
- managing timers to implement periodic reading for pseudo-live processing

Then the adapter implementation only needs to handle the actual querying of the back end to generate points given a set of time bounds and a maximum buffer size.

davidvgalbraith:
Started thinking through some technical design here. Let's break it down function by function.

### Read methods

#### set_time_range(options) 
Similar to https://github.com/juttle/juttle-opentsdb-adapter/blob/master/lib/read.js#L70 and https://github.com/juttle/juttle-elastic-adapter/blob/master/lib/read.js#L67, takes `.from/.to/.last` out of `options` and sets them to be the time range for the query.

#### initialize(options, params)
`read` will have an `initialize` function that just sets `this.options = options` and `this.params = params`, plus any extra setup we want to do for everybody e.g. `set_time_range`. Adapters don't need their own `initialize`s.

#### emit(points)
`read` will overwrite `emit` to just buffer `points` in `this.buffer`. 

#### get_points(n)
Remove and return the first `n` points from `read`'s buffer.

### Adapter methods

#### fetch(n, offset)
Gets up to `n` points in the current time range and calls `emit` on them. Returns a promise that resolves when it's done fetching and emitting. `offset` says which point to start at: e.g., with `n` = 10 and `offset` 1000, the adapter would ignore the first 1000 points matching its time range and emit numbers 1001-1010.

For paging, the scheduler will use a ~~cheap hack~~ elegant algorithm like https://github.com/juttle/juttle-elastic-adapter/blob/master/lib/query.js#L147: fetch `n` points (`n`~10,000), drop the ones with the last timestamp, and call `set_time_range` to start the next query at that timestamp. The only case where we have to use `offset` is if all `n` points have the same timestamp. Then, we'll emit our first `n` and get the next `n` with `fetch(n, n)`. That's how we'll page through timeless points, too.

[...]

PR #206, opened 2016-01-18
source: add base class validation for -from, -to, and -last

demmer:
Add support in the base class to handle the various time bounds
options:

* ensure they are the appropriate types
* verifying that -from isn't greater than -to
* disallowing combinations of -from or -to with -last

Removed the scattered logic for the above from the adapters and from
the implementation of emit.

As a side-effect of this change, we now support `emit -last :x:`.

Relates to #168


demmer:
@davidvgalbraith @bkutil @dmajda

lib/runtime/procs/source.js
```diff
@@ -3,16 +3,74 @@
...
+
+    // Process the common time options (from / to / last) to ensure they are all
+    // the correct types and to convert `last` into the corresponding from/to
+    // combinations.
+    //
+    // Stashes the time bounds in `this.from` and `this.to`.
+    handleTimeOptions: function(options) {
+        if (options.from) {
+            if (! options.from.moment) {
+                throw this.compile_error('RT-MOMENT-ERROR', {
```
davidvgalbraith: `function throwMomentError(option) { throw self.compile_error('RT-MOMENT-ERROR', {option: option, value: values.inspect(options[option])})}` could DRY this up a bit.

davidvgalbraith:
+1

dmehra:
Please update docs/sources/emit.md with `-last :x:` option.

bkutil:
Comparing the code to [time range semantics](https://github.com/juttle/juttle/wiki/Time-Range-Semantics) doc, I noticed that we don't handle the 'all of/none of `to/from/last` is present at the same time' case. Should this be part of `source`, or is that something that the subclasses should take care of?

If the former, two test cases covering that would be useful, too...

demmer:
@bkutil I can add a test for the error case where all three options are present.

As far as what to do if none of the options are present, there was some discussion after the fact about how the current proposal doesn't do a good job of handling adapters like file / http / sql that don't necessarily have a mandatory time field in the query, or cases like `emit` where we want the default to just emit a single point (i.e. it should default to `-from :now: -limit 1`).

It seems like we have a couple options here:

(A) Keep the current default proposal and live with the problems outlined above

(B) Change the default for `-from` to be the dawn of time, i.e. `:1970-01-01:`. This would work well for timeless sources or for simple cases where we want to read all the data out of a query, but could of course end up fetching a lot of data.

(C) Allow the default behavior to be set by each source independently, so it can do what makes the most sense given its constraints. Each adapter would need to document what its defaults are when the time fields aren't specified.

Also if we go with either A or B we will need to decide if `emit` should follow the same rules or whether it should be treated specially.

I'm leaning towards (C) which happens to also be what's implemented in this PR. @dmajda / @dmehra / @bkutil / @davidvgalbraith / @mstemm / @VladVega / @rlgomes / @go-oleg do you guys have thoughts on this?

dmehra:
I'm thinking (C) but have a default (-from :0: -to :now:) that applies unless the adapter overrides it. Would that work?

mstemm:
I also vote for C. I think it's going to be difficult to impose one set of behavior on an increasing set of adapters talking to a more heterogenous set of backends.

demmer:
@dmehra I'm not sure what you mean by that -- if we go with (C) then there needs to be some differentiation between the case where the user explicitly wrote `-from :0: -to :now:` vs omitted the option altogether.

VladVega:
I actually think A:
Throw is `from

go-oleg:
C sounds good to me too.

Re what @dmehra said, I agree that it would be nice to somehow encourage consistency if possible (since we can't enforce it). Having the base class implement a function that returns the default values and letting subclasses override or pull the defaulting out into a separate util function and encourage people to use in the "how to write your own adapter" docs are options.

dmajda:
I think we have 3 classes of sources:

  1. Those where the default value of both `-from` and `-to` should be `:now:` and at least one of these options needs to be specified. This is the case of most DB-backed adapters.
  2. Those where the default value of `-from` should be `:beginning:`, the default value of `-to` should be `:end:`, and both can be omitted (producing a query encompassing all time). This is the case of `file`, `http`, etc.
  3. Those which use `-from` and `-to` to influence how points are generated, having a special semantics. This is the case of `emit`.

IMO it would make sense for a source to say into which class it belongs and have `handleTimeOptions` enforce the right semantics. This would enforce consistency across adapters. For the first two classes this semantics is pretty clear, for the third class we may want to do just some minimal checks and leave the rest to the proc itself.

If we agree on the above, I’ll adapt the time range semantics docs appropriately.

The fact that a source belongs to one of these classes can be indicated in various ways (e.g. a subclass of `source`, reincarnation of the `sourceType` property, or a parameter to `handleTimeOptions`). I don’t have a strong preference there.

As a intermediate step toward all this, (C) sounds fine to me.

demmer:
I'm good with that proposal -- FWIW think that all the sources in the second class should also require a `timeField` option to use from and to.

-m

bkutil:
> The fact that a source belongs to one of these classes can be indicated in various ways (e.g. a subclass of source, reincarnation of the sourceType property, or a parameter to handleTimeOptions). I don’t have a strong preference there.

Rather than relying on a property/inheritance check, we could provide three modules/classes for validation (exposing a single `validate(from, to, last)` method), one for each category, and make sure they conform to the spec.

These modules could be injected into the adapter either during initialisation, or instantiated in adapter's `initialize` and passed as a parameter to `handleTimeOptions`.

This way, the validations would be easily unit-testable and if we decide to change the validation logic for `to`/`from`/`last`, we don't have to change the adapters.

demmer:
I just pushed up a rework of the implementation that implements the model proposed by @dmajda in which sources fall into one of the three categories.

@bkutil Rather that complicate things with a module injection for something this simple, I opted to change the contract so that `handleTimeOptions` is explicitly called by the source and takes a constant option to indicate how to handle the defaults.

davidvgalbraith:
I don't want to burden the adapters with having to know about all this `DEFAULT_TIME_RANGE` flimflam. How about the default default is that it throws an error if neither `from/to` nor `last` is specified, and if an adapter wants to opt into one of these fancy defaults it can use one of the `DEFAULT_TIME_RANGE` options.

VladVega:
One thing that could be addressed here in addition:
Required options: Pass in an array of required options as a result of all the logic in the function. For example If `this.to` is empty, `throw`.

dmajda:
> I don't want to burden the adapters with having to know about all this `DEFAULT_TIME_RANGE` flimflam.

I’d prefer the adapters to be explicit about what behavior they want, unless we have like 80% adapters swinging one way, in which case the default would probably start making sense.

bkutil:
> takes a constant option to indicate how to handle the defaults.

I don't think injection would be that complicated, or rather - the benefit it would bring - easily testable code that clearly corresponds to the specs - would outweigh the extra two/three classes.

demmer:
@bkutil @davidvgalbraith I modified the implementation to split out the parsing of the options from the setting of defaults into two separate calls.

Given that the eventual goal of these is to set `this.from` and `this.to` I'm still not feeling like this is worthy of a separate module. @bkutil if you feel like it's worthwhile then go ahead and I can see what I think about it.

Note that when adapters change from being direct proc subclasses (as described in #241) then the actual contract on the adapter may not exactly match what's being done here.

IMO this is enough of an improvement that we can merge this as-is and then continue to refine.

davidvgalbraith:
Looks great! +1 if you can get it to work.

demmer:
Cleaned up and pushed a new PR in #242.

PR #207, opened 2016-01-19
view options: emit moment.duration instead of seconds

go-oleg:
This is necessary as part of the work we are doing with [jsdp](https://github.com/juttle/juttle-jsdp) so browser views can be instantiated with real values (instead of numeric or string representations) on the other side of the websocket.

@demmer @dmajda

### DIFF
--- a/lib/runtime/procs/procs.js
+++ b/lib/runtime/procs/procs.js
@@ -208,7 +208,7 @@ Juttle.alloc_channel = alloc_channel;
 function convert_options(o) {
     if (o instanceof JuttleMoment) {
         if (o.duration) {
-            return o.valueOf();
+            return o.duration;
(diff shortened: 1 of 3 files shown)


go-oleg:
Preferably this can get into the 0.3.0 release of juttle.

dmajda:
As discussed in private, I believe moments and durations should be represented by `JuttleMoment` everywhere inside Juttle. So my preferred solution here would be to kill `convert_options` completely and if any conversions need to be done in order to send data over the wire, do them in Outrigger.

go-oleg:
Sure, that sounds good for consistency with values that are emitted in the points themselves.

Closing this, opened https://github.com/juttle/juttle/pull/210.

PR #209, opened 2016-01-19
Add json module

bkutil:
A draft for implementing Json module for Juttle.

The reasoning behind introducing a `Json` module is that adapters can produce serialised JSON values, which the user might want to parse, e.g. in the cases when the JSON type cannot be inferred from the schema.

There are two open questions:

1. how should we report parse errors from JSON parser? As a warning?
2. there seems to be a bug in juttle-spec dealing with nested objects, as mocha outputs (this can be observed when running the tests):
```
    -      {
    -        "key": "value"
    -      }
    +      [
    +        [Function]
    +      ]
```

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -48,6 +48,7 @@ var SemanticPass = Base.extend({
         this.import_date_module();
         this.import_duration_module();
         this.import_object_module();
+        this.import_json_module();
         this.import_juttle_module();
     },
     import_math_module: function() {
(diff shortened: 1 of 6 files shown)


bkutil:
cc: @dmajda - could you please take a look if this is moving in the right direction?
cc: @demmer

dmajda:
> how should we report parse errors from JSON parser? As a warning?

The module function should throw a Juttle exception with proper code (i.e. it should wrap the parser exception). This means that when parsing blows in compile-time, an error based on this exception will be produced, and when it blows in runtime (e.g. in `put`), a warning & point drop will automagically occur.

> there seems to be a bug in juttle-spec dealing with nested objects, as mocha outputs (this can be observed when running the tests)

Please file an issue with a minimal test case and feel free to either fix this yourself or assign to me.

> could you please take a look if this is moving in the right direction?

It is.

bkutil:
Fixed the throw, renamed to `JSON`, the other issues were fixed in referenced PRs.

bkutil:
Lint fixes, squashed the commits to be atomic.

dmajda:
+1

PR #210, opened 2016-01-19
view options: don't use convert_options

go-oleg:
Emit view options as they are, instead of converting moments to JS Dates and moment durations to seconds.

This is necessary as part of the work we are doing with jsdp so browser views can be instantiated with real values (instead of numeric or string representations) on the other side of the websocket.

@dmajda cc @demmer

### DIFF
--- a/test/compiler/views_sourceinfo.spec.js
+++ b/test/compiler/views_sourceinfo.spec.js
@@ -2,6 +2,7 @@ var expect = require('chai').expect;
 var compiler = require('../../lib/compiler');
 var views_sourceinfo = require('../../lib/compiler/flowgraph/views_sourceinfo.js');
 var juttle_test_utils = require('../runtime/specs/juttle-test-utils'); // jshint ignore: line
+var JuttleMoment = require('../../lib/moment').JuttleMoment;
 
 describe('Views get info on source time bounds', function() {
     function test(juttle, spec) {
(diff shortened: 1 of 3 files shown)


test/compiler/views_sourceinfo.spec.js
```diff
@@ -37,19 +38,19 @@ describe('Views get info on source time bounds', function() {
...
-    var last = '01:00:00.000';
+    var from = new JuttleMoment({ raw: '2015-01-01T00:00:00.000Z' });
+    var to = new JuttleMoment({ raw: '2015-02-02T00:00:00.000Z' });
+    var last = new JuttleMoment.duration(1, "hour");
 
     test('read stochastic -source "cdn" | view view ', [[{from: null, to: null, last: null}]]);
-    test('read stochastic -source "cdn" -from :' + from + ': | view view ', [[{from: from, to: null, last: null}]]);
-    test('read stochastic -source "cdn" -from :' + from + ': -to :' + to + ': | view view ', [[{from: from, to: to, last: null}]]);
-    test('read stochastic -last :' + last + ': -source "cdn" | view view ', [[{from: null, to: null, last: last}]]);
+    test('read stochastic -source "cdn" -from :' + from.toJSON() + ': | view view ', [[{from: from, to: null, last: null}]]);
```
dmajda: I think `valueOf` would be slightly cleaner than `toJSON` here. This is what the compiler, `Date.toString`, and other code uses and in fact `toJSON` delegates to it.

dmajda:
+1 beside the `valueOf` note

test/compiler/views_sourceinfo.spec.js
go-oleg: Ah, of course.

PR #211, opened 2016-01-19
CSV stdio serializer supports arrays/objects

bkutil:
Serialize array/object values using JSON, to keep point structure flat
and processable by CSV writer.

### DIFF
--- a/lib/adapters/serializers/csv.js
+++ b/lib/adapters/serializers/csv.js
@@ -2,6 +2,7 @@ var _ = require('underscore');
 var base = require('./base');
 var csv = require('csv-write-stream');
 var errors = require('../../errors');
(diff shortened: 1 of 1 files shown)


bkutil:
cc: @dmajda

refs: #55

rlgomes:
@bkutil looks good I do think at some point we need to come up with a nicer way of apply *P* functions to every point or *F* functions to every field in a point in a way that doesn't mean having to iterating through the same *N* points a bunch of times. This just occurred to me given how we're currently doing a few of these *N* loops for adding `timeField` and now applying this values thing to each field.

rlgomes:
+1

dmajda:
+1

PR #212, opened 2016-01-19
Add a few more tests for array/object fields

bkutil:


bkutil:
cc: @dmajda

refs: #55

dmajda:
I don’t see any value in `keep` and `remove` tests as these procs don’t care at all about what is stored in the fields they manipulate. I’d just omit them.

With that, +1.

bkutil:
Removed the two unnecessary tests, will merge on green.

PR #216, opened 2016-01-20
Update CHANGELOG.md for 0.3.0

dmajda:


dmajda:
@bkutil

bkutil:
Looks good. +1

CHANGELOG.md
```diff
@@ -1,34 +1,71 @@
...
 
+This file documents all notable changes to Juttle. The release numbering uses [semantic versioning](http://semver.org).
+
+## 0.3.0
+
+Released 2016-01-20
+
+### Major Changes
+
+- Added support for points with array/object field values. [#55]
```
bkutil: Would it make sense to have issue references as links?

dmajda: I know @demmer didn’t want them. I think the main reason was they would be too much hassle to add. But given this is mostly one `s/.../.../` command in Vim I added a commit which adds them and will add them in similar way in the future.

CHANGELOG.md
```diff
@@ -1,34 +1,71 @@
...
+
+## 0.3.0
+
+Released 2016-01-20
+
+### Major Changes
+
+- Added support for points with array/object field values. [[#55](https://github.com/juttle/juttle/issues/55)]
+- Added support for Grok parsing to `file` and `stdio` adapters (`-format "grok" -pattern ...`). [[#37](https://github.com/juttle/juttle/issues/37)]
+- Adapters are now loaded only when they are used. [[#83](https://github.com/juttle/juttle/issues/83)]
```
demmer: This also changed the way adapters are configured -- we should mention that as well.

PR #224, opened 2016-01-21
http_server adapter

### LINKED ISSUE #111: http read -listen
_Wouldn't it be great if you could use to power of juttle to set up an http service and write incoming points to your favorite storage backend?_

Proposing listen mode for the builtin `read http` adapter. A simple program would look something like this:

```
read http -listen 2000
| write elastic
```

With this program now running, we can now do a `POST` request to `localhost:2000`:

```
customer_id, product_id, price
12, 1209, 3.00
1, 1400, 5.00
```

### Revised `read http` options:

I'm proposing renaming the `-url` option to `-fetch`. The `-listen` option takes a port number. Either the `-fetch` option or `-listen` option must be specified. If both are specified an error is thrown.

If `-listen` is specified on the `-method` and `-timeField` options will apply. All other fields applicable to `-fetch` will be ignored.

Parameter         |             Description          | Required?
----------------- | -------------------------------- | ---------:
`-fetch`            | URL to issue the HTTP request at |  Yes: either `fetch` or `listen` must be specified 
`-listen` | Port for http endpoint to listen on | Yes: either `fetch` or `listen` must be specified 
`-method`         | HTTP method to use               | No; `-fetch` default: `GET`; `listen` default: `POST`
`-headers`        | headers to attach to the HTTP request in the form `{ key1: "value1", ..., keyN: "valueN" }` | No; default: `{}`
`-body`           | body to send with the HTTP requests | No; default: `{}`
`-timeField`      | The name of the field to use as the time field <br><br>The specified field will be renamed to `time` in the body of the HTTP request. If the points already contain a field called `time`, that field is overwritten. This is useful when the source data contains a time field whose values are not valid time stamps.  | No; defaults to keeping the `time` field as is
`-includeHeaders` | When set to true the headers from the response are appended to each of the points in the response | No; default: `false`

### Content Types

Just like `read http -fetch`, `read http -listen` accepts `text/csv` and `application/json` `content-type` headers. Requests without a `content-type` header or with a different `content-type` value will recieve a `400` bad request response.

#### JSON

JSON requests will be parsed in the same fashion as `read http` fetch. Any serialization errors will be returned as a `400 BAD REQUEST`

#### CSV

Each CSV request needs to have a header row with the keys for the following points.

### Implementation Details

I'd like to use the node stdlib to create the server, using `require('http')` as opposed to 3rd party libraries like express. The goal would be to minimize dependencies.

mnibecker:
Take `-listen` functionality described in #111 and move it all to its own adapter called `http_server`.

Supports the following options:
- `port` - Not required, defaults to `8080`. The port number the server will listen to
- `method` - Not required, defaults to `POST`. Defines the accepted http method for the created endpoint. Currently only supports `PUT` or `POST`.
- `timeField` - Not required, defaults to `time`. Defines the field to treat as time.

Implementation detail: Points do not actually get emitted until client stops sending server data. This is because I wanted to return a 400 error message if the request is not perfect. I'd be curious to hear any objections over this.

Resolves #111


mnibecker:
@rlgomes @demmer

mnibecker:
just tested this from `cli`, it works and its totally awesome!

rlgomes:
1. lets stick to just single quotes so the soon to be single quote enforcement has to touch less things.
2. I think it would be a nice feature to be able to specify the HTTP method and path to accept requests on and also the ability to dictate which address to bind on. This could simply be handled by acccepting a string to `-listen "http://172.22.23.3:8000/some/special/path" -method DELETE` because who knows if someone wants to record all delete requests they're sending to some other product by splitting the HTTP traffic and hitting an HTTP listener on our end ? (I'm sure there are other good uses for this).

For 2. we could simply create other issues to fill that kind of functionality in I just wanted to bring it up to see if others find that to be a nice thing to have or not.

mnibecker:
@rlgomes, agree with the path functionality I'll add that. As for supporting different methods, I'm kind of scratching my head at that. How would `GET` work? `GET` doesn't have a body, you just insert the query params as a point? How would `PUT`/`PATCH` work? `PUT`/`PATCH` are supposed to modify currently existing properties, I'm not quite sure the use here. Same goes for `DELETE`.

rlgomes:
@mnibecker I know that I've seen different people doing things that aren't really the "correct" way to use HTTP which does include doing GETS with a body and using PUT instead of  POST as you're directly correctly here.

mnibecker:
I'm going to do this with `POST` first, because its what I want. I think we can follow up in a separate pr to add other methods. @go-oleg made some good comments about supporting `PUT` as well. I personally would find it confusing if it did a `GET` request and the server was expecting content in the body as opposed to query params.

mnibecker:
Alright moved everything to its own adapter called `http_server`. Updated pr to reflect this.

@rlgomes- I added support for both `POST` and `PUT`. As for supporting other methods I think we can attack this on a separate pr. Same for support path (we discussed this here and couldn't really think of a good use case for path- you can't create multiple listeners on the same port).

rlgomes:
+1 looks good just missing docs like the other adapters under: 

https://github.com/juttle/juttle/tree/master/docs/adapters

mnibecker:
I will do this and also update the changelog

lib/adapters/http_server.js
```diff
@@ -0,0 +1,133 @@
...
+            throw this.compile_error('RT-ADAPTER-UNSUPPORTED-FILTER', {
+                proc: 'read http_server',
+                filter: 'filtering'
+            });
+        }
+    },
+
+    start: function() {
+        var self = this;
+        var format;
```
davidvgalbraith: With `format` scoped here, it will be preserved across requests, so if you make a request with one content type and then the next request's content type throws an error on line 56 `format` will be the old content type instead of `undefined`.

lib/adapters/http_server.js
```diff
@@ -0,0 +1,133 @@
...
+
+        this.server = http.createServer(function(req, res) {
+            try {
+                format = contentType.parse(req).type.split('/')[1];
+            }
+            catch(err) {
+                // type is undefined
+            }
+            try {
```
davidvgalbraith: This should be wrapped in `Promise.try` so you don't have to have separate `catch`es for sync/async errors.

lib/adapters/http_server.js
```diff
@@ -0,0 +1,133 @@
...
+                if (format !== 'json' && format !== 'csv') {
+                    throw self.runtime_error('RT-INVALID-TYPE', {
+                        types: 'csv, json'
+                    });
+                }
+
+
+                var emit_points = [];
+
+                var parser =  parsers.getParser(format , {rootPath: self.rootPath});
```
davidvgalbraith: `rootPath` isn't an allowed option?

lib/adapters/http_server.js
```diff
@@ -0,0 +1,133 @@
...
+                parser.parseStream(req, function(points) {
+                    if (!_.isArray(points)) {
+                        points = [points];
+                    }
+
+                    self.parseTime(points, self.timeField);
+                    emit_points = emit_points.concat(points);
```
davidvgalbraith: Why buffer all the points like this instead of emitting them here?

test/adapters/http_server.spec.js
```diff
@@ -0,0 +1,298 @@
...
+            { 'name': 'ted', 'age': 53, 'weight': 100},
+            { 'name': 'ted', 'age': 53, 'weight': 100}
+        ];
+
+        return compile_juttle({
+            program: 'read http_server'
+        })
+        .then(function(prog) {
+            program = prog;
+            programFinish = run_juttle(prog);
```
davidvgalbraith: This is a neat testing strategy!

lib/adapters/http_server.js
mnibecker: yep, great point

lib/adapters/http_server.js
mnibecker: I suppose I should add `rootPath`

lib/adapters/http_server.js
mnibecker: Did not know that's how `Promise.try` worked. Hot tip.

lib/adapters/http_server.js
mnibecker: Because if there's an error parsing a line I wanted to throw back a `400` and reject the entire request. The alternative was returning a `200` and not ingesting some of the points, which felt weird to me.

mnibecker:
Great comments @davidvgalbraith. I'll throw up a pr that addresses your comments/fixes

PR #226, opened 2016-01-21
updating to our own node-grok fork

### LINKED ISSUE #225: node-grok pulls in collections with buggy version
discovered by @oleg the problem is explained here: 

https://github.com/montagejs/collections/issues/139

we'll temporarily fork node-grok and depend on a version there where collections@^2.0

rlgomes:
fixes #225

due to issues in a package dependency of node-grok this is a quick fix
to depend on our own fixed version of the node-grok source.


go-oleg:
looks good

Issue #167, opened 2016-01-13
add time-synchronization


demmer:
_This proposal was originally suggested by @mccanne in an internal design document._

The current juttle computation model is entirely data-driven and uncoordinated. Sources fetch or generate data autonomously according to their own rate control policy (if any) which may be as fast as possible, then push that data into the flowgraph. As points arrive, procs process them and make decisions about doing stuff. Marks and ticks communicate batching and timing information between nodes in a data-driven fashion.

This model has certain problems and limitations:

**Flow control**: There is no way for feedback from the flowgraph or sinks to slow down senders.  If sources are faster than sinks, the memory buffers fill up (until they hit a limit).

**Livelock**: Because senders are unconstrained, they may generate a large batch of points that need to be processed by a downstream consumer. This can cause the (single-threaded) node.js runtime to become busy for unacceptably large amounts of time.

**Merge / Join Buffering**: When multiple sources produce data at different rates, the fanin logic will use an ever increasing amount of memory to account for the rate difference in order to merge the streams to preserve time order.  This problem can be more and more apparent when mixing and matching different sources using various backend adapters.

## Approaches

Various approaches to fix this problem have been discussed during the development of Juttle that have significant downsides:

One approach would be to do hop-by-hop flow control.  This would involve a fair amount of new code in the procs and complex dynamics to debug, especially when it comes to fanout and fanin.

Another idea is to get rid of marks and ticks and drive everything off time using a scheduler, which would schedule all of the point arrivals at each proc in a sort of discrete event simulation.  This would add processing overhead for each point (insert/delete into a heap) and seems like overkill.

Instead, Juttle should adopt a hybrid approach:

Keep the data driven model and communicate timing information between flowgraph nodes using the existing tick and mark abstractions, but add a scheduler to precisely control how and when points are sent into the flowgraph. Flow control can be managed by interacting with the scheduler to start and stop certain sources in response to whole program signaling.

In this approach, we need only modify the existing sources and sinks and don't have to change any of the other proc implementations. The internal procs are still all data driven and run off the arrival of points, ticks, and marks, and don't know if they're running in "historical" or "live" mode.

## Sources

The gist of the scheduling algorithm is that in steady-state, each source adapter manages a bounded buffer of points that are ready to be sent into the flowgraph based on shared logic that processes the time bounds of the query (i.e. the `-from` and `-to` arguments to `read`). On startup, the adapter starts filling the buffer and schedules itself to emit based on the timestamp at the head of the queue.

### Historical

[...]

PR #227, opened 2016-01-21
source base class read logic

demmer:
NOT FOR MERGE

Initial cut at an implementation of the source base class support described in #168 for reading data from the subclass using a new `read(from, to, limit, state)` API function that is responsible for actually fetching or generating the data.

This should enable support for paging and pseudo-live queries in each of the adapters and is amenable to future integration with the flow-control scheduler described in #167. Also all the management of ticks is handled by the base class as opposed to having to be scattered through all the sources.

In this PR I also reimplemented `emit` as if it were an adapter using this contract. However given the unusual semantics that emit has for its behavior with -limit, -from, and -to options, this turned out to be a bit complicated in parts and seems like a step backwards in clarity from the older implementation.

I still haven't given up hope though -- maybe it makes sense to change some of the emit expectations to be closer to that of other sources... still TBD.


demmer:
@davidvgalbraith @bkutil @dmajda @mstemm @VladVega Some initial feedback would be great -- I'd like to follow up with an implementation of one of the real adapters (influx probably) using this model to see that it actually works in practice with something real.

dmajda:
On the high level I like what is in the `source` class (I’d have some small comments about naming and splitting things up a bit more but I’ll hold them at this phase).

Regarding state passing, I realized a cleaner way to implement this could be an iterator. In TypeScript-like syntax:

```typescript
class Iterator {
    // Here goes the iteration state, e.g. next offset in SQL adapters.
	
    getNext(): Promise<Point[]> | null {
        // ...
    }
}

class Adapter {
    read(from, to, n): Iterator {
        return new Iterator(...);	
    }
}
```

Inside `read` one would do something like:

```javascript
var iterator = adapter.read(...);

// This loop would really be async & promisified, use the scheduler, etc.
while (points = iterator.getNext()) {
    this.emit(points);
}
```

This would eliminate the ugliness of calling `read` once with empty state and then with a non-empty one.

As for `emit`, the problem with `to` and stopping the live iteration could be avoided by adding some other signaling path, e.g. adding `eof: true` to the result. While this would work, I’m actually starting to suspect that `emit` is just too different from the adapters. For adapters, one wants to implement live using *pull* while the natural thing for emit is *push*. Maybe it makes most sense to split the implementation so that:

  * `source` would contain the scheduler and a way to enqueue incoming points (that part coming later, for now it would just emit them)
  * `emit` would just push points to the enqueuing API at the right time, i.e. the implementation would not change much from today
  * `read` would additionally contain logic for pseudo-live, lagging, etc. and would call `read` on adapters

I won’t claim I thought through all the consequences though.

demmer:
Regarding iterators I thought about that idea but it seemed like it would make the simple adapters that don't actually do paging overly complicated.

My high level idea was to actually use that kind of pattern inside the adapter -- i.e. a given adapter could actually return an iterator object as the state variable and then easily dispatch to its methods on subsequent calls to read.

As far as how to handle emit I feel like I came to the same conclusion -- it's just different enough that it doesn't make sense to couple them together.

I would like to move the tick generation logic into source but other than that I think we agree on the breakdown of responsibilities.

bkutil:
> one of the real adapters (influx probably)

I'd like to take care of that, as soon as the approach is clear

demmer:
Closing in favor of #248

Issue #208, opened 2016-01-19
Change JuttleSpec output specification format to native Juttle objects


dmajda:
We use a weird format to describe Juttle output in JuttleSpec. It’s basically a series of JSON objects separated by whitespace, but with a special tweak that the objects can have unquoted keys and allow `Infinity` and `NaN` values. On the other hand, there is no representation of native Juttle values such as moments and durations, which are entered as strings (the points coming from Juttle go through `fromNative` before being compared to the expected output). We had to invent [`points-parser`] (https://github.com/juttle/juttle/tree/master/extlib/points-parser) to parse this thing.

Let’s adjust the format so that the points are described by Juttle object literals (which allow native syntax for moments and durations) with one object per line (this is similar to [JSON Lines](http://jsonlines.org/)). This will allow us to get rid of `points-parser`.

PR #230, opened 2016-01-21
To json compatible

bkutil:
Converting juttle values to JSON compatible representation occurs
frequently enough to warrant a 'canonical' implementation.

Add `values.toJSONCompatible` and switch `JuttleUtils.fromNative` to use it.

refs: #209

### DIFF
--- a/lib/runtime/juttle-utils.js
+++ b/lib/runtime/juttle-utils.js
@@ -25,19 +25,7 @@ function toNative(records, epsilon) {
 }
 
 function fromNative(records) {
-    function convert(record) {
-        _.each(record, function(value, key) {
-            if (values.isDate(value) || values.isDuration(value)) {
-                record[key] = value.valueOf();
-            } else if (values.isObject(value) || values.isArray(value)) {
-                convert(value);
-            }
-        });
-    }
-
-    _.each(records, convert);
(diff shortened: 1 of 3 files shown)


bkutil:
cc: @dmajda

bkutil:
`fromNative` being JSON-compativle` will affect the fixing of https://github.com/juttle/juttle/issues/208

bkutil:
Fixed indentation, added tests covering identities (handling of NaN and Infinities was fixed to return strings - open question is whether `fromNative` should use it then), updated the function to use `_.map` and `_.mapObject`.

dmajda:
I’d actually say that `NaN` and `Infinity` are JSON-compatible, but they are serialized to `null`. Compare e.g. with a function, which is not JSON-compatible:

```
> JSON.stringify(Infinity);
'null'
> JSON.stringify(function(){});
undefined
> JSON.stringify({a: Infinity});
'{"a":null}'
> JSON.stringify({a: function(){}});
'{}'
```

So I’d revert the `Infinity`/`NaN` change. With that, +1.

bkutil:
Removed the special handling of `NaN` and infinities. Will merge on green. Thanks :)

PR #231, opened 2016-01-21
juttle-spec: Fix handling of nested arrays/objects

bkutil:
When generating test expectation, `util.inspect` was used with the default
level of depth 2. This lead to fake test failures when the expected
object was deeper than that and its values were represented by a
placeholder, e.g. `[Object]` or `[Array]`.

Pass `depth: null` parameter to `util.inspect` for infinite depth.


bkutil:
cc: @dmajda 

refs: #209

dmajda:
+1

PR #232, opened 2016-01-21
Change FTS sytnax to use the ?= operator

### LINKED ISSUE #56: Add a ? operator for FTS
Instead of naked strings implying FTS, we should have an operator for the expressions to avoid confusion, especially now that most backends won't support it. (Currently, only `elastic` and `gmail` adapters support FTS.)

- [x] Add spec / design doc (move content from Confluence)
- [ ] Modify grammar
- [ ] Update docs
- [ ] Update examples

dmajda:
Before this commit, FTS was expressed by using standalone strings in
filter expressions:

    read elastic "keyword"

This commit changes the syntax to use a new `?=` operator:

    read elastic ?= "keyword"

The change includes updating tests and documentation. See the commit message for more details.

Resolves #56.

### DIFF
--- a/docs/concepts/filtering.md
+++ b/docs/concepts/filtering.md
@@ -107,21 +107,19 @@ present in any field.
 The [filter](../processors/filter.md) processor does not support full-text search -- it can only be used as part of a read from an external backend that supports search.
 
 
-The search terms for full-text search are currently expressed as standalone strings in the filter expression for `read`. Search terms and other filter expressions can be combined with the AND, OR, NOT operators.
-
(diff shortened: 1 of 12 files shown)


dmajda:
@demmer 

I ended up actually not liking the change much, e.g. the examples in the docs now don’t feel very natural.

I also realized the `?=` operator we agreed upon is actually inconsistent with the matching operator (`=~`) in the position of the `=` sign. But `=?` looks even more odd.

dmajda:
Scrapped per https://github.com/juttle/juttle/issues/56#issuecomment-177901363.

Issue #165, opened 2016-01-13
Streamline error-related code


dmajda:
* [x] Kill `BaseError` (it just duplicates `JuttleError`)
* [x] Remove unnecessary cruft from `JuttleError` and its child classes
* [x] Simplify the strings file
  * [x] No top-level `error` key
  * [x] Use single `{...}` for interpolation
  * [x] No `info.` prefix in messages
  * [x] Rename the file
  * [x] No `Error: `/`Warning: ` prefix in message strings
  * [x] No `RT-` prefix in message keys (either replace by `JUTTLE-` or completely remove)

The idea is to do these changes before #59 so that the resulting API for backends is simpler and more stable.

PR #234, opened 2016-01-21
Simplify error strings file

dmajda:
Simplify various small things in the error strings file.

Part of work on #165.


dmajda:
@bkutil

bkutil:
+1

PR #235, opened 2016-01-21
Kill FilterSearcher

dmajda:
It became dead code.

While it is conceivable that adapters may need a similar helper class at some point, let’s not predict the future and implement it when needed, possibly with more suitable API.


dmajda:
@bkutil

bkutil:
For the record, the code was already unused when the project was open-sourced.

+1

PR #236, opened 2016-01-21
updating coverage numbers

rlgomes:
we seemed to have received another bump here so lets make sure to keep
the numbers up


rlgomes:
@juttle/developers quick +1 please

go-oleg:
higher!

looks good

rlgomes:
will revert since master dipped below 92% on function coverage and then put up a new branch.

PR #242, opened 2016-01-21
source base class time options

demmer:
Add support in the base class to handle the various time bounds
options:

* ensure they are the appropriate types
* verifying that -from isn't greater than -to
* disallowing combinations of -from or -to with -last

Also add support for setting defaults for from / to with an optional
constraint that one of them must be supplied.

Removed the scattered logic for the above from the adapters and from
the implementation of emit.

As a side-effect of this change, we now support `-last :x:` in both
`emit` and the file adapter.

This is the same code as #206 but squashed and using a more appropriate branch name.

Relates to #168


demmer:
@davidvgalbraith After a long long slog through travis I've got a green build. Maybe take a final look at this one before I merge?

davidvgalbraith:
Still looks good!

Issue #243, opened 2016-01-21
Reading from CSV doesn't support -timeField


dmehra:
Problem first reported [on gitter by davidbcook](https://gitter.im/juttle/juttle?at=56a151c82bc35f6c1c1b7b51)

When you read from a CSV file and ask juttle to look at a custom timeField, the results are very strange because CSV has no understanding of types, and we end up treating the fields intended to be unix timestamps since epoch as strings, then converting them to juttle moments and ending up with crazy values.

```
$ cat /tmp/data.csv 
mytime,a
1,a0
267000,a1
267001,a2
268000,a3
```

```
$ juttle -e "read file -file '/tmp/data.csv' -timeField 'mytime' -format 'csv'"
┌─────────────────────────────────────────┬──────────┬──────────┐
│ time                                    │ a        │ mytime   │
├─────────────────────────────────────────┼──────────┼──────────┤
│ 2001-01-01T08:00:00.000Z                │ a0       │ 1        │
├─────────────────────────────────────────┼──────────┼──────────┤
│ +267000-01-01T08:00:00.000Z             │ a1       │ 267000   │
├─────────────────────────────────────────┼──────────┼──────────┤
│ +267001-01-01T08:00:00.000Z             │ a2       │ 267001   │
├─────────────────────────────────────────┼──────────┼──────────┤
│ +268000-01-01T08:00:00.000Z             │ a3       │ 268000   │
└─────────────────────────────────────────┴──────────┴──────────┘
```

Workaround:

Don't use -timeField with CSV format, and instead do time assignment using `Date.new()`. Note that it expects seconds, so if your file had millisecond time values, do the conversion as below.

```
juttle> read file -file '/tmp/test.csv' -format 'csv' | put time = Date.new(Number.fromString(mytime)/1000)
┌────────────────────────────────────┬──────────┬──────────┐
│ time                               │ a        │ mytime   │
├────────────────────────────────────┼──────────┼──────────┤
│ 1970-01-01T00:00:00.001Z           │ a0       │ 1        │
├────────────────────────────────────┼──────────┼──────────┤
│ 1970-01-01T00:04:27.000Z           │ a1       │ 267000   │
├────────────────────────────────────┼──────────┼──────────┤
│ 1970-01-01T00:04:27.001Z           │ a2       │ 267001   │
├────────────────────────────────────┼──────────┼──────────┤
│ 1970-01-01T00:04:28.000Z           │ a3       │ 268000   │
[...]

PR #244, opened 2016-01-22
juttle-utils: parse numeric strings in toNative

davidvgalbraith:
Date returns odd results given numeric strings. Turn any numeric
strings in toNative into numbers, and multiply them by 1000 to view
them as seconds since Date expects milliseconds

Fixes https://github.com/juttle/juttle/issues/243

@juttle/developers


demmer:
That was easy! Looks good to me.

lib/runtime/juttle-utils.js
```diff
@@ -3,12 +3,21 @@ var ops = require('./ops');
 var values = require('./values');
 var JuttleMoment = require('../moment').JuttleMoment;
 
+function makeDate(time) {
+    var number = Number(time);
+    if (number === number) { // check for NaN
```
rlgomes: Oh the infamous check for NaN ;)

bkutil: `Number.isNaN` works equally well.

PR #246, opened 2016-01-22
Http server fixes

mnibecker:
Some follow up fixes/improvements from @davidvgalbraith comments in pr #224.

All these pertain to the `http_server` adapter. Here's the skinny
- Added support for `-rootPath` option
- Fix improperly scoped `format` variable
- Use `Promise.try` instead of a million try/catch blocks.
- Add link to `http_server` documentation, in adapter index


davidvgalbraith:
Huh, well I thought I +1ed this one before lunch but must have been hallucinating again. +1

PR #247, opened 2016-01-22
add all .js files to test lint checking

rlgomes:
we were missing lint checking on a few `.js` files in test that were
not `.spec.js` files so lets make sure to cover those.


rlgomes:
@juttle/developers quick and easy... we were not lining .js files in `/test`

mstemm:
lgtm.

PR #248, opened 2016-01-22
new adapter read and write api

### LINKED ISSUE #168: add source base class support for reading

demmer:
This PR is a big rework of the implementation of `read` and `write` and the adapter API.

Instead instead of having adapters directly implement procs, add normal proc implementations of read and write along with a new set of abstract base classes for adapters that have a cleaner contract that is more isolated from the details of the rest of the runtime.

Also convert the test adapters and the file adapter to use the new API and add a new test adapter that simulates a timeseries database to test the logic for reading pseudo-live.

More details are in the individual commit messages.

Not yet ready for merge but initial review would be appreciated.

To do before merge:
- [x] Convert the file adapter
- [x] Convert the http adapter
- [x] Convert the stdio adapter
- [x] Convert the stochastic adapter
- [x] Rework emit so that ticks are generated by the source proc
- [x] Implement one of the external adapters using this contract

Fixes #168.


demmer:
@juttle/developers

lib/adapters/http/read.js
```diff
@@ -1,13 +1,94 @@
...
+var Promise = require('bluebird');
+
+class Iterator {
+    constructor(adapter, req, res) {
+        this.points = [];
+        this.done = false;
+
+        var format;
+        try {
+            format = contentType.parse(res).type.split('/')[1];
```
davidvgalbraith: Know this was just a copy-paste but I'm confused by the logic...you take the content type of `res` and then use that to get a parser that you use on `req`? Seems like this should be `contentType.parse(req)`.

lib/adapters/http/read.js
```diff
@@ -1,13 +1,94 @@
...
+
+            // The current parser API doesn't offer any kind of backpressure,
+            // so all we can do is manage our own buffer of points and return
+            // them to read as fast as possible.
+            adapter.logger.debug('got', points.length, 'points');
+            this.points = this.points.concat(points);
+            this.notify();
+        })
+        .catch((err) => {
+            adapter.trigger('error', err);
```
davidvgalbraith: As we discussed on the file adapter, this should just throw.

lib/adapters/http/read.js
```diff
@@ -1,13 +1,94 @@
...
+var AdapterRead = require('../../runtime/adapter-read');
 var parsers = require('../parsers');
+var errors = require('../../errors');
 var request = require('request');
+var Promise = require('bluebird');
+
+class Iterator {
+    constructor(adapter, req, res) {
+        this.points = [];
+        this.done = false;
```
davidvgalbraith: This constructor does a ludicrous amount of work, let's cut it off here at line 13 and move the rest into a `start()` method.

lib/adapters/stochastic.js
```diff
@@ -27,11 +29,11 @@ var Read = Juttle.proc.source.extend({
...
         this.type = options.type;
         if (this.type && this.type !== 'metric' && this.type !== 'event') {
-            throw this.compile_error('RT-METRICS-OR-EVENTS');
+            throw new errors.compileError('RT-METRICS-OR-EVENTS');
         }
 
-        this.handleTimeOptions(options);
-        this.setDefaultTimeRange(true);
+        this.from = options.from || NOW;
+        this.to = options.to || NOW;
```
davidvgalbraith: Is there a use case for Stochastics that go from now to now? Seems like it'd be better to require `from`/`to`.

lib/adapters/http/read.js
demmer: I don't really see the point of splitting it into a separate `start` function since there's no case in which you would use an iterator without starting it immediately. Also FWIW I was burned for a while by the fact that if you don't start actually reading from the `res` object inside the `.on('response', (res) => {...}` handler, then you get no data at all.



However I do agree the constructor is long, so I split out the guts of the point processing into a separate function inside the iterator which tightens things up somewhat.

lib/adapters/http/read.js
demmer: Yeah I agree -- it turns out that either works, but it does seem confusing to parse the content type of the request and not the response.

I tightened things up so we don't use the `req` object at all any more.

lib/adapters/http/read.js
demmer: Yep. Done.

lib/adapters/stochastic.js
demmer: Actually `-from / -to` are required to be supplied, but weren't consistently being set in the options passed to the adapter as opposed to in the read proc.

Inspired by this problem, I just pushed a rework of how the adapters indicate whether or not they require time options (see fe75f52) so that it's a static property as opposed to an instance property, which lets the `read` proc make sure to always pass a valid from/to as part of the adapter constructor options.

demmer: After rebasing the relevant logic is now in 981efb1

demmer:
@davidvgalbraith I addressed all your comments, rebased, and then also made some additional changes. Everything from 0c7b30e onward is new.

Seems like we're close enough that it would make sense to merge this sooner rather than later and then start the process of converting the adapters. Once this merges I'd probably want to cut a new juttle release so we have something to base the adapter rework off of.

lib/runtime/procs/read.js
```diff
@@ -0,0 +1,282 @@
...
+            this.program.scheduler.schedule(this.program.scheduler.now(), _.bind(this.run, this));
+        }
+    },
+
+    teardown: function() {
+        if (! this.reading) {
+            this.emit_eof();
+        } else {
+            this.reading.then(() => {
+                this.emit_eof();
```
davidvgalbraith: This will emit 2 EOFs if the read hits an error due to line 237. I don't think that's a big deal, though.

lib/runtime/procs/source.js
```diff
@@ -74,46 +79,52 @@ var source = base.extend({
...
+    // Check if we're all done, and emit any needed ticks.
+    advance: function(endTime) {
+        if (endTime.eq(this.to)) {
+            this.logger.debug('advance: all done');
+            this.emit_eof();
+            return;
+        }
 
-        juttle_utils.toNative(points);
+        while (this.nextTick.lt(endTime)) {
```
davidvgalbraith: This tick business could use a comment.

lib/runtime/procs/read.js
demmer: Good point -- that doesn't seem like a good thing.
Even worse, I noticed that if there is a pending call into the scheduler that never gets cancelled or cleaned up.

test/adapters/stochastic.spec.js
```diff
@@ -25,14 +25,16 @@ describe('stochastic adapter options', function() {
         });
     });
 
-    it('supports starting at :beginning:', function() {
+    it('complains about starting at :beginning:', function() {
         return check_juttle({
             program: 'read stochastic -from :beginning: '+
                      '-to :1970-01-01T00:00:10.000Z: ' +
                      '| reduce count() | view text'
+        }).then(function(res) {
```
davidvgalbraith: There's this thing called `test_utils.expect_to_fail` that can simplify tests like this if you want.

davidvgalbraith:
+1

lib/runtime/juttle-utils.js
```diff
@@ -35,6 +36,26 @@ function toNative(records, epsilon) {
...
+            context.trigger('warning', new errors.runtimeError('RT-POINT-MISSING-TIME', {field: timeField}));
+        }
+
+        var time = pt[timeField || 'time'];
+        if (time) { pt.time = time; }
+    });
+
+    toNative(points);
+
+    return points;
```
davidvgalbraith: This can just `return toNative(points)`.

lib/runtime/procs/read.js
```diff
@@ -0,0 +1,282 @@
...
+            //
+            // Notify the base class so it can emit any necessary ticks and
+            // check if we're all done, or call us back when it is time to
+            // reschedule.
+            //
+            // Otherwise we need to reschedule another call.
+            if (result.readEnd) {
+                this.advance(result.readEnd);
+            } else {
+                this.moreToRead = true;
```
davidvgalbraith: You don't set `moreToRead` false at the end of the read, how does the ending logic work?

davidvgalbraith: Oh ho, it's in `advance`. I wonder if it would be possible to have `emit_eof` in fewer places here, little hard to keep track of them all.

lib/runtime/adapter-write.js
```diff
@@ -0,0 +1,48 @@
...
+    write() {
+        throw new Error('write must be implemented');
+    }
+
+    // Signal that the stream is done.
+    //
+    // Returns a promise once the adapter has completed writing all the data and
+    // the flowgraph can be shut down.
+    eof() {
+        throw new Error('write must be implemented');
```
davidvgalbraith: Should be "eof must be implemented".

lib/adapters/stdio/write.js
```diff
@@ -21,26 +22,24 @@ var Write = Juttle.proc.sink.extend({
...
+    }
 
-    process: function(points) {
+    write(points) {
         this.serializer.write(points);
-    },
+    }
 
-    eof: function() {
+    eof() {
```
davidvgalbraith: The contract says we're supposed to return a promise here.

demmer: The base class uses `Promise.try` so I didn't see any need here.

davidvgalbraith: Well then we should change the contract. Are we synchronously done here after calling `this.serializer.done()` anyway?

demmer: Well that seems like a good thing to do.

lib/runtime/adapter-read.js
```diff
@@ -0,0 +1,59 @@
...
+    // Hook to indicate that the read has been started
+    start() {
+    }
+
+    // Core API to read up to `limit` points between the specified time
+    // intervals.
+    //
+    // Returns a promise that resolves with an object containing:
+    //   points: the points that are ready to emit in the flowgraph
+    //   done: true if all points between [from..to) were produced
```
mstemm: I believe this comment is out of date, and it should be readEnd with values null or a JuttleMoment with the updated maximum to value for whch points have been read?

lib/runtime/procs/source.js
```diff
@@ -16,12 +15,19 @@ var INFO = {
     options: {
         from: {},
         to: {},
-        last: {}
+        last: {},
+        queueSize: {}
```
VladVega: Should queueSize be instantiated here? And if so as an object?

lib/runtime/procs/read.js
```diff
@@ -0,0 +1,270 @@
...
+
+            if (result.points && result.points.length !== 0) {
+                this.logger.debug('read returned', result.points.length, 'points',
+                                  'readEnd', result.readEnd ? result.readEnd.toJSON() : 'null');
+                this.enqueue(result.points);
+            } else {
+                this.logger.debug('read returned no points');
+            }
+
+            this.readState = result.state;
```
VladVega: It's unclear to me what this could be used for.

VladVega: Actually, nm saw the usage in this PR.

lib/runtime/procs/source.js
demmer: This is the expected format for the (still unused) autocompleter.
Until we get that teed up again I'd just consider this a no-op.

demmer:
@juttle/developers I took another pass at changing the API to better support adapters that wabt to determine their behavior relating to the time options at runtime instead of statically. I also rebased to reconcile all the latest changes to the various builtin adapters (thanks rodney) and addressed the various comments in the PR history.

I think this is ready to merge so we can get on with porting the actual adapters.

lib/adapters/file/read.js
```diff
@@ -75,18 +83,19 @@ var Read = Juttle.proc.source.extend({
                     validPoints.push(point);
                 }
             });
-            self.emit(validPoints);
         })
-        .catch(function(err) {
-            self.trigger('error', self.runtime_error('RT-INTERNAL-ERROR', {
+        .catch((err) => {
```
davidvgalbraith: This `catch` can go after the following `then`.

lib/runtime/adapter-write.js
```diff
@@ -0,0 +1,53 @@
...
+    write() {
+        throw new Error('write must be implemented');
+    }
+
+    // Signal that the stream is done.
+    //
+    // Returns a value or a promise once the adapter has completed writing all
+    // the data and the flowgraph can be shut down.
+    eof() {
+        throw new Error('write must be implemented');
```
davidvgalbraith: Still should be `eof must be implemented`.

lib/adapters/file/read.js
demmer: Changed

lib/runtime/adapter-write.js
demmer: Oops. Thanks.

davidvgalbraith:
+1

demmer:
Dave said "ship it red!"

rlgomes:
I'll look at that file adapter test failure...

PR #249, opened 2016-01-22
enable strict mode

### LINKED ISSUE #240: use strict mode to enable ES6 classes
Some ES6 features (class, const, let, etc) require strict mode in the current version of node.js.

Since strict mode is generally useful anyway, we should enable it globally to both make it easier to start using these features, but also to avoid the requirement for each module to have the boilerplate "use strict" at the top of each file.

A trivial way to do this in both the juttle CLI and the outrigger components would be to just use something like:
https://www.npmjs.com/package/use-strict

rlgomes:
fixes #240

flip the switch to enforce `'use strict';` with eslint


rlgomes:
@juttle/developers `'use strict';` everywhere!

demmer:
Oh so strict... +1 on green.

One question -- did eslint --fix do this for you?

rlgomes:
@demmer eslint wouldn't add the line but [use-strict-cli](https://www.npmjs.com/package/use-strict-cli) did the trick!

demmer:
Good to know. Thanks.


On Fri, Jan 22, 2016 at 1:36 PM, Rodney Lopes Gomes <
notifications@github.com> wrote:

> @demmer <https://github.com/demmer> eslint wouldn't add the line but
> use-strict-cli <https://www.npmjs.com/package/use-strict-cli> did the
> trick!
>
> —
> <https://github.com/juttle/juttle/pull/249#issuecomment-174057421>.
>



-- 
Michael Demmer
Jut - CTO / VP Engineering
demmer@jut.io

rlgomes:
merging this on the +1 Demmer gave me and just a small tweak for that scoping issue in the graphi-compiler.js file.

PR #251, opened 2016-01-23
remove unwanted tabs

rlgomes:


rlgomes:
@juttle/developers I still need to figure out how to solve this for us as we don't want those nasty tabs getting in there and eslint has an open bug on this matter.

rlgomes:
merging as these tabs were introduced by me and should have never gone in.

PR #252, opened 2016-01-23
Update README.md with OpenTSDB adapter link

VladVega:
@mstemm @demmer


mstemm:
lgtm. I'm gonna assume that build failure isn't you, it looks like the master build is also broken.

PR #253, opened 2016-01-23
http_server: use a free port for each test

demmer:
Stop using hard-coded ports in the http server test since they're causing travis flakiness, and instead grab a free port for each test case.

Note that I tried grabbing a single free port for the whole test suite, but it seemed like there were still races in which one test case wouldn't release the port before the next one started, so I just used a simple beforeEach() to grab a new one each time.

### DIFF
--- a/lib/adapters/http_server.js
+++ b/lib/adapters/http_server.js
@@ -117,11 +117,11 @@ var Read = Juttle.proc.source.extend({
     },
 
     teardown: function() {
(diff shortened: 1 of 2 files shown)


demmer:
@mnibecker if this passes travis you can close #250

demmer:
bump @mnibecker @rlgomes @davidvgalbraith

rlgomes:
+1 looks good

PR #255, opened 2016-01-25
errors: Remove “Error:” and “Warning:” prefixes from messages

dmajda:
Instead of having “Error:” and “Warning:” prefixes embedded in error and warning messages, let’s add them dynamically when displaying the each error or warning.

There are 3 reasons for this change:

  1. The prefixes were duplicated in almost all messages, cluttering them.

  2. The prefixes were not used consistently (some messages were prefix-less), so some unification was needed in any case.

  3. There were many situations where a message labeled as error ended up being used in a warning, creating confusion. This typically happened when a runtime error occurred and was converted into a warning. For example, this program:

         bin/juttle -e 'emit -limit 3 | put a = null, b = 1 + a

     produced these 3 warnings before the change:

         Error: Invalid operand types for "+": number (1) and null.
         Error: Invalid operand types for "+": number (1) and null.
         Error: Invalid operand types for "+": number (1) and null.

     After the change it produces:

         Warning: Invalid operand types for "+": number (1) and null.
         Warning: Invalid operand types for "+": number (1) and null.
         Warning: Invalid operand types for "+": number (1) and null.

Part of the work on #165.


dmajda:
@mstemm @demmer

demmer:
Much cleaner. +1

PR #257, opened 2016-01-25
better json serializer appending

rlgomes:
By using the `jsoner` package we can append to JSON objects in a file
without having to load the whole file. This removes the need for those
various options to handle maximum number of points and maximum filesize.

### DIFF
--- a/docs/adapters/file.md
+++ b/docs/adapters/file.md
@@ -85,8 +85,7 @@ Parameter         |             Description          | Required?
 ----------------- | -------------------------------- | ---------:
 `-file`           | File path on the local filesystem, absolute or relative to the current working directory  | Yes
 `-bufferLimit`    | Maximum number of points that will be written to the file. | No; defaults to 100
-`-maxFilesize`    | Maximum size of the file being written to, limited since the entire JSON needs to fit in memory | No; defaults to 200MB
-`-limit` | Maximum number of points in a file before we'd stop writing points to the file to avoid running out of memory | No; defaults to 100000
+`-append`         | When set to true your JSON data is append to an existing file otherwise the file is truncated and data from the current program is written out | No; defaults to false
 
 If the file already exists and contains a valid JSON array, the write will append new data rather than overwrite.
 
--- a/lib/adapters/file/write.js
+++ b/lib/adapters/file/write.js
@@ -1,14 +1,14 @@
 'use strict';
 
 var _ = require('underscore');
+var fs = require('fs');
 var Juttle = require('../../runtime/index').Juttle;
-var Promise = require('bluebird');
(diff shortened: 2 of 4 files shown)


rlgomes:
@davidvgalbraith @demmer wanted to add the `-format` to file adapter but before I can do that I think we should clean the file adapter to do "better" writing of JSON format that doesn't include having to read the whole file into memory in order to append. I used a package I developed on my own free time and then tweaked it a bit just now to make it more friendly for our use [jsoner](https://www.npmjs.com/package/jsoner). Let me know what you think but this removes limitations of writing out large JSON files.

With this change I can then follow up with adding the `-format` option so we can write jsonl and csv and also expose an `-append` option to dictate if you want to append to the existing file or simply truncate the file which. Since JSON would have this existing append mechanism and if you were not appending we could rely on the existing `serializer/json.js` to take care of business.

@dmehra with this you wouldn't need to output the data in jsonl and could just continue to use JSON.

demmer:
@rlgomes I was also thinking that it would make more sense to have an `-append` option that would default to false since I suspect that many cases of using the `file` adapter would actually be more inclined to overwrite as opposed to append.

As far as `jsoner` goes, I took a quick look and the biggest hesitation I would have is that you seem to only expose a blocking API and we definitely want to keep all our adapter I/O asynchronous.

rlgomes:
@demmer I was thinking about exposing a non blocking ;) just took the "easy" route when writing this to see if would even work.

dmehra:
+1 to append being not the default behavior, it has surprised me a few times when i didn't realize i already had an output file with the same name, and ended up with huge concatenations. It's great to have the option to append, but the default be replace.

rlgomes:
@davidvgalbraith @demmer updated and ready for your review

rlgomes:
bump!

davidvgalbraith:
jsoner.append still has calls to statSync and writeFileSync and openSync and closeSync, can we async the whole lot of it?

rlgomes:
You are correct I will fix that all up..

rlgomes:
@davidvgalbraith fixed up the stray synchronous fs calls.

rlgomes:
@davidvgalbraith @demmer so how about I make a PR that basically removes the appending behavior when doing `write file` and simply exposes `-format` with the default of `json` which truncates and uses the available serializers to write out data (just like `write stdio` does). Then we can add the -append option for formats that support this easily such as `jsonl` and possibly 'csv'. ? Then I can simply not do any of this magic with the `jsoner` library and we can add that back if someone REALLY wants to do appending to a JSON file.

rlgomes:
moved to simply rip out the whole append behaviour: https://github.com/juttle/juttle/pull/280

PR #259, opened 2016-01-26
juttle-utils#toNative: Convert JavaScript Date to JuttleMoment

go-oleg:
So adapters can successfully convert points with time fields of type `Date` into points with `JuttleMoment`  values (for emitting into the flowgraph).

This does create some asymmetry between `toNative` and `fromNative` (which returns an ISO string for `JuttleMoment`). But `toNative` does already accept ISO strings _or_ number of seconds so being liberal in what we accept on the `toNative` side makes sense.

@davidvgalbraith @demmer

### DIFF
--- a/lib/runtime/juttle-utils.js
+++ b/lib/runtime/juttle-utils.js
@@ -6,6 +6,8 @@ var values = require('./values');
 var JuttleMoment = require('../moment').JuttleMoment;
 
 function makeDate(time) {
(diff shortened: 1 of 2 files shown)


davidvgalbraith:
I think it'd be neater to just do `if (time instanceof date) { return time; }` in `makeDate`.

go-oleg:
Done.

davidvgalbraith:
+1

PR #260, opened 2016-01-26
fix memory leak in JSON parser

### LINKED ISSUE #258: Reading a large file is slow and results in OOM
I was running this program on an input file that's 442MB in size, and noticing that it takes a very long time (9 min) just to get to the desired `from` location in the file (which is 3/4 of the way through). Shortly after getting there and starting to process points, the program runs out of memory and quits.

```
read file -file '/tmp/kbs_metrics.json'  -from :2014-02-23T16:45:27: 
| put num = Number.fromString(String.slice(host,4,String.search(host,/\./))) 
| put skips = num%10 
| filter skips = 0 
| remove num, skips 
| write file -file '/tmp/mini_new_kbs_metrics.json' -limit 3000000
```

Even simpler the OOM can be reached with just this

```
read file -file '/tmp/kbs_metrics.json' 
| filter banana=true"
```

Github won't let me attach the data set (zipped to 30MB) but @rlgomes has it. Here is a snippet:

```
[
    {
        "name": "app.ep.client_http.kbytes_out",
        "host": "host146.samwell.jutt.le",
        "time": "2014-02-22T21:52:35.000Z",
        "kb_out": 13110.5087890625
    },
    {
        "name": "app.ep.client_http.kbytes_out",
        "host": "host605.hodor.jutt.le",
        "time": "2014-02-22T21:52:35.000Z",
        "kb_out": 1366.416015625
    },

```

rlgomes:
fixes #258

we were not correctly returning `oboe.drop` while using oboe to steam
JSON objects and therefore oboe was holding onto the references to each
of the parsed JSON objects and eventually blowing up as identified in
the issue #258

### DIFF
--- a/lib/adapters/parsers/json.js
+++ b/lib/adapters/parsers/json.js
@@ -42,6 +42,8 @@ module.exports = base.extend({
                     emit(buffer);
                     buffer = [];
                 }
+                
+                return oboe.drop;
             })
             .on('end', function(){
                 emit(buffer);


rlgomes:
@dmehra this fixes the OOM but does not fix any of the slowness issues. I think I can fix the slowness by moving to use the `jsoner` package I'm introducing in another PR but will have to add the `rootPath` feature to `jsoner` before I can attempt to verify it is in fact faster. At least with this PR we don't OOM any longer.

@demmer @davidvgalbraith pretty easy review.

davidvgalbraith:
Nice! +1

rlgomes:
@dmehra give this a shot withyour json file locally to confirm we at least don't OOM.

dmehra:
Yes the OOM is gone; also processing is faster, to the tune of 2 min where I was seeing 9 min previously. Still not lightning fast, but improved.

Issue #57, opened 2015-12-26
Generalize filter expressions


dmehra:
Remove filter expression checks, which enables more expressive filter expressions in filter (ones which include field-to-field comparisons, function execution on fields, etc.).

Implementation plan (see [below](https://github.com/juttle/juttle/issues/57#issuecomment-174889819) for details):

  * [x] Move all structural filter checks from the semantic pass into `FilterChecker`
  * [x] Extend `FilterJSCompiler` to handle dynamic filters
  * [x] Split `FilterChecker` into `StaticFilterChecker` and `DynamicFilterChecker`
  * [x] Switch `filter` and `sequence` to dynamic filters

dmajda:
So the current idea is that we will have two kinds of filters in Juttle:

  * **Static filters.** These will be similar to filters we have now. They will allow only comparison of fields against each other and against compile-time values. Optionally they will also allow FTS. The intent is to translate them into backend queries.

  * **Dynamic filters.** These will be the “generalized filters” as originally envisioned. They will allow all kinds of streaming expressions like computing functions over fields, the `*` unary operator, etc. They won’t allow FTS. The intent is to translate them into JavaScript code.

Procs like `filter` will allow dynamic filters. Adapters will be able to chose which kind of filter they allow (note this means the kind of a filter used in `read` won’t be known statically). Typically, big data adapters will allow only static filters (translating them into backend queries) while small data adapters will allow dynamic filters (translating them into JavaScript which they will run on incoming points).

---

Now the implementation plan:

  1. **Move all structural filter checks from the semantic pass into `FilterChecker`.** This means all filters will pass through the semantic pass like any other expression.
  2. **Extend `FilterJSCompiler` to handle dynamic filters.** This mostly means adding support for compiling streaming expressions.
  3. **Split `FilterChecker` into `StaticFilterChecker` and `DynamicFilterChecker`.** The former will do the same thing as `FilterChecker` today, the latter will probably just check for FTS.
  4. **Switch `filter` and `sequence` to dynamic filters.** The adapters will still use static filters.

The next step (subject of some follow-up issue) is allowing adapters to opt-in into dynamic filters and generally make their dealing with filters easier (I’m thinking about having some declarative way to specify what is allowed in each adapter’s filter). There is an API to be discussed and designed here.

PR #262, opened 2016-01-26
filters: Don’t semantically check FilterLiterals in FilterChecker

dmajda:
Inputs of type `filter` produce `FilterLiteral` AST nodes that need to be semantically checked before they are compiled as part of a filter. The check is needed mainly to convert any `Variable` nodes into field
references (`UnaryExpression` nodes with `*` as the operator).

Until now, these checks were done in `FilterChecker` after `FilterLiteral` nodes coming from input values were embedded into filter expressions used by procs in the build phase. This was not ideal for two reasons:

  1. `FilterChecker` is not supposed to modify checked filter AST, but applying a semantic check to `FilterLiteral`s did modify it.

  2. We want to check the `FilterLiteral` itself, not its (possible many) embeddings.

This commit addresses the problems by moving the semantic check to `GraphBuilder` and applying it to ASTs of all input values for generality. This seems a little bit cleaner.

Part of the work on #57.

### DIFF
--- a/lib/compiler/graph-builder.js
+++ b/lib/compiler/graph-builder.js
@@ -7,6 +7,7 @@ var _ = require('underscore');
 var values = require('../runtime/values');
 var Juttle = require('../runtime/procs/procs');
 var errors = require('../errors');
+var SemanticPass = require('./semantic');
 
 function proc_name(node) {
     return (node.name || node.type).replace('Proc', '').toLowerCase();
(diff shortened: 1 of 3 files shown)


dmajda:
@bkutil

lib/compiler/graph-builder.js
```diff
@@ -304,6 +307,10 @@ var GraphBuilder = Base.extend({
         }
 
         return null;
+    },
+    _semantic_ast: function(ast) {
+        var semantic = new SemanticPass({ now: this.now });
```
bkutil: Creating the instance here and repeatedly for each value seems a bit odd, but it is more 'self contained'. What do you think?

bkutil:
+1

lib/compiler/graph-builder.js
dmajda: This is a bit of defensive move as there are no guarantees about the state of the `SemanticPass` instance after one runs `sa_expr` which really wasn’t intended as a public API. I didn’t want to attack that problem now as I hope to solve more generally when I’ll make `SemanticPass` a visitor.

PR #264, opened 2016-01-26
filters: Rewrite FilterChecker in more declarative style

dmajda:
`FilterChecker` was a bit of a spaghetti code which was hard to orient in. This commit rewrites it in more declarative style and makes it driven by a table of allowed forms for each operator. Beside making the code easier to follow this solution is more flexible and extensible (for example, it would be relatively easy to make the checks customizable for adapters).

The change doesn’t change the semantics of filter expressions in any way, but it changes some error messages (to the better, I hope).

Part of the work on #57.

### DIFF
--- a/lib/compiler/filters/filter-checker.js
+++ b/lib/compiler/filters/filter-checker.js
@@ -13,6 +13,7 @@
 // already (the idea is to do them as soon as possible in the compiler
 // pipeline).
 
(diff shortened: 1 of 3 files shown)


dmajda:
@bkutil

bkutil:
Looks good. +1

PR #265, opened 2016-01-26
filters: Remove redundant structural filter checks in the semantic pass

dmajda:
After [rewriting `FilterChecker` in declarative style](https://github.com/juttle/juttle/pull/264), structural checks in the semantic pass became redundant as `FilterChecker` now performs them too. This PR removes them.

The change doesn’t change the semantics of filter expressions in any way, but it changes some error messages (to the better, I hope).

Note that I had to add

    ast.expression.d = false;

in `sa_ExpressionFilterTerm` to prevent replacing build-time expressions such as `1 < 2` by their results before `FilterChecker` has a chance to inspect them.

Part of the work on #57.


dmajda:
@bkutil

bkutil:
+1

PR #266, opened 2016-01-26
WIP scheduler simulation

bkutil:
A simulation of a data store, adapter (`fetch` and `drain`) and a scheduler.

The idea is that the scheduler tries to emit all points older than the 'least recent start of a window for all adapters' and also that adapters have a cap on the data they can store in the buffer.

*NOT FOR MERGE*

### DIFF
--- a/extlib/scheduler-playground/lib/scheduler.js
+++ b/extlib/scheduler-playground/lib/scheduler.js
@@ -1,0 +1,156 @@
+var _ = require('underscore');
+
+var Scheduler = function(options) {
+    // Scheduling information
+    this.info = {};
+
+    // Adapter id dispenser
+    this.nextId = 0;
+};
+
+Scheduler.prototype.addAdapter = function(adapter, from, to) {
+    var self = this;
+
+    // Meh
+    var id = this.nextId;
+
+    // Create scheduling record, the time window. Originally <from, to>, adapters update
+    // the <from> as they fetch data. When the <from> reaches <to>, we're done - which can
+    // be immediately for historical, never for :end: or in the future for superqueries.
+    this.info[id] = { adapter: adapter, from: new Date(from.getTime()), to: new Date(to.getTime()) };
+
+    this.nextId++;
+
+    function log(name, e) {
+        console.log(id
+                + ' ('
+                + e.head.time.getTime()
+                + ', '
+                + e.tail.time.getTime()
+                + '), '
+                + name
+                + ' points: '
+                + e.size
+                + ' qlength: '
+                + e.qlength
+                );
+    }
+
+    // When adapter fetches data in the interval between <from> and <to>,
+    // there are two cases to handle:
+    //
+    // 1. fetch did not return any points (disregarding errors for now) or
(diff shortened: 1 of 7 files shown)


bkutil:
cc: @demmer @davidvgalbraith

extlib/scheduler-playground/lib/scheduler.js
```diff
@@ -0,0 +1,156 @@
...
+                + e.head.time.getTime()
+                + ', '
+                + e.tail.time.getTime()
+                + '), '
+                + name
+                + ' points: '
+                + e.size
+                + ' qlength: '
+                + e.qlength
+                );
```
VladVega: Consider trying out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings here.

extlib/scheduler-playground/lib/store.js
```diff
@@ -0,0 +1,38 @@
...
+    this.data = [];
+
+    this.init();
+};
+
+Store.prototype.init = function() {
+    var range = this.last.getTime() - this.first.getTime();
+    var maxPoints = (range / 1000) * this.density;
+
+    while (this.data.length < maxPoints) {
```
VladVega: Density is great parameter to be able to pass. Question: what happens when the maxPoints threshold is reached with all points representing of the first quarter of the time range requested?

extlib/scheduler-playground/lib/scheduler.js
```diff
@@ -0,0 +1,156 @@
...
+    var nextFetch = this.getNextFetch();
+
+    // All busy or finished
+    if (!nextFetch) { return; }
+
+    var cur = new Date(Date.now());
+
+    // If <to> is in the future, fetch until now
+    // This is where the window logic would go
+    if (nextFetch.to.getTime() > cur.getTime()) {
```
VladVega: Question about historical and super query:
Will there be a need to break the scheduled requests into windows for the historical parts of the query? Or will it always be `from` until `now > to ? to : now`? And then scheduled time windows for live.

extlib/scheduler-playground/lib/scheduler.js
bkutil: This is cool, thanks! :)

extlib/scheduler-playground/lib/store.js
bkutil: What happens now is, that the number of points will always be `maxPoints` in the time interval - the average will equal to `density`, but the times are generated randomly (from a uniform distribution between `first` and `last`. That they would all fall into a sub-interval is possible, but won't happen on the average.

In that case, the adapter would just crunch the points in the first quarter as fast as it's queue size allows, then process the remaining few (or zero) points by querying with `from/to = (1/4, 1)`

Does that answer your question? I think that the comment above the `density` option is misleading and I should probably change that.

Also, for debugging purposes, to be able to specify a more regular sequence (e.g. tick every nth of a second) would be useful.

dmajda:
Closing, the real implementation will be done in #167.

PR #267, opened 2016-01-26
adding head optimization for `read file`

### LINKED ISSUE #261: `read file` could use head/tail optimization
Especially `head` as something strange is going on there. When i run this program:

```
`read file -file '/tmp/bigfile.json' | head 1
```

(my file happens to be 42MB), then I immediately see the first point output on the CLI, but after that the program sits there for 6-7 seconds longer, doing mysterious activitiy. Probably reading through the rest of the file just to throw it all away.

When I run `read file | tail 1` the pause at least makes sense, since we read through the whole file to get to the last point; with `head 1` it's puzzling to get the output and then still have the same pause.

Ideally we would optimize both `head` and `tail` so the operations aren't slower than in the unix shell (currently, they are).

rlgomes:
fixes #261

wanted to first get some feedback on the approach since I'm doing the
optimization in the parser code so that all adapters can take advantage
of the ability.

with this first pass made "| head n" optimized for `read file` and
there really is no optimization for "| tail n" since you can't possibly
guess how far in the stream to advance to get to the `nth` last points
for any format.

missing tests to verify the parsers directly and then I'll reorganize
the whole PR so that in a follow up PR I can easily add the same `head`
optimization for stdio and http.


rlgomes:
@davidvgalbraith @demmer checkout them optimizations...

@dmehra here's the work but we really can only optimize `head` at this point since you have no clue when the stream ends or how far to advance to get to the tail.

dmehra:
For `tail`, it depends on what we are already doing when reading the file. If we parse the timestamp for each line when reading, we could skip doing that for `tail` because we only need to parse time for the last N lines. If it's already done that way, then yeah no more optimization.

lib/adapters/parsers/csv.js
```diff
@@ -22,10 +28,19 @@ module.exports = base.extend({
             });
 
             csvStream.on('data', function(pt){
-                buffer.push(pt);
-                if (buffer.length >= self.limit) {
-                    emit(buffer);
-                    buffer = [];
+                self.totalParsed++;
+
+                if (self.stopAt !== -1 && self.totalParsed > self.stopAt) {
```
davidvgalbraith: Let's have the unoptimized `stopAt` be `Infinity` so we don't need the extra `!== -1` check.

lib/adapters/parsers/csv.js
```diff
@@ -22,10 +28,19 @@ module.exports = base.extend({
...
+                    if (!self.doneParsing) { 
+                        self.doneParsing = true;
+                        this.emit('end');
```
davidvgalbraith: Should this be `self.emit`?

lib/adapters/parsers/csv.js
```diff
@@ -9,6 +9,12 @@ module.exports = base.extend({
 
     initialize: function(options) {
         this.separator = options.separator ? options.separator : ',';
+        this.totalParsed = 0;
+        this.stopAt = -1;
+
+        if (options.optimization && options.optimization.type === 'head') {
```
davidvgalbraith: Can you do this in the base parser class?

davidvgalbraith:
Is there no way to stop parsing altogether once we get to `stopAt`? Looks like you keep reading the rest of the file and just drop it on the floor.

lib/adapters/parsers/csv.js
rlgomes: will do.

lib/adapters/parsers/csv.js
rlgomes: `this.emit` is on the csvStream not the parser class which I'll investigate why exactly this doesn't stop everything instead of continuing to parse the inbound stream.

lib/adapters/parsers/csv.js
rlgomes: Will do.

rlgomes:
@davidvgalbraith will see about finding a way to stop reading the stream at `stopAt` not all of the underlying parsers have something like what `oboejs` has to `this.abort` the run.

rlgomes:
@davidvgalbraith fixed up all the things you mentioned and added tests for each of the individual parsers. The parser tests also verify exactly how many lines are "parsed" - actually fed ahead into the flowgraph vs "read" - were read by the parser due to the inability to truly stop on a dime when reading certain types of data. So json and jsonl work well and stop on a dime when we hit the limit imposed by `head` but grok and csv have read ahead buffers and emit those points they've already read out of the stream even after being told to end/close/etc. We could open issues to find better parsers for those two but I think its not a horrible thing to read just skip over those few points.

davidvgalbraith:
+1

PR #270, opened 2016-01-27
Implement Juttle value ↔ AST conversions

dmajda:
Implement `values.toAST` and `values.fromAST` functions which convert Juttle values from/to their AST representations. These functions replace ad-hoc, ugly, bug-ridden implementations of this functionality that we already had in the codebase.

The original motivation for this was getting rid of two lists of type display names ([in `values.js`](https://github.com/juttle/juttle/blob/612a8713a204ecb58255e1ed876e445613c86369/lib/runtime/values.js#L10-L21) and [in `filter-checker.js`](https://github.com/juttle/juttle/blob/master/lib/compiler/filters/filter-checker.js#L20-L33)). This will actually be done in a follow-up PR.

Vaguely related to #57.

### DIFF
--- a/test/runtime/values.spec.js
+++ b/test/runtime/values.spec.js
@@ -573,6 +573,17 @@ describe('Values tests', function () {
         },
     ];
 
+    describe('fromAST', function() {
+        it('returns correct value', function() {
+            _.each(AST_TESTCASES, function(testcase) {
+                expect(values.fromAST(testcase.ast)).to.deep.equal(
+                    testcase.value,
+                    JSON.stringify(testcase.ast)
+                );
+            });
+        });
+    });
+
     describe('toAST', function() {
         it('returns correct AST', function() {
             _.each(AST_TESTCASES, function(testcase) {
--- a/test/runtime/values.spec.js
+++ b/test/runtime/values.spec.js
@@ -483,4 +483,104 @@ describe('Values tests', function () {
             });
         });
     });
+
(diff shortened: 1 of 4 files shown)


dmajda:
@bkutil

test/runtime/values.spec.js
```diff
@@ -7,6 +7,105 @@ var JuttleMoment = require('../../lib/moment').JuttleMoment;
...
+        type: 'ExpressionFilterTerm',
+        expression: {
+            type: 'BinaryExpression',
+            operator: '<',
+            left: { type: 'Variable', name: 'a' },
+            right: { type: 'NumericLiteral', value: 5 }
+        }
+    };
+
+    var AST_TESTCASES = [
```
bkutil: This is probably a matter of taste, but I think its usually better to keep the test cases close to the actual tests. If you prefer constants, would you mind moving the other tests, too, please?

bkutil:
Overall, looks good (+1). One thing - the `values.js` is getting a bit long - not sure if it is worth splitting, though. What do you think?

test/runtime/values.spec.js
dmajda: I’ll just move the variable right above the tests (I was split about its placement anyway).

dmajda:
Updated, will merge on green.

I agree that `values.js`is becoming long. I think we should split it the next time we add a new non-trivial functionality there.

PR #272, opened 2016-01-27
Remove dropped proc

bkutil:
As a part of cleaning up publish/subscribe procs, remove dropped.

This was used to catch (at a distance) dropped points and re-emit them
into the flowgraph. However, currently no other proc uses it.


bkutil:
cc: @dmajda @demmer

dmajda:
+1 technically

@demmer Not sure what our plans are for resurrecting `dropped` or implementing some similar mechanism. But unless we are sure we want `dropped` revived and want to implement feeding points into it from sources now (meaning next month or so), I’d go ahead with dropping.

dmehra:
When we supported live in 1.0, the dropped proc was a necessary debugging tool. Now that we are bringing back live (pseudo live?) but proposing to rip out `dropped` I'd like to understand why the need for similar debugging proc won't be there. I don't have full understanding of the new scheduler and live mode just yet - perhaps it eliminates the problem altogether or calls for different ways of debugging? Please explain.

bkutil:
Oh, OK. I wasn't aware about this use case, as at the moment, none of the adapters/procs use pub/sub to report dropped points.

dmajda:
@dmehra Let me clarify. The `dropped` proc currently isn’t hooked in, it is effectively dead code. We can either (a) hook it in, (b) keep it lying around unhooked, or (a) delete its code (without any harm to Juttle-as-is).

My position is: If we are sure we want to have `dropped` in the language (and not implement some other debugging mechanism instead), let’s hook it in. If we are not sure, or do not want to do the hooking work soonish, let’s delete the code and at some later point either restore it or implement some other debugging mechanism. What I’d like to avoid is the status quo where the the code is just lying around unhooked and prevents @bkutil’s refactoring.

This is not a discussion about whether we need a debugging mechanism (we do), but about what to do with a bit of currently unused piece of code.

demmer:
@dmehra I second what @dmajda said -- debuggability is still going to be an important thing for us to think about and `dropped` was certainly a step in the right direction on that front. However the existing implementation is really only useful for "true live" sources in which the timestamps might have clock skew. Currently only twitter, http_server, and stdio fall into that category. All of the pseudo-live adapters don't need `dropped` in the same way it worked before because time will be sorted by the back end as part of each query.

There are, however, a bunch of other places in the flowgraph and reasons why points might get dropped -- perhaps someone reads from a source that doesn't have time in-order and then sends the points through `reduce` or `join`, perhaps the scheduler backs up and can't keep up with the flow of points without overwhelming a buffer, perhaps we hit big data too big, etc. All these seem like more useful and relevant places to get debugging information than the old dropped proc.

Also the point of this PR is that the publish / subscribe mechanism used by dropped is one that we want to remove and that any future debugging should use a different mechanism.

So all that said, my vote would be to delete the code and then add a new implementation when we figure out a real plan for debuggability. +1 to merge from my standpoint

bkutil:
Ok, thanks for everyones input on this. I created an issue to design/discuss `dropped` replacement later: https://github.com/juttle/juttle/issues/284. Will merge.

PR #273, opened 2016-01-27
ASTVisitor: Move FilterLiteral to other literals

dmajda:
The new place makes more sense and better corresponds to ordering used e.g. in `values.js` or `FilterJSCompiler`.


dmajda:
@bkutil Trivial one.

bkutil:
+1

Issue #77, opened 2015-12-29
Should "emit -points" stamp points with time if none is present?


go-oleg:
Currently if you generate some timeless points with `emit -points`, time is added to them.

juttle:
```raw
emit -points [ { a: 1 }, { a: 2 } ]
```

output:
```raw
┌────────────────────────────────────┬──────────┐
│ time                               │ a        │
├────────────────────────────────────┼──────────┤
│ 2015-12-29T03:52:42.593Z           │ 1        │
├────────────────────────────────────┼──────────┤
│ 2015-12-29T03:52:43.593Z           │ 2        │
└────────────────────────────────────┴──────────┘
```

I don't think we want to be mutating points the user specified.

Thoughts?

dmehra:
@dmajda please weigh in on this. The behavior of adding timestamps to -points array was surprising to me when I was playing around with `join` like a new user might. When I thought about it, it made sense since `emit` always generates points with time... and yet, the initial reaction was consternation. 

If we want to keep the behavior, let's make the docs very clear. The current emit doc says 
```
-points	An array of points to stream
```
with no mention of adding time to them.

demmer:
It seems to me the default should be to emit timeless points, but if the user specifies a `-every` option, then emit should timestamp accordinvgly.

go-oleg:
`-every` with `-points` could also mean emit all of these points every `-every` seconds instead of one of these points every `-every` seconds serially. Not sure which is more common/intuitive, just bringing it up.

dmajda:
I agree the behavior is unintuitive. I think its rationale was that `-points` was mostly used in tests (it replaced old `demo ticker` there) where the auto-add behavior helped with creating meaningful point sequences.

I tried to comment-out the relevant code and only 8 test failures popped-up, 1 of which was in `emit` tests. So we don’t depend on the behavior much and I’d vote for changing it. As long as we support timeless points (which I’m not a fan of at all, as some may remember from old discussions), let’s just emit what is in `-points` verbatim (but still disallow a mix of timestamped and non-timestamped points).

> It seems to me the default should be to emit timeless points, but if the user specifies a -every option, then emit should timestamp accordinvgly.

This is little bit too magical. I’d say `-every` and `-points` should be mutually exclusive. And if we really have a strong use case for time auto-add, let’s invent a dedicated option to enable that.

demmer:
@dmajda That sounds good to me.

dmajda:
OK, morphing into a task.

[...]

PR #274, opened 2016-01-27
Emit -points does not add time

bkutil:


bkutil:
cc: @dmehra @go-oleg @dmajda

bkutil:
refs: #77

dmehra:
Does this also cover "not emitting ticks and emitting all points at once" as per last comments in #77?

bkutil:
Yes, updated the commit message.

demmer:
This looks good, though I think we want to add a check that the user shouldn't use `-points` with any of `-from, -to, -last, -limit, -every` since they'll now be ignored.

demmer:
Also given this new behavior, you should modify `next_point` so it doesn't have the check for `this.points` since that'll never happen any more.

bkutil:
> This looks good, though I think we want to add a check that the user shouldn't use -points with any of -from, -to, -last, -limit, -every since they'll now be ignored.

Ok, I added the missing checks and fixed the tests.

> Also given this new behavior, you should modify next_point so it doesn't have the check for this.points since that'll never happen any more.

Won't `this.points` check still be needed when we iterate over `-points` containiing time?

lib/runtime/procs/emit.js
```diff
@@ -54,23 +54,30 @@ var emit = source.extend({
             throw this.compile_error('RT-EMIT-TO-LIMIT-ERROR');
         }
         if (_.has(options, 'points')) {
+            // Invalid options
             if (_.has(options, 'to') || _.has(options, 'limit')) {
                 throw this.compile_error('RT-EMIT-POINTS-LIMIT-ERROR');
             }
+            if (_.has(options, 'from') || _.has(options, 'every')) {
```
demmer: What about `-to`?

demmer:
> Won't this.points check still be needed when we iterate over -points containing time?

Oh of course -- I misread the PR. You're correct.

Modulo the missing check for `-to` I think this looks good. We will have a conflict once #248 merges since I reworked a bit of the internals of `emit`, but that should be easy enough to resolve.

lib/runtime/procs/emit.js
bkutil: Three lines above this one? :)

PR #277, opened 2016-01-27
add delete and rename for time

VladVega:
1. delete `pt[timeField]` after putting the value into `pt.time`. 
2. if there exists a `pt.time` with a separate `timeField`, avoid simply discarding by setting that value to key `pt._time`.  

(also note that I will have to do a little rework on this after https://github.com/juttle/juttle/pull/248 since the function moves to a new class - but the function does not change in logic in that pr)

This PR originated from @dmajda comment here: https://github.com/juttle/juttle-sql-adapter-common/pull/16

And has to do with bringing this back: https://github.com/juttle/juttle-sql-adapter-common/pull/14

### DIFF
--- a/lib/runtime/procs/source.js
+++ b/lib/runtime/procs/source.js
@@ -28,7 +28,7 @@ var source = base.extend({
                 }
                 if (timeField !== 'time') {
                     if (pt.time) {
-                        pt._time = pt.time;
+                        self.trigger('warning', self.runtime_error('RT-OVERWRITING-TIME', {field: timeField}));
                     }
                     pt.time = pt[timeField];
                     delete pt[timeField];
--- a/lib/runtime/procs/source.js
+++ b/lib/runtime/procs/source.js
@@ -22,17 +22,21 @@ var source = base.extend({
         var self = this;
 
         _.each(points, function(pt) {
-            if (timeField && !pt[timeField]) {
-                self.trigger('warning', self.runtime_error('RT-POINT-MISSING-TIME', {field: timeField}));
+            if (timeField) {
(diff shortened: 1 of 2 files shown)


lib/runtime/procs/source.js
```diff
@@ -22,17 +22,21 @@ var source = base.extend({
...
         _.each(points, function(pt) {
-            if (timeField && !pt[timeField]) {
-                self.trigger('warning', self.runtime_error('RT-POINT-MISSING-TIME', {field: timeField}));
+            if (timeField) {
+                if (!pt[timeField]) {
+                    self.trigger('warning', self.runtime_error('RT-POINT-MISSING-TIME', {field: timeField}));
+                }
+                if (timeField !== 'time') {
+                    if (pt.time) {
+                        pt._time = pt.time;
```
demmer: In talking this over with @dmajda this morning, we agreed that having this "magic" behavior was undesirable.

Instead the logic should be that if there is a field called `time` in the point and the `timeField` is not time (either undefined or something else), then we should throw a warning and remove `time` from the point.

To support the case where the user wants to keep that field but stash it in something else, then we should add a `-timeRenameField` option which would implement the logic to move the conflicting time into another field. I don't love the name of that option FWIW but I think it's the least bad that I could think of.

VladVega:
@demmer what do you thing about this solution?

VladVega:
@davidvgalbraith also.

VladVega:
Notes from @demmer:
1. Consider toNative + parseTime => same function.
2. Ensure point is dropped with error when `time = foo`.

PR #280, opened 2016-01-28
expose -format for `write file`

### LINKED ISSUE #125: file adapter should expose format for writing
This is very simple to do now that we've introduced the notion of serializers and can easily expose csv, json and jsonl output formats.

rlgomes:
fixes #125

remove the ability to append to a file when writing and simply rely on the
serializers to write out data in the various supported serializer formats.

we can revisit adding a `-append` option to the file adapter when it
becomes a requirement.

### DIFF
--- a/lib/adapters/file/write.js
+++ b/lib/adapters/file/write.js
@@ -4,11 +4,12 @@ var _ = require('underscore');
 var Juttle = require('../../runtime/index').Juttle;
 var Promise = require('bluebird');
 var fs = Promise.promisifyAll(require('fs'));
+var serializers = require('../serializers');
 
 var Write = Juttle.proc.sink.extend({
     procName: 'write-file',
     initialize: function(options, params) {
-        var allowed_options = ['file', 'limit', 'maxFilesize', 'bufferLimit'];
+        var allowed_options = ['file', 'format'];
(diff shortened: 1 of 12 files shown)


rlgomes:
@davidvgalbraith @demmer as we discussed lets just drop the append behavior and then add the `-format` option so we can write out data in any of the formats supported by our serializers.

rlgomes:
I'll fix the coverage issue in the morning.

demmer:
I think it's useful to continue to support append for formats like `jsonl` and maybe `csv`.

rlgomes:
@demmer opened an issue to follow up once this goes in #286

lib/adapters/file/write.js
```diff
@@ -24,120 +25,27 @@ var Write = Juttle.proc.sink.extend({
...
-        currentFlush.then(function() {
-            // after the current flush is done lets make sure to fire off
-            // another one as the previous one could have been somewhere
-            // in the middle of the async write and without another flush we
-            // would leave points on the self.queue
-            return self.flush();
-        })
-        .then(function() {
-            self.done();
+        this.serializer.done()
```
davidvgalbraith: Should probably return this promise.

davidvgalbraith:
+1

PR #281, opened 2016-01-28
Improve FilterJSCompiler tests

dmajda:
Rewrite the `FilterJSCompiler` tests so that they don’t use ASTs, match structure of tested code better, and are more idiomatic. See individual commit messages for details.

Overall, these changes will make work with tests easier as capabilities of `FilterJSCompiler` are extended. 

Part of work on #57.


dmajda:
@bkutil

bkutil:
+1

PR #282, opened 2016-01-28
Extend FilterJSCompiler

dmajda:
Extend `FilterJSCompiler` to handle AST nodes that will be part of dynamic filters. It can now handle essentially all expressions. The code was mostly adapted from `code-generator.js`, `build.js` and `graph-compiler.js` (and yes, it would be good to unify these instead of duplication — but that has to wait as we need to do more high-priority work now).

Note the chosen method of testing didn’t allow to test some expressions (e.g. function calls). These were noted in the test file and tested manually (with lifted filter checks). One way to test them would be to rewrite the tests into JuttleSpec, but that would make them much bigger and more clumsy so I didn’t want to do that (at least not now).

Part of work on #57.

### DIFF
--- a/lib/compiler/filters/filter-js-compiler.js
+++ b/lib/compiler/filters/filter-js-compiler.js
@@ -123,6 +123,18 @@ var FilterJSCompiler = ASTVisitor.extend({
         return 'juttle.ops.str(' + this.visit(node.expression) + ')';
     },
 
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
Looks good (addressed the typo issue over Slack). +1

PR #283, opened 2016-01-28
Split FilterChecker and enable dynamic filters

dmajda:
Split `FilterChecker` into `StaticFilterChecker` and `DynamicFilterChecker`, which are then tweaked and used in appropriate places in order to support dynamic filters in `filter`.

Final step of #57.


dmajda:
@bkutil

bkutil:
Looks good. +1

PR #285, opened 2016-01-28
Juttle tutorial

dmehra:
Fixing up the tutorial, not for merge yet, but for @demmer to look at.

### DIFF
--- a/docs/concepts/juttle_tutorial.md
+++ b/docs/concepts/juttle_tutorial.md
@@ -2,27 +2,1055 @@
 title: Juttle Tutorial | Juttle Language Reference
 ---
 
-## Juttle Tutorial
+# Juttle Tutorial
 
-:warning: XXX/dmehra this is unfinished
+[TOC]
 
-If you're here, you're probably just starting out with Juttle.
+This tutorial is meant for new users who have a basic understanding of Juttle, having read at a minimum the [Overview](overview.md).
 
-Juttle is a high-level language that is used to process and visualize
-streams of data. At its core lies a simple dataflow language which is
-used to declare and define processing flowgraphs, by stitching together
-individual processing steps into a directed graph.
+By stepping through the examples, we will practice coding in Juttle, including reading data from files, s3 and Elasticsearch; writing data out; filtering and aggregating data; visualizing data as charts in the browser.
 
-We'll introduce you to the key concepts in Juttle, starting off with
-simple synthetic data, and then graduating to a real dataset coming from
-GitHub.
+## Setup
 
-After completing these tutorials, you will know enough Juttle to start
-exploring and analyzing your own data. For more detailed Juttle
-documentation than this, see [Juttle concepts](juttle_overview.html). 
+This tutorial works best if you install the juttle-engine assembly that provides both the CLI and the browser execution environment. This can be done via `npm install juttle-engine`, or using a docker container we provide. The demo system demo.juttle.io is a public deployment of that container.
 
-In Juttle, *the basic unit of data is a point*. A point consists of a
-number of key/value pairs, where the keys are strings and the values are
-numbers, strings, Booleans, or null. Points flow through flowgraphs and
-can be transformed, aggregated, or joined at each processing step.
+### Demo Option
 
+You can follow along by executing the Juttle programs, and viewing their source code, on the demo system at URL:
+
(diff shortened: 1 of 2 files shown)


dmehra:
Finally, this is the tutorial text that matches juttles in https://github.com/juttle/juttle-engine/pull/85

The [RUN] links won't work right now, as the new container with juttles hasn't been deployed to demo.juttle.io yet, it will be after juttle-engine PR#85 merges.

@demmer @dmajda @mstemm 
@juttle/developers if there's general interest in the long-coming tutorial.

docs/concepts/juttle_tutorial.md
```diff
@@ -2,27 +2,1055 @@
...
+To launch the Juttle CLI in interactive mode, you would do:
+
+```
+docker exec -it examples_juttle-engine_1 juttle
+```
+
+To run programs via CLI from saved juttle files, you would do:
+
+```
+docker exec examples_juttle-engine_1 juttle /examples/tutorial/filename.juttle
```
mstemm: I think the path within the container is /opt/juttle-engine/juttles/examples, so when running juttle you need to specify that path, not the path relative to the juttle-engine root.

docs/concepts/juttle_tutorial.md
```diff
@@ -2,27 +2,1055 @@
...
+```
+read elastic -from :0: -to :now:
+|(
+  head 1;
+  tail 1
+)
+| keep time
+```
+
+XXX using workaround for [bug #115](https://github.com/juttle/juttle-elastic-adapter/issues/115)
```
mstemm: Want to replace this with a friendlier comment?

docs/concepts/juttle_tutorial.md
```diff
@@ -2,27 +2,1055 @@
...
+
+*27_timechart_moving_window.juttle* [[RUN](http://demo.juttle.io/?path=/examples/github-tutorial/27_timechart_moving_window.juttle)]
+
+```
+read elastic -from :2015-01-01: -to :2015-07-01:
+| reduce -over :2 weeks: -every :day: count=count()
+| view timechart
+```
+
+XXX this hits [bug #116](https://github.com/juttle/juttle-elastic-adapter/issues/116)
```
mstemm: Same here, I won't mention any other similar cases.

mstemm:
Do you really want the raw data both in the juttle and juttle-engine repositories? I'd guess just juttle-engine.

mstemm:
I'm ok with merging. I didn't actually run through the tutorial to make sure every program worked but I'm happy with the discussion and examples.

I think a good follow on would be to try it out with a new user to Juttle and get their feedback.

mstemm:
How about adding a link to the tutorial from the top level README?

go-oleg:
Thoughts on putting this tutorial in the `juttle-engine` repo instead? I'm concerned about all the references to juttle-engine and browser-based views creating a circular reference between juttle-engine and juttle.

This makes it unclear when juttle-engine specific aspects of the tutorial should be updated. That is, we shouldn't update them when releasing juttle because juttle-engine hasn't been released and by the time we release juttle-engine, juttle has already been released.

dmehra:
The bugs I referenced should disappear soon... very soon: https://github.com/juttle/juttle-elastic-adapter/pull/118. But I suppose the fix won't make it into the docker container for another 2 weeks. I'm not sure what to do in a case like this - I can make the notes prettier, but how will I even remember to update the tutorial when next release goes out? Ugh. 

As for the tutorial location: we originally wanted it in juttle repo so it can be published at the doc site (which isn't even a thing for juttle-engine), and so relative links to other doc articles work. I think those reasons are still valid.

Does the circular reference remain a problem if you consider that the doc site is published out of band anyway (that is, it's published whenever i remember to do so)? We can hold off on updating the doc site until juttle-engine release is done, so the tutorial is synced up with that (and not with juttle).

go-oleg:
The _out-of-band_ issue can/should be solved by publishing docs in `scripts:prepublish` in `package.json`.

Since we don't have docs in juttle-engine, I can't think of a better place to put this right now. Guess we can leave it as is.

dmehra:
@mstemm another look with updates? I'll squash before merging. The travis failure is not me, syntax check passes on the tutorial juttles.

mstemm:
lgtm.

PR #290, opened 2016-01-28
remove node-grok from the default juttle installation

### LINKED ISSUE #289: make node-grok optional when installing
there's no reason to depend on the node-grok library that pulls in a few other things (mainly the oniguruma parser) and makes our install a good bit longer and depending on some c code compiling which we can avoid and simply let users konw to `npm install xyz` if they intend on taking advantage of the grok parser.

rlgomes:
fixes #289

now node-grok will not be installed by npm and the user that has any
interest in using the grok parser will be prompted to install it
themselves (currently off our own fork).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ node_js:
 
 before_script:
     - npm install -g gulp
+      # grok parser tests need node-grok
+    - npm install rlgomes/node-grok#b3b810f4277d7b9d765853214bf20adfdb010ebe
 
 script:
     - gulp lint examples-check test-coverage
--- a/lib/adapters/parsers/grok.js
+++ b/lib/adapters/parsers/grok.js
@@ -2,7 +2,7 @@
 
 var base = require('./base');
 var errors = require('../../errors');
-var grok = require('node-grok');
+var grok; // initialized in constructor
 var Promise = require('bluebird');
 var readline = require('readline');
 
@@ -10,6 +10,19 @@ module.exports = base.extend({
 
     initialize: function(options) {
         this.pattern = options.pattern || '%{GREEDYDATA:message}';
+
+        try {
+            grok = require('node-grok');
+        } catch(err) {
+            if (err.toString().match(/Cannot find module 'node-grok'/)) {
+                // we're doing this to avoid installing a more heavy weight dependency
+                // as part of the regular juttle dependencies.
+                throw Error('Grok parser requires a manual installation of node-grok like so:\n' +
+                            ' > npm install rlgomes/node-grok#b3b810f4277d7b9d765853214bf20adfdb010ebe\n');
+            } else {
+                throw err;
+            }
+        }
     },
 
     parseStream: function(stream, emit) {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "moment-duration-format": "^1.3.0",
     "moment-parser": "^0.3.0",
     "moment-timezone": "^0.4.0",
-    "node-grok": "github:rlgomes/node-grok#b3b810f4277d7b9d765853214bf20adfdb010ebe",
     "oboe": "^2.1.2",
     "pegjs": "^0.9.0",
     "request": "^2.67.0",
--- a/test/adapters/file/file.spec.js
+++ b/test/adapters/file/file.spec.js
@@ -7,6 +7,7 @@ var path = require('path');
 var tmp = require('tmp');
 var fs = require('fs');
 var juttle_test_utils = require('../../runtime/specs/juttle-test-utils');
+var withModuleIt = juttle_test_utils.withModuleIt;
 var check_juttle = juttle_test_utils.check_juttle;
 var expect_to_fail = juttle_test_utils.expect_to_fail;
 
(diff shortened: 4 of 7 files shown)


rlgomes:
@demmer made the magic happen, unfortunately i can't find a good way to test this than what I did locally to verify you can use `read file` without having the `node-grok` packaged installed and things work till you attempt to use `-format 'grok'` and then are told to:

```
rlgomes@x230> ./bin/juttle -e "read file -file '/var/log/syslog' -format 'grok' -pattern '%{SYSLOGLINE}'"
Error: Grok parser requires a manual installation of node-grok like so:
 > npm install rlgomes/node-grok#b3b810f4277d7b9d765853214bf20adfdb010ebe
```

Let me know what you think I can also add a similar set of instructions on installing node-grok to the docs.

lib/adapters/parsers/index.js
```diff
@@ -28,13 +27,6 @@ module.exports = {
...
 
-        if (parser) {
-            return new parser(options);
-        } else {
-            throw errors.runtimeError('RT-INVALID-OPTION-VALUE',{
-                option: 'format',
-                supported: _.keys(parserLookup).join(', ')
-            });
-        }
+        return new parser(options);
```
demmer: It seems to be simpler to keep this file the way it was and just throw an error in the grok parser initialize() function if you can't require grok.

rlgomes: Sure

rlgomes:
@demmer a lot less lines changed, this does make running the tests locally as a developer a bit of a pain when it fails to tell you to run the `npm install node-grok` command.

demmer:
I like this implementation better but I don't like the idea that a run of `gulp test` will fail unless we install `node-grok` by hand, since that kinda defeats the point of making this optional.

How about if we add a little test utility that checks if `node-grok` is installed and then we use that instead of `it()`. Something like:

```
// grok-test-utils: simple wrapper to automatically skip grok tests if node-grok isn't installed

var grok;
try {
    grok = require('node-grok');
} catch(err) {
    console.log('WARNING: skipping all tests that depend on grok');
}
function ifGrok(description, function) {
    if (grok) {
        it(description, function);
    } else{
        it.skip(description, function);
    }
}
```

Then all the tests that depend on grok would use `ifGrok(...)` instead of `it(...)`.

rlgomes:
@demmer checkout the last commit I built a test it that can accept an options argument and then you can do special requirements. This can be used or our case here of needed a specific module that we don't want to install by default but can also be used for other requirements like depednent services, etc. Let me know what you think.

demmer:
LGTM. +1

dmajda:
Late to the party, but I don’t think the module check should be conflated with the `it` function — they are two different things. I’d find it much cleaner to see something like this in the tests:

```javascript
ifModuleInstalled('node-grok', function() {
    it('can instantiate a grok parser', function() {
        var parser = parsers.getParser('grok');
        expect(parser).to.not.be.undefined();
    });

    // ...
});
```

Or perhaps more generically:

```javascript
condition({ installedModules: 'node-grok' }, function() {
    it('can instantiate a grok parser', function() {
        var parser = parsers.getParser('grok');
        expect(parser).to.not.be.undefined();
    });

    // ...
});
```

The implementation would just pass the function with `it` blocks to `describe`/`describe.skip` with synthesized description (e.g. “with node-grok module installed”).

@rlgomes What do you think?

rlgomes:
@dmajda not sure how would the `condition` or `ifModuleInstalled` functions modify the underlying it function calls ? so they would all show up as skipped ? instead of not appearing in the test report at all ? or just under a single skip ?

dmajda:
@rlgomes My idea was that `ifModuleInstalled`/`condition` would just morph into `describe` or `describe.skip`, like this:

```javascript
function ifModuleInstalled(module, fn) {
    var installed = checkForModule(module);
    var description = 'with module "' + module + '" installed';
  
    (installed ? describe : describe.skip)(description, fn);
}
```

rlgomes:
@dmajda understood the only problem is that there are a few just `it` tests in some files like these ones https://github.com/juttle/juttle/pull/290/files#diff-7a886cc67329f5d0b14d46f836d42702R197 that I'd have to move into another describe block ? This was just to remove the installing/building of the oniguruma package from building its c dependencies in our juttle runtime. I have something I'm working on which could replace the whole `node-grok` package by a much simpler depdendency that wouldn't rely on the oniguruma package.

dmajda:
> the only problem is that there are a few just `it` tests in some files like these ones https://github.com/juttle/juttle/pull/290/files#diff-7a886cc67329f5d0b14d46f836d42702R197 that I'd have to move into another describe block ?

Not sure what you mean by “move into another describe block”. On the source code level, you’d just replace:

```javascript
withModuleIt('fails to optimized tail followed by head with -format "grok"', function() {
    //...
}, 'node-grok');

withModuleIt('can optimize nested "| head 2 | head 1" with -format "grok"', function() {
    //...
}, 'node-grok');

withModuleIt('can optimize "| head 1" with -format "grok"', function() {
    //...
}, 'node-grok');
```

with:

```javascript
ifModuleInstalled('node-grok', function() {
    it('fails to optimized tail followed by head with -format "grok"', function() {
        //...
    });

    it('can optimize nested "| head 2 | head 1" with -format "grok"', function() {
        //...
    });

    it('can optimize "| head 1" with -format "grok"', function() {
        //...
    });
});
```

I find that much cleaner.

> I have something I'm working on which could replace the whole `node-grok` package by a much simpler depdendency that wouldn't rely on the oniguruma package. 

OK, if this is temporary then it’s probably not worth tweaking.

PR #292, opened 2016-01-29
Remove pubsub

bkutil:
As the `TestView` was the last view depending on the pub sub system, modify it so it uses `program.events`. Then gradually remove dependencies on the `publish` and `subscribe` procs from the inheritance chains and from the code.

### DIFF
--- a/lib/program.js
+++ b/lib/program.js
@@ -56,24 +56,6 @@ var Program = Base.extend({
 
         this.scheduler.start();
     },
-    // connect a publish/subscribe channel that is inside the flowgraph closure to the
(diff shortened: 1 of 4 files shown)


bkutil:
cc: @dmajda @demmer

dmajda:
+1

@bkutil The [CI failure](https://travis-ci.org/juttle/juttle/builds/105644840) seems to be because of decreased coverage, which seems odd given this PR mostly removes code. Please investigate and adjust the numbers if necessary before merging.

bkutil:
Rebased

bkutil:
I'll move the channel allocation to a method on program in a separate PR.

PR #293, opened 2016-01-29
append option for file adapter

### LINKED ISSUE #286: add -append option to file adapter
at least for `jsonl` and `csv` this should be doable.

rlgomes:
fixes #286

This will enable the ability to append on `write file` for `jsonl` and
`csv` formats. We don't verify that the existing files are valid in any
way and the only "trick" here is that for CSV if the file exists we
assume it has an existing header with valid data and append to it.

documentation updated


rlgomes:
@juttle/developers 

@demmer you ask and you shall receive

rlgomes:
@demmer rebased so we could get this in before the upcoming repo changes ?

rlgomes:
@demmer updated with your request for reading back the first line to determine the csv headers and added 2 tests.

lib/adapters/file/write.js
```diff
@@ -29,25 +30,95 @@ class WriteFile extends AdapterWrite {
...
+        this.serializerOptions = {};
+        if (options.append && !_.contains(['jsonl', 'csv'], options.format)) {
+            throw new errors.compileError('RT-INVALID-OPTION-COMBINATION',{
+                option: 'append',
+                rule: 'format="csv" or format="jsonl"'
+            });
+        }
+    }
+
+    _readHeaders() {
```
demmer: Please add a comment explaining why this is here.

lib/adapters/file/write.js
```diff
@@ -29,25 +30,95 @@ class WriteFile extends AdapterWrite {
...
+        if (this.append && this.format === 'csv') {
+            return fs.accessAsync(this.filename)
+            .then(() => {
+                return new Promise((resolve, reject) => {
+                    // since the file exists lets make sure to let the csv
+                    // serializer to not send the headers
+                    this.serializerOptions.sendHeaders = false;
```
demmer: This all really seems like it better belongs inside the serializer instead of here... obviously this is a little tricky since the current serializer API takes a stream and not the filename, but this really feels like a lot of logic that should be abstracted away in some better manner.

rlgomes:
@demmer added a comment to the `_readHeaders` function but I don't think there's much other than moving that logic to read the headers out of the file adapter code as a helper function off of the csv parser that can be done to move that code to a better location.

rlgomes:
bump!

test/adapters/file/read.file.spec.js
```diff
@@ -0,0 +1,304 @@
+'use strict';
```
davidvgalbraith: I think `read.spec.js` would be a better name for this file since it's already in the `file` directory.

lib/adapters/file/write.js
davidvgalbraith: I agree, gonna want the same power in the S3 adapter without reimplementing all this.

rlgomes: I shall figure out what to do then but for this to work @davidvgalbraith you'd have to supply an `ReadableStream` and `WriteableStream` when appending so we can read the first line for CSV and then do the right thing within the serializer ?

rlgomes:
So I'm going to redo this PR by first getting the test split up don here: https://github.com/juttle/juttle/pull/345#partial-pull-merging and then putting together a PR to add the ability to append to the serializers that can do append (https://github.com/juttle/juttle/issues/346) and then a simple PR to use the append feature from the file adapter itself.

PR #294, opened 2016-01-29
accept date object as valid value

VladVega:
Accept Date Object as valid value.

@davidvgalbraith @demmer 

(postgres adapter contains date value when value is not under time field.)

### DIFF
--- a/lib/runtime/values.js
+++ b/lib/runtime/values.js
@@ -105,6 +105,9 @@ var values = {
                         case '[object Array]':
                             return 'Array';
 
+                        case '[object Date]':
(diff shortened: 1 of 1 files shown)


demmer:
It seems to me that any Date objects should be wrapped in JuttleMoments before sending them from an adapter, regardless of whether they're in the `time` field or elsewhere.

I think this change will make the runtime think it's a JuttleMoment when it's really a Date, but any time operations will fail since it'll expect the former.

VladVega:
Ok. SQL adapter will iterate through all point values and turn all dates into JuttleMoments.

PR #295, opened 2016-01-29
do not multiply any number by 1000

VladVega:
I know I missed the ball here on the conversation before but:

Why are we assuming that any number in the time field is in seconds? Wouldn't it be better to not make this assumption?

The contract would then be that any valid param passed into JS `Date` is ok. I think that makes more sense than multiplying any number by 1000 no?

@demmer @davidvgalbraith @go-oleg @dmajda


VladVega:
Closing this one => let's eventually accept only dates in the parse time function?

PR #296, opened 2016-01-29
Adding http server doc to the docs tree

dmehra:
Merging this so docs can be deployed with this new file.


dmehra:
Ignore the travis failure that was entirely unrelated to this one-liner docs change.

PR #297, opened 2016-02-01
Consolidate Juttle runtime: Independence on JuttleMoment and Filter

dmajda:
Make compiled Juttle programs independent on `JuttleMoment` and `Filter` variables.

Part of work on #97.


dmajda:
@bkutil

bkutil:
+1

PR #299, opened 2016-02-01
Remove test/runtime/humanized-moments.spec.js

dmajda:
Removed test file exercised moment/duration parsing which is now handled by `moment-parser` and tested there.


dmajda:
@demmer

demmer:
Although somewhat redundant as you point out, my instinct is that we should
keep these tests to ensure that any changes to the grammar in moment-parser
(which affect the juttle semantics) are reflected properly in juttle.



On Mon, Feb 1, 2016 at 4:35 AM, David Majda <notifications@github.com>
wrote:

> @demmer <https://github.com/demmer>
>
> —
> <https://github.com/juttle/juttle/pull/299#issuecomment-177954474>.
>



-- 
Michael Demmer
Jut - CTO / VP Engineering
demmer@jut.io

dmajda:
Thinking about this more, I think you are right. In principle, we should treat `moment-parser` as just another library which can be potentially replaced by some other implementation (however unlikely that is). From that POV, (re-)testing moment syntax & semantics here makes sense.

PR #300, opened 2016-02-01
Simplify lib/parser structure

dmajda:
Instead of wrapping `lib/parser/juttle-parser.js` into a nearly empty `lib/parser/index.js` module, let's put all the functionality into `lib/parser/index.js` directly.


dmajda:
@bkutil

bkutil:
+1

PR #301, opened 2016-02-01
ESLint move cleanups

dmajda:
After [moving to ESLint](https://github.com/juttle/juttle/pull/203) there still remained few places where JSHint and JSCS were referenced. Let’s eliminate them.


dmajda:
@rlgomes

rlgomes:
+1 thanks for catching these

dmajda:
@rlgomes Thank `grep` :-)

PR #302, opened 2016-02-01
Do not call JuttleMoment.duration with "new"

dmajda:
`JuttleMoment.duration` is not a constructor (just a factory function), so it should not be called using the `new` operator. This PR removes `new` from any such calls.


dmajda:
@bkutil

bkutil:
Looks good. +1

PR #306, opened 2016-02-01
test: fix checks for no errors / warnings

demmer:
When testing that a juttle program ran with no errors or warnings,
use `expect(result.errors).deep.equals([])` instead of checking for
a zero-length array since in the event of a failure, it will actually
print what the errors are instead of uselessly reporting that some
error happened but not indicating what error it was.


dmajda:
+1

(Didn’t check whether all instances were changed.)

PR #309, opened 2016-02-02
Values reducers

bkutil:
Instead of native javascript arithmetic operators, use `ops.add` and `ops.sub` in reducers. This ensures that the operations are type-checked and that adding incompatible types or types for which the operation is not defined, throws an error, instead of being silently converted.

### DIFF
--- a/lib/runtime/reducers.js
+++ b/lib/runtime/reducers.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var JuttleMoment = require('../moment').JuttleMoment;
 var DEQueue = require('double-ended-queue');
 var Digest = require('tdigest').Digest;
 var values = require('./values');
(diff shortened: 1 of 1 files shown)


bkutil:
cc: @dmajda

dmajda:
I went through `reducers.js` and I think these instances still need to be converted:

  * [`result` in `avg`](https://github.com/juttle/juttle/blob/values-reducers/lib/runtime/reducers.js#L116) — one can average e.g. durations since `duration + duration` and `duration / number` are both defined
  * [`result` in `sigma`](https://github.com/juttle/juttle/blob/values-reducers/lib/runtime/reducers.js#L290) — I don’t think this line can get executed for anything but number (because all other types will fail to type check in `update` already), but I don’t think we want to embed this knowledge into the reducer

dmajda:

dmajda:
+1

bkutil:
For the record, we had a discussion on slack whether `math.sqrt` in sigma should use JS native `Math` or Juttle wrapper. We decided to call Juttle sqrt for consistency (i.e. reducers should not care / be aware of which operations are defined for which types - that's a responsibility of modules/`ops`).

Will merge on green.

PR #310, opened 2016-02-02
Consolidate Juttle runtime: Don’t set Juttle.FilterJSCompiler

dmajda:
Anyone who needs to use `FilterJSCompiler` should require it directly.

Can be merged only after juttle/juttle-opentsdb-adapter#7 and juttle/juttle-sql-adapter-common#19 get merged.

Part of work on #97.


dmajda:
@bkutil

bkutil:
+1

PR #311, opened 2016-02-02
Don’t set Juttle.reducers

dmajda:
Don't set `Juttle.reducers`. Instead, add reducers to the runtime and let compiled Juttle programs use them as `juttle.reducers.*`. This makes the programs less dependent on the `Juttle` variable.

Part of work on #97.


dmajda:
@bkutil

bkutil:
+1

PR #312, opened 2016-02-02
Move classes representing Juttle types to lib/runtime/types

dmajda:
The idea is that `lib/runtime/types` is the right place for classes representing Juttle types which can’t be represented by native JavaScript values (currently these classes are `JuttleMoment` and `Filter`).

As part of the move I also dissolved `lib/moment/index.js`, which was just a thin and useless wrapper around `juttle-moment.js`. This necessitated changing many `require` paths, which was a pain but they are now more logical and consistent with other requires (no `.JuttleMoment` at the end).

**Note this change will also require adapting `require` paths in adapters.** They should be changed from:

```javascript
var JuttleMoment = require('juttle/lib/moment').JuttleMoment;
```

to:

```javascript
var JuttleMoment = require('juttle/lib/runtime/types/juttle-moment');
```


dmajda:
@bkutil As discussed yesterday…

bkutil:
+1

PR #314, opened 2016-02-02
expose the list of adapters to the runtime

### LINKED ISSUE #313: expose debugging information about loaded adapters
Because of the current requirement that juttle and the adapters need to be installed side-by-side as npm modules in order for each to be able to `require` the other, we've run into situations in which it's not immediately clear where the adapter module is actually loaded from.

To help address this, the juttle runtime should expose the list of configured adapters with the version and the path from which it is loaded.

demmer:
Add an `adapters.list()` API function that exposes the list of configured adapters with their versions and paths on disk.

Expose this as a constant array in the `Juttle` module so it can be printed by a simple program like:

```
emit -points Juttle.adapters
┌─────────────┬───────────────────────────────────────────┬─────────┐
│ adapter     │ path                                      │ version │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ influx      │ /root/juttle-influx-adapter/index.js      │ 0.3.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ elastic     │ /root/juttle-elastic-adapter/lib/index.js │ 0.4.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ file        │ /root/juttle/lib/adapters/file/index.js   │ 0.4.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ http        │ /root/juttle/lib/adapters/http/index.js   │ 0.4.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ stdio       │ /root/juttle/lib/adapters/stdio/index.js  │ 0.4.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ stochastic  │ /root/juttle/lib/adapters/stochastic.js   │ 0.4.0   │
├─────────────┼───────────────────────────────────────────┼─────────┤
│ http_server │ /root/juttle/lib/adapters/http_server.js  │ 0.4.0   │
└─────────────┴───────────────────────────────────────────┴─────────┘
```

Also add missing documentation for the Juttle module.

Fixes #313

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -48,8 +48,9 @@ var implicit_views = require('../lib/compiler/flowgraph/implicit_views');
 var modes = _.without(compiler.stageNames, 'eval').concat('run');
 
 function usage() {
-    console.log('usage: juttle [--version] [--mode <mode>] [--config <config>] [--color/--no-color] [--show-locations] [--optimize] [--input name=val] [juttle-file]');
+    console.log('usage: juttle [--version] [--adapters] [--mode <mode>] [--config <config>] [--color/--no-color] [--show-locations] [--optimize] [--input name=val] [juttle-file]');
     console.log('     --version show juttle CLI version');
+    console.log('     --adapters show versions and paths to all configured adapters');
     console.log('     --mode <mode>: one of ' + modes.map(JSON.stringify).join(', '));
     console.log('     --config <config>: path to the juttle interpreter configuration file');
     console.log('     --optimize runs optimization');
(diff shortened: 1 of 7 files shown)


demmer:
@dmajda @mstemm @dmehra

demmer:
As per @dmehra's request also added support for `juttle --adapters` to print the list of configured adapters.

mstemm:
lgtm.

dmehra:
Let's merge!

dmehra:
Tangentially related: we handle command-line options with single - oddly, 
```
juttle -version
0.4.0
```
but 
```
juttle -help
Error: input.charCodeAt is not a function
```
And all options other than `-version` only work as `--optionName`. I went there because i thought, it's `-version` so it should be `-adapters` not `--adapters` but it turned out, only `-version` is special. This is all a bit strange, including the bad error "not a function".

demmer:
In talking things over with @dmajda we decided that it would be somewhat cleaner if the adapters list were exposed to the language as a function and not a constant.

Somewhat related, we also agreed to avoid using ES6 properties.

I'll push another commit that does both of these things then merge away.

demmer:
@dmehra The actual CLI behavior is that `-version` is actually parsed the same as `-v -e -r -s -i -o -n`. This "works" because we accept `-v` or `--version` and we don't reject invalid options.

In the case of `--adapters` there's no `-a` shortcut, hence `-adapters` doesn't work.

PR #315, opened 2016-02-02
parse time to specification

VladVega:
Changes coming from time discussion and code: https://github.com/juttle/juttle/pull/277

I realize I still need to fix some tests. But I'd rather put this out there and get some feedback before continuing.


VladVega:
@dmajda @demmer @davidvgalbraith

demmer:
FYI @VladVega see #339

VladVega:
Ok - check out my implementation of the spec? #339

davidvgalbraith:
What are we gaining from combining `parseTime` and `toNative`? This `toNativeTime` thing is too long, makes my head hurt.

lib/runtime/juttle-utils.js
```diff
@@ -68,6 +70,63 @@ function filterPoints(points, from, to) {
...
+                pt.time = new JuttleMoment({rawDate: date});
+            } else {
+                try {
+                    pt.time = new JuttleMoment(pt.time);
+                } catch(e) {
+                    if (self.trigger) {
+                        self.trigger('warning', new errors.runtimeError('RT-INVALID-TIME-ERROR', {time: pt.time}));
+                    }
+                    return null;
+                }
```
VladVega: @davidvgalbraith we could move this warning to `toNative` and use `toNative` function. Does that work?

VladVega:
I cleaned this up and rebased: maybe it's a little clearer now? take a look.

lib/runtime/juttle-utils.js
```diff
@@ -73,6 +75,61 @@ function filterPoints(points, from, to) {
...
+    }
+    return pt;
+}
+
+// Assigns the time from timeField to 'time' and attempts to convert it
+// into JuttleMoment
+//
+// If not specified, timeField defaults to 'time'. If specified and point
+// is missing that field, emits a warning using the supplied context.
+function toNativeTime(points, options) {
```
davidvgalbraith: The way it's set up, you have to invoke it like `toNativeTime.apply(this, [points, options])` which is pretty confusing. Would rather return an object that includes an array of valid points and an array of invalid points and have the caller worry about handling warnings.

VladVega: Ok but then each caller needs to handle the warnings and errors. It's just less reusable code, don't you think? The other way is to pass `this` as a context param but I like that less. Or maybe this code should go into `adapter-read.js`?

demmer: I feel like explicitly passing in the context is clearer than this approach.

I also think it makes total sense to add `AdapterRead.toNativeTime(points)` that just calls `toNativeTime(points, this.options, this)` so that in most of the cases we don't need to worry about it either way.

VladVega:
changed to `context`.

davidvgalbraith:
What is `timeRenameField` for again? I thought batches and merges and joins and maybe other stuff gets totally broken if the time isn't in the `time` field.

VladVega:
It's for when you want to use a different `timeField` like `create_time` but you want to save what is in your `pt.time` and stash it into a new field. The other possible use case is it you explicitly don't want a time field in your points.

VladVega:
Or just refer to https://github.com/juttle/juttle/issues/339

davidvgalbraith:
https://github.com/juttle/juttle/issues/339 calls for it but doesn't justify it. Has there actually been a case where we've needed it?

VladVega:
I forget if maybe @dmehra had a use case for this?

demmer:
Without `-timeRenameField` there would be no way for juttle to read any data that has an unparsable date in the `time` field.

davidvgalbraith:
Is that essential? Once upon a time I put up https://github.com/juttle/juttle/pull/170 and it got shot down to avoid option bloat, and I don't see this as having greater value than that.

VladVega:
IMO - Not essential but a nice to have.

VladVega:
poke?

VladVega:
Poke - any problems with this? Can I get a +1?

davidvgalbraith:
Ah well I'm -1 on `timeRenameField` as per the "let's not add inessential options" thinking.

VladVega:
@dmehra I remember you had a use case for this option `-timeRenameField` do you mind describing that use case? (the doc is here: #339)

demmer:
Unless there's a concrete use case where we need `-timeRenameField`, I'd be ok with leaving it out for now on the principle of not cluttering unnecessary options. I spoke with @dmajda and he's on the same page.

dmehra:
I don't have a use case right now, if there was one before it must have been reading from file and I worked around it by renaming the field before reading into juttle.

VladVega:
I revised the PR to exclude `-timeRenameField`.

davidvgalbraith:
+1

PR #316, opened 2016-02-03
adding support for -from/-to to http adapter

### LINKED ISSUE #194: http adapter: add support for -from/-to
Support time filtering in `read http`. Currently it does not support `-from`/`-to` options.

rlgomes:
fixes #194 

preliminary version of how to do get this done and also expose the
functionality in a way that we can apply to other adapters with ease.

### DIFF
--- a/lib/adapters/file/read.js
+++ b/lib/adapters/file/read.js
@@ -65,18 +65,7 @@ class ReadFile extends AdapterRead {
         var validPoints = [];
         return this.parser.parseStream(stream, (points) => {
             this.parseTime(points, this.timeField);
-
-            _.each(points, (point) => {
-                // timeless points just pass through at this point
-                if (point.time && point.time.moment) {
-                    if (point.time.gte(from) && point.time.lt(to)) {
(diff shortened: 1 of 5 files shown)


rlgomes:
@demmer @davidvgalbraith what do you think of this approach ? I can then apply it to stdio adapter as well so we get -from/-to support in there. I was also thinking we're starting to have many of these helper functions that `parseTime`, `toNative`, `filterPoints` that do multiple passes on the same array of points and could probably benefit from having a nice way of only doing a single passing while applying N functions which would modify/drop the point. The exact way of doing this can be discussed but I think we'd benefit quite drastically in terms of performance from not doing the multiple passes.

rlgomes:
bump!

davidvgalbraith:
If you do X work in one pass and Y work in another pass, or if you do X + Y work in a single pass, you've still done the same amount of work. So I don't think there'd be a noticeable performance difference from consolidating like that. +1 on this, though.

demmer:
lgtm

PR #317, opened 2016-02-03
adapters: validate options in base class

davidvgalbraith:
@juttle/developers

### DIFF
--- a/lib/adapters/http_server.js
+++ b/lib/adapters/http_server.js
@@ -12,7 +12,6 @@ var Read = source.extend({
     procName: 'read-http_server',
 
     initialize: function(options, params, location, program) {
-        this.options = options;
         var allowedOptions = ['port', 'method', 'timeField', 'rootPath'];
         this.validate_options(allowedOptions);
         var allowed_methods = [ 'POST', 'PUT' ];
--- a/lib/adapters/http_server.js
+++ b/lib/adapters/http_server.js
@@ -12,22 +12,10 @@ var Read = source.extend({
     procName: 'read-http_server',
 
     initialize: function(options, params, location, program) {
+        this.options = options;
+        var allowedOptions = ['port', 'method', 'timeField', 'rootPath'];
+        this.validate_options(allowedOptions);
         var allowed_methods = [ 'POST', 'PUT' ];
-
-        var allowed_options = [
-            'port',
-            'method',
-            'timeField',
-            'rootPath'
-        ];
-
-        var unknown = _.difference(_.keys(options), allowed_options);
-        if (unknown.length > 0) {
-            throw this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
-                proc: 'read http_server',
-                option: unknown[0]
-            });
-        }
 
         if (options.method && allowed_methods.indexOf(options.method) === -1) {
             throw this.compile_error('RT-INVALID-OPTION-VALUE', {
--- a/lib/runtime/adapter-read.js
+++ b/lib/runtime/adapter-read.js
@@ -69,6 +69,11 @@ class AdapterRead {
     // Common read options that should be supported by all adapters.
     static get commonOptions() { return ['from', 'to', 'last', 'timeField', 'every', 'lag']; }
 
+    // subclasses can override this method to return an array
+    // if they do we'll check that all options passed to the
+    // proc are in the array
+    allowedOptions() { return false; }
+
     // Assigns the time from timeField to 'time' and attempts to convert it
     // into JuttleMoment.
     parseTime(points, timeField) {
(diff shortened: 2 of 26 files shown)


rlgomes:
+1 looks good, only wondering why not set `this.options = options` in the `procs/base.js` so you don't have to do it in a few places just to be able to run the `validate_options` method which seems a little clunky ?

lib/adapters/http_server.js
```diff
@@ -12,23 +12,11 @@ var Read = Juttle.proc.source.extend({
     procName: 'read-http_server',
 
     initialize: function(options, params, location, program) {
+        this.options = options;
+        var allowedOptions = ['port', 'method', 'timeField', 'rootPath'];
+        this.validate_options(allowedOptions);
```
demmer: This adapter doesn't follow the model of the others. Why is that?

lib/runtime/procs/read.js
```diff
@@ -81,6 +81,13 @@ var Read = source.extend({
             throw err;
         }
 
+        this.procName += ' ' + adapter.name;
+        var adapterAllowedOptions = this.adapter.allowedOptions();
+        if (adapterAllowedOptions) {
+            var allAllowedOptions = AdapterRead.commonOptions.concat(adapterAllowedOptions);
```
demmer: All things being equal, I don't think it makes sense to mandate that all adapters necessarily accept the common read options. Although slightly more work for the adapter implementation, it seems safer if the subclass allowedOptions explicitly opts in to the common read options.

lib/adapters/http_server.js
davidvgalbraith: It hasn't been converted to the fancy API.

davidvgalbraith:
Good idea @rlgomes, let's do that.

lib/runtime/adapter-read.js
```diff
@@ -69,6 +69,11 @@ class AdapterRead {
     // Common read options that should be supported by all adapters.
     static get commonOptions() { return ['from', 'to', 'last', 'timeField', 'every', 'lag']; }
 
+    // subclasses can override this method to return an array
+    // if they do we'll check that all options passed to the
+    // proc are in the array
+    allowedOptions() { return false; }
```
demmer: It seems like this would be cleaner if it were implemented as a static property in the same way that `commonOptions` is handled. That way the read implementation could validate the options without needing to actually instantiate the proc.

We could actually change `commonOptions` to be the default implementation of `allowedOptions` so that if an adapter just wants to support the common ones, they don't have to do anything, but if they want to extend then they would do something like:

```
static get allowedOptions() { return AdapterRead.allowedOptions.concat(['myOption']); }
```

lib/adapters/http_server.js
demmer: Oh of course... sloppy on my part.

davidvgalbraith:
Updated.

bkutil:
+1

lib/runtime/procs/base.js
```diff
@@ -32,6 +33,16 @@ var base = Base.extend({
         this.logger = JuttleLogger.getLogger(this.logger_name);
     },
 
+    validate_options: function(allowedOptions) {
+        var unknown = _.difference(_.keys(this.options), allowedOptions);
+        if (unknown.length > 0) {
+            throw new this.compile_error('RT-UNKNOWN-OPTION-ERROR', {
```
VladVega: I still think we should start using `new errors.compileError()` everywhere. But maybe that's just me.

demmer:
I had a long talk with @dmajda this morning about how we want to handle ES6 constructs like properties, and we concluded that despite what I said earlier, it would be better to avoid get/set properties and instead just explicitly use functions.

I'll write up the reasons for this in a separate issue that covers our plan for ES6, but as far as this specific PR goes, I'm going back a bit on what I said last night and now think we should change the `adapterOptions` property into a static function.

lib/runtime/adapter-read.js
```diff
@@ -69,6 +69,11 @@ class AdapterRead {
     // Common read options that should be supported by all adapters.
     static get commonOptions() { return ['from', 'to', 'last', 'timeField', 'every', 'lag']; }
 
+    // subclasses can override this method to return an array
+    // if they do we'll check that all options passed to the
+    // proc are in the array
+    static get allowedOptions() { return false; }
```
demmer: I thought we had decided that instead of returning false by default, this should return the list of commonOptions, i.e. the adapter gets only the common ones by default.

lib/runtime/procs/base.js
demmer: `this.compile_error` will always attach the location of the proc to the error so it's useful.

I actually think we should add a `this.compile_error` (or `this.compileError`) function to the adapters in order to make the calling conventions consistent.

Also note that in neither case should there be an explicit `new` -- the function returns a new instance so the proper usage is `throw this.compile_error(...)`.

lib/runtime/adapter-read.js
davidvgalbraith: Oh yeah! But thinking now the one issue there is it'd break all the other adapters until they implemented allowedOptions. Is that ok?

PR #321, opened 2016-02-03
Consolidate Juttle runtime: Don’t set Juttle.proc

dmajda:
Kill `Juttle.proc` after some preparatory refactoring. See individual commits for details.

Note the change set leaves `lib/runtime/procs/procs.js` in a weird state where it builds the `Juttle` object and has otherwise nothing to do with procs. I plan to dismantle that file completely over time.

Part of work on #97.


dmajda:
@bkutil

bkutil:
+1

PR #323, opened 2016-02-03
update vim syntax file

### LINKED ISSUE #43: move vim plugin to its own repo
From @dmajda on 12/08: 

The list of procs in Vim Juttle mode is out of sync

rlgomes:
fixes #43

with this change I've also included an easy to use `update.sh` script
that can be used to update the proc list with minimal effort and could
be expanded to handle a few other mappings between our pegjs grammar
file and the resulting `juttle.vim` file.

### DIFF
--- a/misc/vim/README.md
+++ b/misc/vim/README.md
@@ -8,6 +8,6 @@ To install automatic syntax highlighting for Juttle programs:
 
 ## Updating the rules
 
-There's a simple ash script in syntax directory that can be called like so
+There's a simple bash script in syntax directory that can be called like so
 `./update.sh` to simply update the syntax.vim based off the
 `lib/parser/parser.pegjs` grammar directly and update the proc rules.
--- a/misc/vim/README.md
+++ b/misc/vim/README.md
@@ -1,8 +1,13 @@
 # Vim plugins for Juttle
 
 ## Vim syntax highlighting
+
 To install automatic syntax highlighting for Juttle programs:
 1. Copy or link the filetype detection script to the ftdetect directory underneath your vim runtime directory (normally $HOME/.vim/ftdetect)
 2. Copy or link syntax/juttle.vim to the syntax directory underneath your vim runtime directory (normally $HOME/.vim/syntax).
 
+## Updating the rules
 
+There's a simple ash script in syntax directory that can be called like so
+`./update.sh` to simply update the syntax.vim based off the
+`lib/parser/parser.pegjs` grammar directly and update the proc rules.
--- a/misc/vim/syntax/juttle.vim.tmpl
+++ b/misc/vim/syntax/juttle.vim.tmpl
@@ -21,7 +21,7 @@ syn keyword juttleConditional       if else
 syn keyword juttleFunctions         function reducer sub
 
 syn keyword juttleSinks             view
-syn keyword juttleProcs             $(grep "'processor' =" ../../../lib/parser/parser.pegjs  | awk '{print $4}' | xargs)
+syn keyword juttleProcs             $(grep "'processor' =" $CWD/../../../lib/parser/parser.pegjs  | awk '{print $4}' | xargs)
 syn keyword juttleConst             false null true
 
 syn region  juttleLineComment       start=+\/\/+ end=/$/ oneline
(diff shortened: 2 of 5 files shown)


rlgomes:
@dmajda you reported the bug (if I understood that it was in fact the vim syntax file not having been updated), @mnibecker you're one of the few who use vim like me.

misc/vim/syntax/juttle.vim.tmpl
```diff
@@ -0,0 +1,47 @@
...
+hi def link juttleStorage               StorageClass
+hi def link juttleDebug                 Debug
+hi def link juttleKeywords              Statement
+hi def link juttleConditional           Conditional
+hi def link juttleFunctions             Function
+hi def link juttleProcs                 Keyword
+hi def link juttleSinks                 Keyword
+hi def link juttleConst                 Constant
+
+let b:current_syntax = "juttle"
```
mnibecker: looks like this is an accidental check in

misc/vim/README.md
```diff
@@ -1,8 +1,13 @@
...
 
 ## Vim syntax highlighting
+
 To install automatic syntax highlighting for Juttle programs:
 1. Copy or link the filetype detection script to the ftdetect directory underneath your vim runtime directory (normally $HOME/.vim/ftdetect)
 2. Copy or link syntax/juttle.vim to the syntax directory underneath your vim runtime directory (normally $HOME/.vim/syntax).
 
+## Updating the rules
 
+There's a simple ash script in syntax directory that can be called like so
```
mnibecker: s/ash/bash

misc/vim/syntax/juttle.vim.tmpl
mnibecker: what's the point of this file?

mnibecker: oh nvm, got it

mnibecker:
not sure if this is the standard method of updating syntax for vim (the whole pegjs parser thing), but it seems like a really error prone process to depend on people making updates in two places. Maybe we just always do this and make building the actually `juttle.vim` file part of the `juttle/juttle` build process.

rlgomes:
@mnibecker let me see how to add this to the gulp build process in a way that we'd rebuild locally so that the developer would check in the new file and also we'd catch in travis if the "newer" version wasn't checked in.

rlgomes:
@juttle/developers misc build task in gulp to take care of updating the vim syntax file but can also be used to update other misc related build steps so we don't forget to do them periodicially.

rlgomes:
bump!

dmajda:
Do we really want to keep editor integrations in the main repo? I’d say no, for several reasons:

  * There are many editors on the world and we obviously can’t support all of them. If we keep editor integration files in the repo, where will we draw the line between what to include and what not to include? What do we say to a third-party contributor who creates a PR implementing support for favorite editor?

  * Editor integration files are a maintenance burden. If I add a new language feature, do I update the files? Or push someone to do that (e.g. if I’m not sure how to do the change)? Should these updates get tracked? Should they block releases?

    We can automate the changes in case of procs, but that’s about it. Anything more complicated still needs manual treatment.

  * My PEG.js experience thought me that editor integrations are a great thing for the community to handle and improve. Again, a separate repo is much better for this (easier to contribute to, easier to pass maintenance to someone, etc.).

  * It is usually easier to install the integration when it lives in a separate repo. For example, I use [Vundle](https://github.com/VundleVim/Vundle.vim) and an external repo would allow me to say just something like `Plugin 'juttle/vim-juttle'` in my `.vimrc` and be done. Right now, one also needs to set a path and the installation will clone the whole Juttle repo!

    I can imagine similar issues exist for other editors.

rlgomes:
@dmajda I agree with you so lets get the vim and emacs plugins pulled out into their own repos like you suggested `vim-juttle` and `emacs-juttle` (@mstemm would take care of the emacs).

mstemm:
I actually had the emacs stuff as a separate repo at one point. No problem with moving it out again.

rlgomes:
ok then closing this and will put up a PR to remove the vim stuff its own PR along with removing it from here and then a simple readme indicating how to install it using pathogen and vundle which are the two popular vim plugin managers.

dmajda:
@rlgomes Cool, thanks!

demmer:
FWIW for the [atom juttle support](https://github.com/juttle/atom-language-juttle) I put it in a separate repo since that seemed to be the precedent for all atom plugins.

PR #326, opened 2016-02-04
convert to es6 classes

### LINKED ISSUE #341: convert classes from extendable-base to es6 classes
Since it's now 2016 we should modernize the Juttle codebase to use the es6 features that are [supported by node.js](https://nodejs.org/en/docs/es6/).

In particular, we should convert all classes from extendable-base to ES6 class syntax.

demmer:
Add a script based on https://github.com/benjamn/recast that converts an `extendable-base` style class definition into the equivalent ES6 class.

This includes removing the unnecessary `require('extendable-base')` and properly adding the superclass constructor invocation to pass along the same set of arguments that the constructor was called with. It also defines getter properties as a stopgap to handle cases in which we assign literals to the prototype since those are not supported in ES6.

The most substantive change required for this was a rework of the implementation of reduce so it no longer relied on a tangled web of mixins. Instead it now has a simple linear inheritance chain. The one distasteful part of this is that the batch reduce (i.e. not `reduce -every`) ends up inheriting the behavior to process the -every options even though it doesn't use it. This seems like the lesser of two evils when compared with the previous mixin approach, but I could go either way. In any event, it still passes all the tests.

Fixes #341

### DIFF
--- a/lib/runtime/procs/read.js
+++ b/lib/runtime/procs/read.js
@@ -7,8 +7,9 @@ var values = require('../values');
 var JuttleMoment = require('../../runtime/types/juttle-moment');
 var Promise = require('bluebird');
 
-var Read = source.extend({
-    initialize: function(options, params, location, program) {
+class Read extends source {
+    constructor(options, params, location, program) {
+        super(options, params, location, program);
         var adapter = adapters.get(params.adapter.name, params.adapter.location);
         if (!adapter.read) {
             throw this.compile_error('JUTTLE-UNSUPPORTED-ADAPTER-MODE', {
@@ -63,15 +64,15 @@ var Read = source.extend({
         });
 
         this.nextTick = this.program.now;
-    },
+    }
 
-    procName: function() {
+    procName() {
         return 'read-' + this.params.adapter.name;
-    },
+    }
 
     // Set the initial time parameters for read from the adapter and
     // schedule the initial run.
-    start: function() {
+    start() {
         // First tell the adapter to kick off the process of reading.
         this.adapter.start();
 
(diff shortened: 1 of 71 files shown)


demmer:
@juttle/developers

lib/runtime/procs/read.js
```diff
@@ -9,10 +9,13 @@ var JuttleMoment = require('../../runtime/types/juttle-moment');
 var Promise = require('bluebird');
 
 
-var Read = source.extend({
-    procName: 'read',
+class Read extends source {
+    get procName() {
```
davidvgalbraith: You'll have to use `this.procName = 'read';` in constructor instead of this as of https://github.com/juttle/juttle/pull/317, since that does `this.procName += ' ' + adapter.name;` once we have an adapter.

dmajda:
Nice! Few notes: 

  * The ugliness in `reduce` is a price I’m willing to pay, the previous mixin approach was nothing nice either. (The resulting inheritance chain should probably be replaced by composition at some point, but this idea requires more thought. Maybe something for @bkutil?)

  * I’m not sure whether it makes sense to add the conversion script into the repo since the conversion is done here and the other repos will presumably be converted soonish. Personally, I’d just put it up as a Gist. But if you find it more useful to have it in the repo, keep it there (for a while).

  * I assume we will get rid of the introduced getters manually over time.

+1

lib/runtime/procs/read.js
demmer: This is a bit tricky since we need the `procName` to be set in the base constructor I actually think it would be a better change if the procName were passed into the proc from the compiler and set in the base proc constructor so we don't need to have all this repetition everywhere.

I'll give this a shot and we can see what we think.

demmer:
Once #334 and #342 land are merged then this can be rebased and merged.

I wasn't planning on including the script in the final commit -- I should have mentioned that in the PR. I moved it to a [gist](https://gist.github.com/demmer/562cfc2343cc9d049b69) already and will remove from the branch once I do the rest of the rebase.

demmer:
@dmajda I found a couple more stragglers that were using prototype properties.

demmer:
bump @dmajda

dmajda:
Sorry, didn’t realize this was blocked on me. Looks good, merging to prevent further bitrot.

PR #327, opened 2016-02-04
Filters simplify field reference representation

### LINKED ISSUE #318: Simplify AST representation of fields
Consider the following Juttles:

```
read elastic host == 'example.com'
```

```
read elastic #host == 'example.com'
```

```
read elastic *'host' == 'example.com'
```

In all of them, the `host` field reference gets translated into the following AST:

```javascript
{
    type: "UnaryExpression",
    operator: "*",
    expression: {
        "type": "StringLiteral",
        "value": "a"
    }
}
```

This complicates processing of static filters (where the dynamic `*'host'` form is not supported) using visitors. Specifically, the `*` operator has to be handled by `visitUnaryExpression`, which is otherwise concerned only with handling `NOT`.

The solution is to translate the first two forms into this AST:

```javascript
{
    type: "Field",
    name: "host"
}
```

This means handling of field references in static filters will be simpler and will be performed by a separate function `visitField` (avoiding cluttering `visitUnaryExpression`).

dmajda:
Until now, field references in expressions like `a < 5` or `#b > 6` were represented as `UnaryExpression` nodes with `*` as an operator and a `StringLiteral` with the field name as an expression. The idea behind this was that they are a syntax sugar for field dereference (the canonical form being `*'a' < 5` or `*'b' > 6`).

The `UnaryExpression` representation proved cumbersome to handle, especially in adapter code dealing with static filter expressions. Therefore, this commit gives field references more first-class status, representing them as `Field` nodes.

All relevant places are updated to handle the new nodes (the code is usually taken from `UnaryExpression` handling).

Note that changes done in `StaticFilterChecker` mean that expressions like `*'a' < 5` are no longer considered valid in static filter expressions. I see this as a fix of an irregularity in the language — the `*`
operator is dynamic by its nature and it shouldn’t be allowed in static filter expressions, even with compile-time constant argument.

Fixes #318.

### DIFF
--- a/lib/compiler/ast-visitor.js
+++ b/lib/compiler/ast-visitor.js
@@ -36,6 +36,7 @@ var NODE_CHILDREN = {
 
     /* Expressions */
     Variable:                 [],
+    Field:                    [],
     ToString:                 ['expression'],
     PropertyAccess:           ['base', 'name'],
     FunctionCall:             ['name', 'arguments'],
--- a/lib/compiler/build.js
+++ b/lib/compiler/build.js
@@ -569,6 +569,8 @@ var Build = CodeGenerator.extend({
                 return this.gen_PropertyAccess(ast, { lhs: true });
             case 'Variable':
                 return ast.uname;
+            case 'Field':
+                return this.gen_Field(ast);
             default:
                 throw new Error('unrecognized expression lhs ' + ast.type);
         }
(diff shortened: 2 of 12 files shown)


dmajda:
@bkutil

bkutil:
+1

Issue #328, opened 2016-02-04
NaN pops up in arity error message


bkutil:
```
function f(a) { 
return a;
}
emit -limit 10 
| put a = f()
```

This correctly fails at compile-time, but the message is:
"Error: function f expects 1 to NaN arguments but was called with 0"
We should say something like "function f expects 1 argument but was called with 0"

dmajda:
I think this should be tested in `test/runtime/specs/juttle-spec/expressions/function-call.spec.md`. I’d just:

  1. Clone the parameters testcases.
  2. Remove optional parameters in one copy and adapt (e.g. remove now irrelevant testcases).
  3. Add `(with optional parameters)` and `(without optional parameters)` suffixes to the descriptions.

PR #329, opened 2016-02-04
semantic.js - fix function arg count

bkutil:
For a function with single required argument, the total number of
arguments would be computed as NaN.

This was caused by the value of the field holding optional number
of arguments being 'undefined' and the addition with an integer
converting the result into NaN.

Fix: #328

### DIFF
--- a/test/runtime/specs/juttle-spec/expressions/function-call.spec.md
+++ b/test/runtime/specs/juttle-spec/expressions/function-call.spec.md
@@ -16,8 +16,8 @@ Works correctly inside a recursive function
 
     { time: "1970-01-01T00:00:00.000Z", result: 120 }
 
-Returns correct result when passed required parameters
-------------------------------------------------------
+Returns correct result when passed required parameters (with optional parameters)
+---------------------------------------------------------------------------------
 
 ### Juttle
 
@@ -31,8 +31,8 @@ Returns correct result when passed required parameters
 
     { time: "1970-01-01T00:00:00.000Z", result: [1, 5] }
 
-Returns correct result when passed required and optional parameters
--------------------------------------------------------------------
+Returns correct result when passed required and optional parameters (with optional parameters)
+----------------------------------------------------------------------------------------------
 
 ### Juttle
 
@@ -46,8 +46,8 @@ Returns correct result when passed required and optional parameters
 
     { time: "1970-01-01T00:00:00.000Z", result: [1, 2] }
 
-Produces an error when passed too few parameters
-------------------------------------------------
+Produces an error when passed too few parameters (with optional parameters)
+---------------------------------------------------------------------------
 
 ### Juttle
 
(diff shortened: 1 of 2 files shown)


bkutil:
cc: @dmajda

dmajda:
I think this should be tested in `test/runtime/specs/juttle-spec/expressions/function-call.spec.md`. I’d just:

  1. Clone the parameters testcases.
  2. Remove optional parameters in one copy and adapt (e.g. remove now irrelevant testcases).
  3. Add `(with optional parameters)` and `(without optional parameters)` suffixes to the descriptions.

dmajda:
The commit message could use some improvement. Without looking at the diff I wasn’t able to get what the change is about. The message mainly needs more context (that the computation is in semantic pass which produces the faulty message described in the bug).

bkutil:
Updated the commit message, simplified the code and added tests. Also caught a small bug in passing array argument to the error message for interpolation instead of a string.

test/runtime/specs/juttle-spec/expressions/function-call.spec.md
```diff
@@ -76,8 +76,8 @@ Produces an error when passed too many parameters
 
   * Error: function f expects 1 to 2 arguments but was called with 3
 
-Returns correct result when passed required parameters (module)
----------------------------------------------------------------
+Returns correct result when passed required parameters (module) (with optional parameters)
```
dmajda: I’d just do `(module, with optional parameters)`. We do it like this elsewhere IIRC.

test/runtime/specs/juttle-spec/expressions/function-call.spec.md
```diff
@@ -151,3 +151,106 @@ Produces an error when passed too many parameters (module)
 ### Errors
 
   * Error: function m.f expects 1 to 2 arguments but was called with 3
+
+Returns correct result when passed required parameters (without optional parameters)
+---------------------------------------------------------------------------------
+
+### Juttle
+
+    function f(a, b) {
```
dmajda: I’d just do `f(a)` in these tests, this would correspond to the existing tests a little bit more.

(We should really have more proper tests, passing 0, 1, and many parameters, but this is something for another day and another PR.)

test/runtime/specs/juttle-spec/expressions/function-call.spec.md
```diff
@@ -151,3 +151,106 @@ Produces an error when passed too many parameters (module)
 ### Errors
 
   * Error: function m.f expects 1 to 2 arguments but was called with 3
+
+Returns correct result when passed required parameters (without optional parameters)
+---------------------------------------------------------------------------------
```
dmajda: Make `-`’s match the length of the title?

(We should probably switch to `##` for JuttleSpec headers globally at some point, managing the `-`’s is pain.)

dmajda:
+1 with the suggested changes

bkutil:
Added/removed the missing underlines and fixed parens in test case titles, functions in new tests accept and return only 1 argument, which required fixing also test messages.

I'll merge on green.

Issue #59, opened 2015-12-26
Resolve how to handle error messages in backends


dmehra:
Backends need to be able to produce custom error messages but at the same time integrate into the overall Juttle error handling & messaging scheme.

dmajda:
I see a couple of ways how to handle this:

  1. **Don’t allow adapters to extend the Juttle message list.** In practice, this means the adapters would use various predefined messages (e.g. `RT-NUMBER-ERROR`) for option handling and other common situations, but for everything else they would use a generic `RT-ADAPTER-ERROR` with a custom message (i.e. the definition in `messages.json` would look like `"RT-ADAPTER-ERROR": "Adapter error: {message}"`).

    The advantage of this approach is simplicity. The disadvantages are scattering the messages around the codebase and different style of error reporting in Juttle core and in adapters.

  2. **Allow adapters to define their own namespaced messages.** For example, `juttle-graphite-adapter` would implicitly define a message namespace with the same name and Juttle would automatically load messages defined in adapter’s `lib/messages.json` file into it. Juttle core messages would live in namespace `juttle`. When creating an error/warning, it would be possible to specify the namespace for message lookup, the default being `juttle`.

    This approach is clean and structured and it ensures errors would be defined and reported the same way across the codebase. On the other hand, it’s not that simple.

  3. **Allow adapters to define their own messages without namespaces.** Same as 2. except that all messages would be in the same namespace and we’d use prefixes. For example, `JUTTLE-INVALID-FILTER` would be a Juttle core error but `JUTTLE-GRAPHITE-ADAPTER-INVALID-FILTER` would be an adapter error.

    This approach is simpler than 2. but isn’t as structured.

All these options assume that we generally don’t want adapters to throw plain JavaScript errors (not descending from `JuttleError`). This is in line with the approach used in Juttle core where plain JavaScript errors generally signal internal errors and bugs while `JuttleError`s are used for “expected” problems.

@demmer @bkutil Thoughts?

demmer:
My instinct is that we should do a combination of (1) and (3).

What I mean by that is that we can support two approaches and leave it up to the adapter implementer how they want to handle their custom errors.

[...]

PR #331, opened 2016-02-04
errors: Remove "RT-" and "JUTTLE-" prefixes from error codes

dmajda:
The prefixes currently serve no purpose and in #59 we decided to go the namespaces route for distinguishing codes from adapters and Juttle itself.


dmajda:
@demmer

demmer:
This looks good, but I think we should also remove `-ERROR` from the codes where it's used while you're at it.

dmajda:
Good point. Will tweak the codes, re-run my `sed` thingy and open a new PR (this one bitrotted).

Issue #330, opened 2016-02-04
incorrect error message coming from reduce


bkutil:
This is originally an issue PROD-8047

> When running

> ```
> read -type 'metric' -from :10m ago: -demo 'srch_cluster' name='responses'
> | reduce -every :2s: avg=avg(value) by pop
> | @timechart
> ```

> we get the warning "Warning: batch dropped out-of-order point(s)" even though there is no batch.

Although the example no longer works, the message is still there and still wrong :)

PR #332, opened 2016-02-04
Fix error message in reducer ooo point drop

bkutil:
When a reducer would process out-of-order points, it drops them and
correctly emits a warning. However, the message is misleading. E.g.:

```
read ... | reduce -every :2s: avg=avg(value) by pop
```

would result in:

"Warning: batch dropped out-of-order point(s)"

although there is no `batch` present. Change the wording to include
the proc name, so the above warning is more precise:

"out-of-order point(s) dropped by reduce"

Fix: #330


bkutil:
cc: @dmajda

dmajda:
+1 with the naming change

bkutil:
Renamed key of the error message and `reducer` parameter to `proc`. Will merge on green.

PR #333, opened 2016-02-04
verify that -from/-to are not equal in read adapters

### LINKED ISSUE #322: Don’t allow -from equal to -to
Currently we require `-from` <= `-to` to avoid results which would always be empty. But the case where `-from` and to `-to` are set to the same value will also always return empty points because `-to` is exclusive.

We should probably change the check to require `-from` < `-to`. At least, I can’t find a good reason why not to.

rlgomes:
fixes #322

make sure to validate the -from/-to options passed to the read adapters
are never equal and if they are throw an appropriate error.

### DIFF
--- a/lib/runtime/procs/base.js
+++ b/lib/runtime/procs/base.js
@@ -40,6 +40,14 @@ var base = Base.extend({
                 option: unknown[0]
             });
         }
+
+        if (this.options.from !== undefined &&
+            this.options.to !== undefined &&
+            this.options.from.eq(this.options.to)) {
+            throw this.compile_error('RT-FROM-TO-EQUAL-ERROR', {
+                proc: this.procName
+            });
+        }
     },
 
     out: function(output_name) {
--- a/test/runtime/adapter-read-timeseries.spec.js
+++ b/test/runtime/adapter-read-timeseries.spec.js
@@ -18,6 +18,19 @@ describe('read testTimeseries', function () {
         });
     });
 
+    it('requires -from and -to to be different', function() {
+        return check_juttle({
+            program: 'read testTimeseries -from :2014-01-01: -to :2014-01-01:'
+        })
+        .then(function(result) {
+            throw new Error('unexpected success');
+        })
+        .catch(function(err) {
+            expect(err.code).equal('RT-FROM-TO-EQUAL-ERROR');
+            expect(err.message).equal('read testTimeseries -from/-to must not be equal');
+        });
+    });
+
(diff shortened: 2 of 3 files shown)


rlgomes:
@juttle/developers

lib/runtime/procs/base.js
```diff
@@ -40,6 +40,15 @@ var base = Base.extend({
                 option: unknown[0]
             });
         }
+
+        if (this.options.from !== undefined &&
+            this.options.to !== undefined &&
+            this.options.from.eq(this.options.to)) {
+            throw this.compile_error('RT-FROM-TO-EQUAL-ERROR', {
+                proc: this.procName,
+                option: unknown[0]
```
bkutil: `option` is a copy-pasta

test/runtime/adapter-read-timeseries.spec.js
```diff
@@ -18,6 +18,19 @@ describe('read testTimeseries', function () {
...
+    it('requires -from and -to to be different', function() {
+        return check_juttle({
+            program: 'read testTimeseries -from :2014-01-01: -to :2014-01-01:'
+        })
+        .then(function(result) {
+            throw new Error('unexpected success');
+        })
+        .catch(function(err) {
+            expect(err.code).equal('RT-FROM-TO-EQUAL-ERROR');
+            expect(err.message).equal('read testTimeseries -from/-to must not be equal');
```
bkutil: You could only test for the code, both is probably a bit too strict/redundant, but OK.

bkutil:
+1

lib/runtime/procs/base.js
rlgomes: copy-pasta the root of all evil

bkutil: :)))

test/runtime/adapter-read-timeseries.spec.js
rlgomes: This was total copy-pasta from the test above as I tend to follow the "behavior" of existing code. I think its ok to verify both.

demmer:
I didn't see this commit when it went in, but looking at the code now I think the check is totally out of place and actually belongs right after the other check for from/to validity here: https://github.com/juttle/juttle/blob/master/lib/runtime/procs/source.js#L57

rlgomes:
@demmer oops I tried to find the right location and obviously missed it... I'll revert this and fix in another PR.

PR #334, opened 2016-02-04
proc names from params

demmer:
Instead of having each proc implementation set a prototype property specifying the `procName`, update the compiler to pass in the name of the proc as part of the params that were passed on instantiation.

This required a minor change to the graph-compiler so that it consistently generated params from an object instead of sometimes manually constructing a string.

Also added a debug log to the proc base class that prints out all the options and params as part of the initialization.

Inspired by the discussion on #326.

### DIFF
--- a/lib/adapters/http_server.js
+++ b/lib/adapters/http_server.js
@@ -9,10 +9,10 @@ var Promise = require('bluebird');
 var parsers = require('./parsers');
 
 var Read = source.extend({
-    procName: 'read-http_server',
-
     initialize: function(options, params, location, program) {
         var allowedOptions = ['port', 'method', 'timeField', 'rootPath'];
+        this.procName = 'read http_server';
+
         this.validate_options(allowedOptions);
         var allowed_methods = [ 'POST', 'PUT' ];
 
--- a/lib/compiler/graph-compiler.js
+++ b/lib/compiler/graph-compiler.js
@@ -178,6 +178,8 @@ var GraphCompiler = CodeGenerator.extend({
     },
     gen_proc: function(ast, procName, params) {
         var pname = ast.pname;
+        params = params || {};
+        params.$procName = procName;
         var soptions = this.gen_options(ast.options,
                                         ast.location,
                                         params);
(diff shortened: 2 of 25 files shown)


demmer:
@dmajda @bkutil what do you think? This seems cleaner and will simplify the logic once we move to ES6 classes.

dmajda:
I don’t see how this is cleaner solution than the existing one. I think the information about proc name belongs to each class and it is not something to be injected from the outside.

I also think proc name should be static (both in the “on the class level” and “unchanging” sense) and it should contain just the proc name, not the adapter name. From that POV, doing `this.procName += ' ' + adapter.name;` in `read` and `write` is wrong.

As far as I can tell, the only reason for tweaking `this.procName` is getting a nice logger name (we do `this.logger_name = 'proc-' + this.procName + job_id;` in `base`). If that’s so, my take is that we actually need a different concept here: a *display name*. This can be expressed as an instance method, which by default would return the static proc name, but `read` and `write` would overwrite it to also append the adapter name. This would be my preferred solution.

demmer:
One of the reasons I was partial to this approach is that we currently have all the proc names specified in the grammar, in the name of the class, in the table in procs.js, and then again as the `procName` string in each class. This change makes things a little more DRY in that we just use the procName string from the grammar for display purposes.

However, I am on board with the idea that procName should be static in the unchanging sense and that the idea of a display name function used in read / write is a cleaner approach over modifying `this.procName` to add the adapter name.

In fact, looking through the code, now that we've converted most of the compile-time checks for sources/sinks to use inheritance checks, all current uses of `procName` are actually only for initializing the logger or as a template parameter to error strings.

So, in the interest of moving forward, I'll nix this for now and change `procName` to be a member function `displayName()` (though my personal taste would be to just call it `name`). That avoids the need for ES6 getters once we convert to classes, which was the main motivation.

demmer:
Closing in favor of #364

dmajda:
BTW, I think the first commit is a worthwhile cleanup and should still go in (either as part of #365 or separately).

PR #336, opened 2016-02-04
eslint: add a few additional rules

demmer:
The code already conformed to all of these rules, so adding them
to prevent regression.


demmer:
@rlgomes @dmajda

rlgomes:
+1 looks good, had to look up `eqeqeq` ;)

PR #338, opened 2016-02-04
verify that -from/-to are not equal in read adapters

### LINKED ISSUE #322: Don’t allow -from equal to -to

rlgomes:
fixes #322

make sure to validate the -from/-to options passed to the read adapters
are never equal and if they are throw an appropriate error.

### DIFF
--- a/test/runtime/adapter-read-timeseries.spec.js
+++ b/test/runtime/adapter-read-timeseries.spec.js
@@ -18,6 +18,19 @@ describe('read testTimeseries', function () {
         });
     });
 
+    it('requires -from to be before -to', function() {
+        return check_juttle({
+            program: 'read testTimeseries -from :2014-01-01: -to :2014-01-01:'
+        })
+        .then(function(result) {
+            throw new Error('unexpected success');
+        })
+        .catch(function(err) {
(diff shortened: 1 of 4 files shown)


rlgomes:
@demmer @bkutil take two

demmer:
This looks fine, but it seems like it'd be cleaner to change the existing `RT-TO-BEFORE-FROM-MOMENT-ERROR` to say "-from must be before -to" and then only having a single check of `if (options.from && options.from.gt(options.to))`.

rlgomes:
@demmer tis done

test/runtime/adapter-read-timeseries.spec.js
```diff
@@ -18,6 +18,19 @@ describe('read testTimeseries', function () {
         });
     });
 
+    it('requires -from to be before -tot', function() {
```
demmer: `-tot`?

demmer:
lgtm except that typo

PR #340, opened 2016-02-04
events: replace backbone.Events with eventemitter3

### LINKED ISSUE #49: Decide which events system to use
@mstemm: We should try to use just a single events system. Currently we're mostly using backbone, but maybe we should use node events instead?

@demmer: The reason we use Backbone.events is that it works in the browser and node.js.

@mstemm: matt says now that we use browserify we can use node events in both as well. I don't have a big opinion either way but the run in both advantage of Backbone.events appears to be gone now.

@mnibecker: eventemitter3 is isomorphic and lets us set context unlike node. lets use that.

(Note: "triaged" label on a "question" issue doesn't mean it's been resolved, only that the discussion is cleared to proceed. Our waffle workflow adds that label.)

demmer:
Go through the handful of places where we were using Backbone.Events
in juttle and replace with eventemitter3.

This required changing all calls from `events.trigger` to `events.emit`.
Note that the procs still contain a method `this.trigger` to send the
event as compared to `this.emit` which emits points.

Also instead of attaching the event emitter functions to the program
object that's shared between the procs, attach a handle on the event
emitter itself, changing all callers from `program.trigger` to
`program.events.emit`.

Note that there are a couple of places where we still use the native
node EventEmitter -- that should be cleaned up separately.

Partially fixes #49

### DIFF
--- a/lib/compiler/index.js
+++ b/lib/compiler/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var events = require('backbone').Events;
+var EventEmitter = require('eventemitter3');
 var Scheduler = require('../runtime/scheduler').Scheduler;
 var Graph = require('../graph');
 var Program = require('../program');
@@ -49,7 +49,7 @@ var stages = {
     eval: function(code, options) {
         var program = new Program();
         program.scheduler = options.scheduler;
-        program.events = _.extend({}, events);
+        program.events = new EventEmitter();
         program.code = code;
 
         program._eval();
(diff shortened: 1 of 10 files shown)


demmer:
One thought is that we could additionally change `proc.trigger` to `proc.emit_event` so we remove all remnants of the backbone event `trigger` legacy. Personally I don't think it's worth the headache but it is a simple search and replace.

demmer:
@juttle/developers

davidvgalbraith:
+1

PR #342, opened 2016-02-05
replace builtin node events with eventemitter3

demmer:
The parser / serializer base classes were the only places where we used the builtin node event emitter, so for consistency, use eventemitter3.

Along the way, clean up the way that we define all the parser / serializer classes to be more consistent and to pave the way for the big ES6 class conversion in #326.

Finishes up #49


demmer:
@juttle/developers

davidvgalbraith:
+1

PR #345, opened 2016-02-05
split up read/write in file adapter tests

rlgomes:
the length of the original file.spec.js was getting a bit out of hand
and this will help with maintaining these tests a bit easier


rlgomes:
@davidvgalbraith @demmer wanted to clean this up separately from the append PR and then I'll follow up with a PR to extend the serializers to have an append option which would require that you provide an input stream along with your output stream for some of the formats (such as csv) which would need to read the first line of the stream before it could figure out exactly what the headers of that csv output was. At which point we can all use that same feature to add the append in the file adapter or s3 adapter like david wats.

davidvgalbraith:
+1

PR #347, opened 2016-02-05
remove time field from data

### LINKED ISSUE #159: read -timeField fieldx should delete fieldx
When the -timeField is used with read, the current behavior converts the specified field into a Juttle time and puts it into a field named `time`, but leaves the original field unmodified in the data points.

For example, given a file containing the following points:

```
[
   {
        "i": 1,
        "my_time": "2015-01-01"
    },
    {
        "i": 2,
        "my_time": "2015-01-02"
    },
    {
        "i": 3,
        "my_time": "2015-01-03"
    }
]
```

Running `read file -file <path> -timeField 'my_time' returns the following:

```
┌────────────────────────────────────┬──────────┬───────────────┐
│ time                               │ i        │ my_time       │
├────────────────────────────────────┼──────────┼───────────────┤
│ 2015-01-01T00:00:00.000Z           │ 1        │ 2015-01-01    │
├────────────────────────────────────┼──────────┼───────────────┤
│ 2015-01-02T00:00:00.000Z           │ 2        │ 2015-01-02    │
├────────────────────────────────────┼──────────┼───────────────┤
│ 2015-01-03T00:00:00.000Z           │ 3        │ 2015-01-03    │
└────────────────────────────────────┴──────────┴───────────────┘
```

It seems better to remove the field `my_time` from the points since the value is now properly converted to a moment in the `time` field.

Besides being somewhat tidier, the current solution will cause problems if we send the data to a timechart, since the series identification logic will interpret the extra field as a series key.

rlgomes:
fixes #159

### DIFF
--- a/lib/runtime/juttle-utils.js
+++ b/lib/runtime/juttle-utils.js
@@ -50,7 +50,12 @@ function parseTime(points, timeField, context) {
         }
 
         var time = pt[timeField || 'time'];
-        if (time) { pt.time = time; }
+        if (time) {
+            pt.time = time;
+            if (timeField !== 'time') {
+                delete pt[timeField];
+            }
+        }
     });
 
     toNative(points);
--- a/test/runtime/juttle-utils.spec.js
+++ b/test/runtime/juttle-utils.spec.js
@@ -62,4 +62,55 @@ describe('juttle utils tests', function() {
             expect(converted.time.unixms()).to.equal(time.getTime());
         });
     });
+
+    describe('parseTime', function() {
+        it('parses a timeField correctly and removes it from resulting data', function() {
+            var points;
+            return new Promise((resolve, reject) => {
+                points = utils.parseTime([
+                    { foo: '2014-01-01T00:00:00.000Z'},
+                    { foo: '2015-01-01T00:00:00.000Z'}
+                ], 'foo', {
+                    trigger: (type, error) => { 
+                        reject(error);
+                    }
+                });
+                resolve();
+            })
+            .then(() => {
+                expect(points).to.deep.equal([
+                    {
+                        time: new JuttleMoment({
+                            rawDate: new Date('2014-01-01T00:00:00.000Z')
+                        })
+                    },
+                    {
+                        time: new JuttleMoment({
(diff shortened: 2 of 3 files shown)


rlgomes:
@juttle/developers

lib/runtime/juttle-utils.js
```diff
@@ -50,7 +50,12 @@ function parseTime(points, timeField, context) {
         }
 
         var time = pt[timeField || 'time'];
-        if (time) { pt.time = time; }
+        if (time) {
+            pt.time = time;
+            if (timeField !== 'time') {
```
mstemm: Not sure this really matters, but this would only actually remove the timeField if it is truthy. So if someone had a -timeField with values that were undefined, the points with undefined timeFields would remain, even though the time  values would be undefined.

test/runtime/juttle-utils.spec.js
```diff
@@ -62,4 +62,55 @@ describe('juttle utils tests', function() {
...
+        });
+
+        it('warns if any points is missing the timeField', function() {
+            var points;
+            return new Promise((resolve) => {
+                points = utils.parseTime([{ foo: 1 }, {},  { foo: 3 }], 'foo', {
+                    trigger: function(type, error) { 
+                        expect(type).to.equal('warning');
+                        expect(error.toString()).to.match(/point is missing a time in foo/);
+                        resolve();
```
mstemm: This would be a good place to test the case of foo being undefined.

rlgomes: We're already doing the `.deep.equal` below that verifies the data only contains the time field and nothing else so that would cover the foo not being there.

lib/runtime/juttle-utils.js
rlgomes: The `if (time)` is from the existing code and I merely am checking that the `timeField` value is not `time` before I delete it from the data. That check on time should be something to fix in another PR where we go through all the code and decide how do we want to check for time not being set which I think @demmer and @dmajda already have some ideas on this matter.

rlgomes:
bump!

davidvgalbraith:
+1

PR #352, opened 2016-02-06
Adding a doc page on debugging

### LINKED ISSUE #307: Add a doc on debugging Juttle
To help juttle developers debug their programs, let's document the tips and tricks.

dmehra:
resolves #307 
I've been meaning to do this for a while, here goes. See what you think @juttle/developers 
If I forgot (or never knew) other useful means of debugging, let's add them.


dmehra:
Note: the screenshot from the visualizer needs to be updated after @go-oleg merges in current juttle, because as is it's showing a bug that was fixed in juttle but not yet in the visualizer.

docs/concepts/debugging.md
```diff
@@ -0,0 +1,229 @@
...
+DEBUG=* juttle <program-file.juttle>
+```
+
+or
+
+```
+DEBUG=* juttle -e "juttle-program"
+```
+
+To run programs in the browser environment with debug logging, set `--log-config <log4js-config-path>` and specify desired log level in the log4js config, as described in the [outrigger README](https://github.com/juttle/outrigger).
```
mstemm: This will be juttle-engine, so might as well update it now.

mstemm:
I think adding a section about adapters and backends would be useful. That's certainly been a common situation to debug. I think the section would talk about verifying the connection to the backend from the location where the program is running, translating the query done by the read command into a native backend query and doing it by hand to see if the results are the ones you expect, and possibly using a network capture tool to see what's going over the wire (when the data from juttle program-backend isn't encrypted).

I also think a juttle-engine debugging doc would be useful as well, talking about running the program in juttle to get better highlighting of the location where the error occurred, possibly using a network capture tool to monitor the communications between browser and juttle-engine, etc. That could just be a pending issue for now as we'll probably be improving the debugging quite a bit.

go-oleg:
@dmehra i updated [juttle-flowgraph-viz](https://github.com/juttle/juttle-flowgraph-viz) to the latest juttle.

docs/concepts/debugging.md
dmehra: Ok, will this be the README with log config? It's currently empty. https://github.com/juttle/juttle-engine

dmehra:
@mstemm please check my usage of juttle-service/engine/viewer in the update.
@go-oleg, thanks, got the new screenshot from visualizer in here.
I'd like to merge this, and have future updates come as part of debug improvements in https://github.com/juttle/juttle-viewer/issues/7 and https://github.com/juttle/juttle-viewer/issues/9.

mstemm:
It's more likely that you'll be debugging with juttle-engine, which includes the browser views via the app, so I think all references to juttle-service should be juttle-engine.

mstemm:
Also, I'm not sure we actually want to call out juttle-viewer in these types of walkthough docs. They'll really be using juttle-engine, which contains juttle-viewer.

bkutil:
Certain adapters support `-debug` option (maybe it should be supported by all of them), which prints out the raw query to be executed. Should we mention this in the docs?

demmer:
While -debug is useful for simple queries, as adapters get more complicated
it is often the case that one `read` command in juttle actually produces
multiple queries to the back end so this isn't so straightforward.

However I do think that all the adapters should print the raw queries as
part of their internal debug logging, so once we implement
https://github.com/juttle/juttle/issues/373 and
https://github.com/juttle/juttle-viewer/issues/7 then we should update
these instructions to tell users to turn on debug logging for a given
adapter to see the raw queries (and responses?).

-m


On Wed, Feb 10, 2016 at 5:26 AM, Balázs Kutil <notifications@github.com>
wrote:

> Certain adapters support -debug option (maybe it should be supported by
> all of them), which prints out the raw query to be executed. Should we
> mention this in the docs?
>
> —
> <https://github.com/juttle/juttle/pull/352#issuecomment-182372021>.
>



-- 
Michael Demmer
Jut - CTO / VP Engineering
demmer@jut.io

PR #353, opened 2016-02-07
adapters: check required options in base

davidvgalbraith:
Also, we don't currently check allowed options for adapter write unless the `allowedOptions` method is overwritten, which is inconsistent with what we do for `read`. So let's check them.

@juttle/developers


mstemm:
lgtm.

PR #354, opened 2016-02-08
Consolidate Juttle runtime: Move Juttle.extend_options to the compiler

dmajda:
`Juttle.extend_options` was a helper function responsible for processing proc options — specifically expanding nested options and `-o`/`-options`. So far it did its work in runtime but there is no good reason for that (options are known in compile-time already).

This PR moves `Juttle.extend_options` into `GraphCompiler`, which runs it in compile-time. Note the move required making it work on AST nodes instead of Juttle values.

Part of work on #97.

### DIFF
--- a/lib/compiler/graph-compiler.js
+++ b/lib/compiler/graph-compiler.js
@@ -192,23 +192,64 @@ var GraphCompiler = CodeGenerator.extend({
         this.emit(code);
         return pname;
     },
+    extend_options: function(opts, location) {
+        var options = { type: 'ObjectLiteral', properties: [], location: location };
+        function setopt(name, val, location) {
+            var parts = name.split('.');
+            var o = options;
+            for (var i=0; i<parts.length-1; i++) {
+                var part = parts[i];
+                var index = _.findIndex(o.properties, function(property) {
+                    return property.key.value === part;
+                });
+                if (index === -1) {
+                    o.properties.push({
+                        type: 'ObjectProperty',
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil or @demmer

demmer:
Based on a scan through the code it doesn't seem like this is tested anywhere... at least I don't see any references to RT-BAD-NESTED-OPTION or RT-INVALID-SINK-OPTIONS-ERROR or their actual string contents.

The code looks like it's correct, but I'd like to see some basic tests of the functionality before merge.

dmajda:
Some tests are in [`test/runtime/object-literals.spec.js`](https://github.com/juttle/juttle/blob/655f4ddf056b81334272e6123321f74956229cd0/test/runtime/object-literals.spec.js), the `RT-INVALID-SINK-OPTIONS-ERROR` error is tested there (at the very end). I’ll add a test for the other error there.

dmajda:
Actually, `RT-BAD-NESTED-OPTION` is [exercised in `test/runtime/sinks.spec.js`](https://github.com/juttle/juttle/blob/655f4ddf056b81334272e6123321f74956229cd0/test/runtime/sinks.spec.js#L217-L248). So I didn’t add anything, just rebased to resolve a conflict.

@demmer Good to merge?

demmer:
Oh ok then. Apparently I can't search. +1

dmajda:
It’s more like I was *certain* I saw these tests in the past, so I digged.

PR #355, opened 2016-02-08
Consolidate Juttle runtime: Kill Juttle.visitGen and Juttle.teardown

dmajda:
Inline `Juttle.teardown` into `Program#deactivate`. Move `Juttle.visitGen` into `Program#visitGen`.

Part of work on #97.


dmajda:
@bkutil or @demmer

demmer:
lgtm

PR #356, opened 2016-02-08
errors: Simplify error codes

dmajda:
Remove `RT-` and `JUTTLE-` prefixes from error codes. Also remove the `-ERROR` suffix in cases where it was redundant.

The removed prefixes currently serve no purpose and in #59 we decided to go the namespaces route for distinguishing codes from adapters and Juttle itself.

This is an improved version of #331. Part of work on #165.


dmajda:
@demmer 

I removed the `-ERROR` suffix where the message still made sense without it (e.g. `REQUIRED-OPTION`, `INVALID-JSON`) but I kept it where the code would be incomplete or sound odd without it (e.g. `TYPE-ERROR`, `JOIN-INPUTNUM-ERROR`). I think there is quite some room for improvement and unification in naming the codes (which would get rid of some more `-ERROR` suffixes) and also in the messages themselves, but I didn’t want to get that deep in this PR in order to keep it mostly mechanical.

If the PR looks good to you, please merge it right away to prevent bitrot. Thanks.

dmajda:
Rebased one more time.

@demmer

demmer:
lgtm +1

PR #362, opened 2016-02-08
http server adapter port to the new adapter api

demmer:
Port the http_server adapter to the new adapter API.

In addition to changes to the adapter itself, this also exposed some issues and limitations in the base class:
- Add wrappers for compileError and runtimeError to the adapter read/write base classes. This is not strictly required for the port, but rather than change the adapter to explicitly call errors.compileError, I figured we would need these for per-adapter error codes as described in #59.
- Add a teardown hook for read so the adapter can clean up properly.
- Add simple eof tracking to read. This avoids sending multiple eof's downstream and also as a workaround for the limitation that we can't currently cancel pending scheduler callbacks.

### DIFF
--- a/lib/adapters/http_server.js
+++ b/lib/adapters/http_server.js
@@ -3,21 +3,18 @@
 var http = require('http');
 var _ = require('underscore');
 var contentType = require('content-type');
-
-var source = require('../runtime/procs/source');
+var AdapterRead = require('../runtime/adapter-read');
 var Promise = require('bluebird');
 var parsers = require('./parsers');
 
-var Read = source.extend({
-    procName: 'read-http_server',
+class ReadHTTPServer extends AdapterRead {
+    constructor(options, params) {
+        super(options, params);
 
-    initialize: function(options, params, location, program) {
-        var allowedOptions = ['port', 'method', 'timeField', 'rootPath'];
-        this.validate_options(allowedOptions);
         var allowed_methods = [ 'POST', 'PUT' ];
 
         if (options.method && allowed_methods.indexOf(options.method) === -1) {
-            throw this.compile_error('RT-INVALID-OPTION-VALUE', {
+            throw this.compileError('RT-INVALID-OPTION-VALUE', {
                 option: 'method',
                 supported: allowed_methods.join(', ')
             });
@@ -27,19 +24,24 @@ var Read = source.extend({
         this.method = options.method || 'POST';
         this.timeField = options.timeField;
         this.rootPath = options.rootPath;
+        this.readEvery = options.every ? options.every.milliseconds() : 50;
 
         if (params.filter_ast) {
-            throw this.compile_error('RT-ADAPTER-UNSUPPORTED-FILTER', {
+            throw this.compileError('RT-ADAPTER-UNSUPPORTED-FILTER', {
                 proc: 'read http_server',
                 filter: 'filtering'
             });
         }
-    },
 
-    start: function() {
-        var self = this;
+        this.points = [];
+    }
 
-        this.server = http.createServer(function(req, res) {
+    static allowedOptions() {
+        return ['port', 'method', 'timeField', 'rootPath', 'every'];
+    }
+
+    start() {
+        this.server = http.createServer((req, res) => {
             var format;
             try {
                 format = contentType.parse(req).type.split('/')[1];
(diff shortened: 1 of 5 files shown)


demmer:
@mnibecker @davidvgalbraith @rlgomes

davidvgalbraith:
What does `read -every` mean for the http_server? I thought it was a true live source.

demmer:
@davidvgalbraith good question... in order to match the contract expected by the base class (where it periodically calls `read` to get new points), with the fact that the adapter basically spends all its time waiting for points to arrive over the wire, in this implementation  each call to `read` adds a timeout handler and holds the promise open to see if new points arrive.

This timeout is set by default to 50ms but can be overridden by the `-every` option. I'm open to making this a different option, but it seemed the most logical.

davidvgalbraith:
Seems complicated. Why don't we just return no points if we don't have any points, and then the next time the adapter calls us we'll give some points if we've gotten any?

demmer:
@davidvgalbraith Yeah... I guess in that case it'd be up to the base class to impose the delay. I can see doing it that way.

lib/adapters/http_server.js
```diff
@@ -48,40 +50,37 @@ var Read = source.extend({
...
 
-                    self.parseTime(points, self.timeField);
-                    emit_points = emit_points.concat(points);
+                    this.parseTime(points, this.timeField);
+                    this.points = this.points.concat(points);
                 })
-                .then(function() {
-                    self.emit(emit_points);
-
+                .then(() => {
```
davidvgalbraith: Let's flatten the chain here.

lib/adapters/http_server.js
```diff
@@ -103,21 +137,25 @@ var Read = source.extend({
...
-        if (this.server) {
-            this.server.close();
-        }
+    }
 
-        this.emit_eof();
+    teardown() {
+        return new Promise((resolve, reject) => {
+            if (this.server) {
+                this.server.close(() => {
```
davidvgalbraith: `this.server.close(resolve)` is the obnoxiously pithy way to express this, if you want.

lib/runtime/procs/read.js
```diff
@@ -222,6 +223,7 @@ var Read = source.extend({
             this.program.scheduler.schedule(nextRead.unixms(), _.bind(this.run, this));
         })
         .catch((err) => {
+            this.reading = null;
```
davidvgalbraith: Can we `.finally`

test/adapters/http_server.spec.js
```diff
@@ -81,6 +81,41 @@ describe('read http_server', function() {
...
+
+        return compile_juttle({
+            program: 'read http_server -every :100ms: -port ' + port
+        })
+        .then(function(prog) {
+            program = prog;
+            programFinish = run_juttle(program);
+        })
+        // Wait a while before posting the data
+        .delay(500)
```
davidvgalbraith: These delays are a little rickety, can we use `bluebird-retry`?

lib/adapters/http_server.js
```diff
@@ -89,9 +88,44 @@ var Read = source.extend({
...
+            .catch(Promise.TimeoutError, (err) => {
+                this.logger.debug('read: no points arrived in', this.readEvery, 'ms');
+                return {};
+            });
+        }
+    }
+
+    readResult(limit) {
+        var ret = {};
+        if (this.points.length <= limit) {
```
davidvgalbraith: The `else` of this `if` will have the same effect as the `if` if `this.points.length <= limit`, so we only need the `ret.points = this.points.splice(0, limit);` line.

lib/adapters/http_server.js
demmer: I didn't actually change this part of the code substantively, but yeah I can do that and I agree it would clean things up.

lib/adapters/http_server.js
demmer: Except that I don't want to pass through any arguments that come back in the callback as my resolution of teardown.

Not that it really matters but I think the less-pithy two extra lines of code improve clarity.

lib/runtime/procs/read.js
demmer: We could -- one concern would be whether the scheduler could end up calling us back before the .finally handler executed. It'll probably work fine, but I'd rather keep things as-is with the duplication rather than rely on that timing to all line up properly.

lib/adapters/http_server.js
demmer: Sure, but wouldn't that unnecessarily create a whole new array?

test/adapters/http_server.spec.js
demmer: How? The idea here is to add some delay during which the adapter is idle, then send the post, then give the events some time to propagate, then get the response. I don't see a cleaner solution using retry, but if you have something in mind then I'd be open to it.

demmer:
@davidvgalbraith How does the read path look now with ee711a8?

lib/adapters/http_server.js
davidvgalbraith: Computers are fast! It wouldn't reallocate all the objects so we're talking about like 10 microseconds here.

demmer: Meh. Not a big deal one way or the other -- I figure the earlier way was cleaner, but I've pushed up a commit that always does the splice whether necessary or not.

davidvgalbraith:
+1

PR #364, opened 2016-02-09
proc name function

demmer:
Change procName from a prototype property on the procs into a function. For most procs it just returns the name of the proc itself, but for read and write it includes the name of the adapter as well.

This was proposed in #334 -- I decided to just stick with `procName()` as the function instead of `displayName()` since in writing it, it just seemed like `display` felt out of place. I don't feel _too_ strongly about it though.

FWIW I tried to just use `name()` but the test views use a `name` property already and I didn't feel like changing that all as well.

As part of this change, updated a few error message strings so they consistently use `proc` as the template parameter.

### DIFF
--- a/lib/adapters/stdio/read.js
+++ b/lib/adapters/stdio/read.js
@@ -10,8 +10,6 @@ class ReadStdio extends AdapterRead {
     constructor(options, params) {
         super(options, params);
 
-        this.procName = 'read stdio';
-
         this.timeField = options.timeField;
 
         this.format = options.format ? options.format : 'json'; // default to json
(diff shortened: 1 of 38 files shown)


demmer:
@dmajda let me know what you think of this one. The relevant commits start at 7bac9c0.

dmajda:
+1

I still think the concepts of proc name and what is essentially a logger name are distinct (and that the former should not include the adapter name) but when viewed purely as a step in the ES6 transition the change makes sense. We can refactor later if we’ll end up needing to distinguish the concepts in practice.

PR #154, opened 2016-01-12
Remove unused proc initializer params

dmajda:
Remove `juttle` and `pname` parameters from proc initializers.

Fixes #132.

PR #365, opened 2016-02-09
pass pname in proc params and attach params to proc instance

go-oleg:
Pretty much just mooching off of @demmer's work in https://github.com/juttle/juttle/pull/334 and the to-be-reviewed https://github.com/juttle/juttle/pull/364.

Previously, `pname` was attached to proc instances at `this.pname`, but that was removed in https://github.com/juttle/juttle/pull/154.

I was using the proc's `pname` in [juttle-flowgraph-viz](https://github.com/juttle/juttle-flowgraph-viz) when monkey-patching [emit](https://github.com/juttle/juttle-flowgraph-viz/blob/848f5333611165568c36d5d9d1d8d3f33453d553/src/app.js#L16-L25) to do a "join" between the node emitting the points and its entry in the flowgraph JSON.

@dmajda cc @demmer

### DIFF
--- a/lib/compiler/graph-compiler.js
+++ b/lib/compiler/graph-compiler.js
@@ -212,26 +212,21 @@ var GraphCompiler = CodeGenerator.extend({
         return code;
     },
     gen_params: function(params) {
-        var code = '';
+        var code = '{';
         var comma = '';
-        if (typeof params === 'string') {
-            code += params;
-        } else {
-            code += '{';
-            _.each(params, function(val, name) {
-                var e;
-                code += comma;
-                if (name[0] !== '$') {
-                    e = val;
-                } else {
-                    e = JSON.stringify(val);
-                    name = name.substring(1);
-                }
-                code+= name + ':' + e;
-                comma = ',\n ';
-            } );
-            code += '}';
-        }
+        _.each(params, function(val, name) {
+            var e;
+            code += comma;
+            if (name[0] !== '$') {
+                e = val;
+            } else {
+                e = JSON.stringify(val);
+                name = name.substring(1);
+            }
+            code+= name + ':' + e;
+            comma = ',\n ';
+        } );
+        code += '}';
         return code;
     },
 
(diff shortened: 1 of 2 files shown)


dmajda:
@go-oleg Do I understand correctly that you just need a unique identifier for each node? If so, I’d set it directly on the instances and call it `id` in the runtime. That is, I’d do `params.$id = pname;` in `gen_proc` and then `this.id = params.id;` in `base`.

The idea is to decouple the notion of a unique node identifier from the concept of `pname`, which I consider an internal compiler-level concept which shouldn’t leak out and be relied upon.

go-oleg:
@dmajda it needs to be both a unique identifier and something I can use to lookup the node in the flowgraph stage output (where its referred to as `pname`).

go-oleg:
If I'm using incorrect/unclear terminology w.r.t. "flowgraph stage output", what I mean is 

```
compiler.compile(juttleSource, { stage: 'flowgraph'})
```
or
```
./bin/juttle -e juttleSource --mode flowgraph`
```


For
```
emit -limit 1 | put v = 1 | view table
```
its
```json
{
 "built_graph": {
  "nodes": [
   {
    "type": "BuiltinProc",
    "location": {
     "filename": "main",
     "start": {
      "offset": 0,
      "line": 1,
      "column": 1
     },
     "end": {
      "offset": 14,
      "line": 1,
      "column": 15
     }
    },
    "name": "emit",
    "options": [
     {
      "id": "limit",
      "val": {
       "type": "NumericLiteral",
       "value": 1,
       "d": true
      },
      "location": {
       "filename": "main",
       "start": {
        "offset": 4,
        "line": 1,
        "column": 5
       },
       "end": {
        "offset": 13,
        "line": 1,
        "column": 14
       }
      }
     }
    ],
...
```

dmajda:
I see. OK, let’s keep the concept of `pname`s for the time being.

+1

PR #366, opened 2016-02-09
Use chai 3.5.0

dmajda:
I wanted to use [`.keys`](http://chaijs.com/api/bdd/#keys) in a test and found I couldn’t because we use an awfully old version of `chai`.


dmajda:
@rlgomes

demmer:
lgtm

Issue #369, opened 2016-02-09
port all adapters to juttle 0.5.0


demmer:
#367 introduced the 0.5.0 adapter API as a cleaner way of exposing the internals to external adapter implementations.

This is a tracking issue to make sure that all adapters are ported over to the new API. As part of this work, we should remove the hacky support in read/write for the "old" adapter model in which adapters directly implemented source/sink procs.

The adapters to be ported include:

- [ ] juttle-elastic-adapter
- [ ] juttle-influx-adapter
- [ ] juttle-postgres-adapter
- [ ] juttle-mysql-adapter
- [ ] juttle-sqlite-adapter
- [ ] juttle-opentdsb-adapter
- [ ] juttle-graphite-adapter
- [ ] juttle-gmail-adapter
- [ ] juttle-twitter-adapter
- [ ] juttle-aws-adapter
- [ ] juttle-cloudwatch-adapter

For all external adapters, we should also remove the juttle peerDependency since it is no longer useful for indicating version compatibility and replace with the adapterAPIVersion entry in the package.json.

demmer:
Done as part of https://github.com/juttle/juttle-engine/issues/60

PR #370, opened 2016-02-09
adapters: remove the read/write shims for legacy adapters

demmer:
All adapters should already be converted or should be converted
soon to use the new adapter API, so remove the hacks in read / write
that supported the "old" contract in which adapters directly
implement source / sink procs.

Connects to #369

### DIFF
--- a/lib/runtime/procs/read.js
+++ b/lib/runtime/procs/read.js
@@ -41,29 +41,6 @@ var Read = source.extend({
             });
         }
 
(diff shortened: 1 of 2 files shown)


demmer:
@juttle/developers at this point I think we should just remove the shims that I added in there for legacy adapters that implement read / write as procs.

davidvgalbraith:
+1

PR #374, opened 2016-02-09
adapter getter cleanup

demmer:
A couple of minor cleanups to avoid the use of ES6 getter functions.


demmer:
@davidvgalbraith

davidvgalbraith:
+1

PR #375, opened 2016-02-09
no more prototype properties

demmer:
Clean up various places in the code we use properties attached to the prototype or mixins of functions directly attached to the prototype instead of regular class inheritance.

This is to pave the way for ES6 classes (#326) without requiring the use of getter functions.

@dmajda


dmajda:
+1

Issue #308, opened 2016-02-01
Make -every and possibly -lag configurable per-adapter


mstemm:
While working through upgrading the aws/cloudwatch adapters to use the new adapter read api, I found that the default -every value of 1 second was a poor fit for the cloudwatch adapter. CloudWatch metrics are delayed by 5 minutes, so with the default value of -every of :1s:, nothing ends up being read at all. Cloudwatch reads before the (delayed) metrics are available, and advances time to beyond when they are eventually created.

To address this, it would be useful if you could specify default values for -every (and while we're in there, possibly -lag as well) at the adapter level. This would probably be easiest done by renaming defaultTimeRange() to defaultTimeOptions() and also allowing specifying every and lag along with from and to.

PR #376, opened 2016-02-10
adapters: specify default lag/every

davidvgalbraith:
defaultTimeRange becomes defaultTimeOptions, which includes
lag and every

fixes https://github.com/juttle/juttle/issues/308

### DIFF
--- a/lib/runtime/procs/read.js
+++ b/lib/runtime/procs/read.js
@@ -9,22 +9,6 @@ var Promise = require('bluebird');
 
 var Read = source.extend({
     initialize: function(options, params, location, program) {
-        this.lag = _.has(options, 'lag') ? options.lag : JuttleMoment.duration(0, 's');
-        if (! this.lag.duration) {
-            throw this.compile_error('DURATION-ERROR', {
-                option: 'lag',
-                value: values.inspect(options.lag)
-            });
-        }
-
-        this.readEvery = options.every || JuttleMoment.duration(1, 's');
-        if (! this.readEvery.duration) {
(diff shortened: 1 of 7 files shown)


davidvgalbraith:
@juttle/developers

lib/runtime/procs/read.js
```diff
@@ -50,9 +34,25 @@ var Read = source.extend({
...
-        var defaultTimeRange = this.adapter.defaultTimeRange();
-        this.from = this.options.from || defaultTimeRange.from;
-        this.to = this.options.to || defaultTimeRange.to;
+        var defaultTimeOptions = this.adapter.defaultTimeOptions();
+        this.from = this.options.from || defaultTimeOptions.from;
+        this.to = this.options.to || defaultTimeOptions.to;
+        this.lag = options.lag || defaultTimeOptions.lag || JuttleMoment.duration(0, 's');
+        this.readEvery = options.every || defaultTimeOptions.every || JuttleMoment.duration(1, 's');
+
+        if (! this.lag.duration) {
```
demmer: I think these type checks should go into `handleTimeOptions()` so that they execute before initializing the adapter. That way if the adapter for some reason wants to verify these options they can be sure it's the right type.

demmer:
Other than that one change, this looks good to me.

lib/runtime/procs/read.js
davidvgalbraith: That would be nice! The thing is `defaultTimeOptions` is an instance method, so we can't call it before we initialize the adapter.

PR #380, opened 2016-02-10
logging: configure log targets through DEBUG environment variable

### LINKED ISSUE #373: control debug logging with DEBUG environment variable
To make it easier to turn on and off individual debugging targets, we should support an additional way of controlling log levels by parsing the DEBUG environment variable similar to how the `debug` package on NPM does it.

demmer:
If the DEBUG environment variable is present, support setting the
named targets to debug level, or * for all targets.

Fixes #373.

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -1,17 +1,10 @@
 #!/usr/bin/env node
 'use strict';
-
 /* eslint no-console: 0 */
-var JuttleLogger = require('../lib/logger');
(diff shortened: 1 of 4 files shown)


demmer:
@mstemm Similar logic as https://github.com/juttle/juttle-service/pull/26 (though with tests this time).

mstemm:
lgtm.

PR #381, opened 2016-02-10
append mode for serializers

### LINKED ISSUE #346: add append feature to serializers
So we fix up #286 and support appending from other adapters that want to use the serializers this way.

rlgomes:
fixes #346 

adding append feature into the serializers themselves.

### DIFF
--- a/lib/adapters/serializers/csv.js
+++ b/lib/adapters/serializers/csv.js
@@ -5,50 +5,100 @@ var base = require('./base');
 var csv = require('csv-write-stream');
 var errors = require('../../errors');
 var values = require('../../runtime/values');
+var readline = require('readline');
 
 var CSVSerializer = base.extend({
 
+    initialize: function(stream, options) {
+        options = options  || {};
+
+        if (options.append && !options.input) {
+            throw errors.runtimeError('INTERNAL-ERROR', {
+                error: 'input stream must be provided when appending'
+            });
+        }
+
+        this.append = options.append;
+        this.input = options.input;
+        if (this.append) {
+            this.base = new Promise((resolve, reject) => {
+                var liner = readline.createInterface({
+                    input: this.input,
+                    terminal: false
+                });
+
+                liner.on('line', (line) => {
+                    // there will be a few lines process before the
+                    // readline instance notices the stream has been closed
+                    if (!this.headers) {
+                        this.sendHeaders = false;
+                        this.headers = line.split(',');
+                        this.input.close();
+                        liner.close();
+                    }
+                });
+
+                liner.on('close', () => {
+                    resolve();
+                });
+
+                this.input.on('error', (err) => {
+                    reject(err);
+                });
+            });
+        } else {
+            this.base = Promise.resolve();
+        }
+    },
+
     write: function(points) {
-        var self = this;
+        this.base = this.base
+        .then(() => {
+            _.each(points, (point) => {
+                var keys = _.keys(point);
 
-        _.each(points, function(point) {
-            var keys = _.keys(point);
+                _.each(point, (value, key) => {
+                    if (values.isArray(value) || values.isObject(value)) {
+                        point[key] = JSON.stringify(value);
+                    }
+                });
 
-            _.each(point, function(value, key) {
-                if (values.isArray(value) || values.isObject(value)) {
-                    point[key] = JSON.stringify(value);
+                if (!this.headers) {
+                    this.headers = keys;
+                }
+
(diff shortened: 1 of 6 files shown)


rlgomes:
@juttle/developers this is the first step to have append feature for file (and other adapters that use the `adapter/serializers`) I even added the last commit in this PR for adding this feature to the file adapter but will move that to a follow up PR just to show how this works. Currently you can append `jsonl` and `csv` and all others will throw an exception.

rlgomes:
@davidvgalbraith @demmer you guys would be the best ones to review this.

lib/adapters/serializers/csv.js
```diff
@@ -5,50 +5,107 @@ var base = require('./base');
...
+        }
+
+        this.append = options.append;
+        this.input = options.input;
+        this.base = Promise.resolve();
+    },
+
     write: function(points) {
-        var self = this;
+        if (this.append && !this.readHeaders) {
```
davidvgalbraith: Can we set `this.base` to the headers-reading promise in the constructor if `this.append` to avoid the slightly awkward `readHeaders` thing?

lib/adapters/serializers/csv.js
```diff
@@ -5,50 +5,107 @@ var base = require('./base');
...
-                    point[key] = JSON.stringify(value);
-                }
+                    liner.on('line', (line) => {
+                        // there will be a few lines process before the
+                        // readline instance notices the stream has been closed
+                        if (!this.headers) {
+                            this.sendHeaders = false;
+                            this.headers = line.split(',');
+                            this.input.close();
+                            resolve();
```
davidvgalbraith: Can we just `resolve()` in `liner.on('end')`? Then we'd know the input is cleaned up before we move on, I think.

test/adapters/file/write.spec.js
```diff
@@ -102,7 +102,8 @@ describe('write file adapter tests', function () {
                 return run_read_file_juttle(tmp_file, { format: format });
             })
             .then(function(result) {
-                expect(result.errors).deep.equals([]);
+                expect(result.errors.length).to.equal(0);
```
davidvgalbraith: `expect(result.errors).deep.equals([]);` is actually superior because if it fails it fails with the message `Expected ["some actual error message"] to deep equal []` as opposed to the useless `Expected 1 to equal 0`.

test/adapters/file/write.spec.js
```diff
@@ -111,4 +112,32 @@ describe('write file adapter tests', function () {
         });
     });
 
+    _.each(['csv', 'jsonl'], function(format) {
+        it('can write a point and read it back with ' + format + ' format', function() {
+            return check_juttle({
+                program: 'emit -limit 1 ' +
```
davidvgalbraith: Multi-line template strings are the future! See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings.

lib/adapters/serializers/csv.js
rlgomes: From experimentation liner doesn't seem to ever hit `'end'` when you close the input like I did her to stop reading. So I'm not sure what else could be done to make things shutdown nicely and open to suggestions ?

rlgomes: Just figured out this is a bug becuase liner doens't emit 'end' its 'close' (:sadface:)

test/adapters/serializers/csv.spec.js
```diff
@@ -79,4 +79,49 @@ describe('serializers/csv', function() {
         });
     });
 
+    it('fails when you attempt to append without providing an input stream', function() {
+        expect(function() { 
+            serializers.getSerializer('csv', undefined, { append: true });
+        }).to.throw('internal error input stream must be provided when appending');
+    });
+     
+    it('can append to an existin stream', function(done) {
```
davidvgalbraith: existing

rlgomes:
@davidvgalbraith noicer ?

rlgomes:
And I'll move that last commit to its own PR once this is ready to merge and merge without that last commit...

davidvgalbraith:
Did we already do jsonl? Don't see it in the diff.

rlgomes:
@davidvgalbraith it just "works" as long as you provide a stream that is at the point of appending (which in the case of a file means opening it in append mode). I've added the missing test to the jsonl serializer tests to validate this does in fact work.

davidvgalbraith:
+1

PR #382, opened 2016-02-11
Remove emacs editor hook.

mstemm:
Now available at https://github.com/juttle/emacs-juttle, which I'll
make public after this PR is merged.

@juttle/developers


dmajda:
+1

Issue #320, opened 2016-02-03
Simplify filter handling in adapters


dmajda:
Right now, adapters which translate static filter expressions into native queries have to walk a non-trivial AST, typically using `ASTVisitor`. This is not a great API, for several reasons:

  * Filter AST is complex and it requires handling of many cases, some of them irrelevant for the adapters (e.g. `FilterLiteral`).
  * Unsupported features have to be explicitly handled (typically by `visitXxx` functions throwing exceptions). This is tedious and prone to errors and divergence across adapters.
  * Various levels of filter expressions are handled at the same place. For example, `visitBinaryExpression` needs to handle expression-level operators `AND` and `OR` as well as term-level comparison operators like `<`.

The goal of this issue is to solve these problems by providing adapters a nicer filter processing API.

## Proposal

The key insight is that filter expressions coming into the adapter consist of these concepts, building on top of each other:

  * *field references*
  * *values*
  * *expression terms* which compare field references and values using comparison operators
  * *full-text terms*
  * *expressions* which combine both kinds of terms using `AND`, `OR` and `NOT` operators

For most adapters, it probably makes sense to handle each of these concepts separately when building a native query.

One way to implement this idea is to provide a class (named e.g. `SimpleFilterCompilerBase`) which the adapters would inherit from and implement methods responsible for translating concepts they support. For unsupported concepts the adapter simply wouldn’t implement the relevant method, which means a default base class implementation which throws a suitable exception would be used.

The `SimpleFilterCompilerBase` class can have an interface like this (in TypeScript):

```typescript
// Juttle AST node.
interface Node {}

// Base class of simple filter compilers.
class SimpleFilterCompilerBase {
    // Compiles a node. Generic method to be called when recursing.
    compile: (node: Node) => any;

    // Compiles a field reference. It is called only with Field nodes, e.g.:
    //
    //     { type: "Field", name: "a" }
    //
    compileField: (node: Node) => any;

    // Compiles a Juttle literal. It is called only with *Literal nodes, e.g.:
    //
    //     { type: "NumericLiteral", value: 5 }
    //
    // Typically, the method will switch on node.type, compile supported node
    // types and throw an exception in the default clause.
    compileLiteral: (node: Node) => any;

    // Compiles an expression term. It is only called with BinaryExpression
    // nodes, e.g.:
    //
    //    {
    //        type: "BinaryExpression",
    //        operator: "<",
    //        left: { type: "Variable", name: "a" },
[...]

Issue #123, opened 2016-01-06
include the batch interval in the time field of reduced points


demmer:
The points that emerge from a batched reduce operation, (either via `reduce -every <interval> ...` or via `batch <interval> | reduce ...`) represent the specified reduce calculation over all the points that occurred during the time interval `[batch_start, batch_end)`.

This is currently represented in Juttle by assigning the `time` field of the points to be the `batch_end - epsilon`, i.e. representing that the reduction result points logically "occurred" immediately before the end of the batch. This preserves causality in the dataflow computation since all points that occur between the batch start and the time right before the end of the batch occurs is included in the batch and hence happens before the batch timestamp, but a point that occurs right at the start of the next batch is not included.

When displaying these points, the current behavior leads to confusing output. For example, a simple program that counts the days in each month appears to be shifted by one month:

```
# juttle -e 'emit -from :2015-01-01: -every :1d: -limit 120 | reduce -every :1M: count()'
┌────────────────────────────────────┬──────────┐
│ time                               │ count    │
├────────────────────────────────────┼──────────┤
│ 2015-02-01T00:00:00.000Z           │ 31       │
├────────────────────────────────────┼──────────┤
│ 2015-03-01T00:00:00.000Z           │ 28       │
├────────────────────────────────────┼──────────┤
│ 2015-04-01T00:00:00.000Z           │ 31       │
├────────────────────────────────────┼──────────┤
│ 2015-05-01T00:00:00.000Z           │ 30       │
└────────────────────────────────────┴──────────┘
```

While it would certainly be possible to convey the epsilon bit to the view, a better representation for the time of these points would be to include the batch duration as well.

This way the above table could display the interval as something like:

```
┌──────────────────────────────────────────────────────────────┬──────────┐
│ time                                                         │ count    │
├──────────────────────────────────────────────────────────────┼──────────┤
│ 2015-01-01T00:00:00.000Z..2015-02-01T00:00:00.000Z           │ 31       │
├──────────────────────────────────────────────────────────────┼──────────┤
│ 2015-02-01T00:00:00.000Z..2015-03-01T00:00:00.000Z           │ 28       │
├──────────────────────────────────────────────────────────────┼──────────┤
│ 2015-02-01T00:00:00.000Z..2015-04-01T00:00:00.000Z           │ 31       │
├──────────────────────────────────────────────────────────────┼──────────┤
│ 2015-04-01T00:00:00.000Z..2015-05-01T00:00:00.000Z           │ 30       │
[...]

PR #384, opened 2016-02-11
filters: allow property accesses

davidvgalbraith:
The static filter checker rejects any read filters that
have property accesses. But these can be handled in some
cases (at least by Elastic), so let's allow them.

@juttle/developers

### DIFF
--- a/lib/compiler/filters/static-filter-checker.js
+++ b/lib/compiler/filters/static-filter-checker.js
@@ -18,7 +18,7 @@ var checks = {
         displayName: 'field',
 
         test: function(node) {
-            return node.type === 'Field';
+            return node.type === 'Field' || checks.isMemberAccess.test(node);
         }
     },
 
@@ -50,6 +50,19 @@ var checks = {
                 || node.type === 'StringLiteral'
                 || node.type === 'MomentLiteral'
                 || node.type === 'DurationLiteral';
+        }
+    },
+
+    isMemberAccess: {
+        displayName: 'property access',
+
+        test: function(node) {
+            var hasRightType = node.type === 'MemberExpression';
+            var fieldNameIsSimple = node.property && checks.isSimpleLiteral.test(node.property);
+            var accessIsFieldOrNest = node.object && (checks.isField.test(node.object) ||
+                checks.isMemberAccess.test(node.object));
+
+            return hasRightType && fieldNameIsSimple && accessIsFieldOrNest;
         }
     },
 
--- a/lib/compiler/filters/static-filter-compiler.js
+++ b/lib/compiler/filters/static-filter-compiler.js
@@ -58,6 +58,10 @@ class StaticFilterCompiler extends ASTVisitor {
         this.featureNotSupported(node, 'fields');
     }
 
+    visitMemberExpression(node) {
+        this.featureNotSupported(node, 'nested fields');
+    }
+
     visitUnaryExpression(node) {
         this.featureNotSupported(node, `the "${node.operator}" operator`);
     }
--- a/test/compiler/filters/static-filter-compiler.spec.js
+++ b/test/compiler/filters/static-filter-compiler.spec.js
@@ -136,6 +136,13 @@ describe('StaticFilterCompiler', () => {
         );
     });
 
+    it('doesn\'t compile MemberExpression', () => {
+        expect('a.b < 5').to.failToCompile(
+            NumberLiteralCompiler,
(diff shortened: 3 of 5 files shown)


dmajda:
Can this wait after the new filter API (#320) ends up implemented by adapters? Otherwise we risk that all non-supporting adapters will be required to implement a handler rejecting property accesses, only to delete it in the next cycle (because not handling a construct gracefully will be the default behavior of the new API).

davidvgalbraith:
This came up again today with Cook, shall we proceed?

dmehra:
Sounds like this was only blocked on https://github.com/juttle/juttle/issues/320 which has been done for a while, let's proceed @davidvgalbraith

dmajda:
> This came up again today with Cook, shall we proceed?

Yes. Please update the PR to the current master, I’ll have a look then.

davidvgalbraith:
Updated.

lib/compiler/filters/static-filter-checker.js
```diff
@@ -53,6 +53,14 @@ var checks = {
         }
     },
 
+    isPropertyAccess: {
+        displayName: 'property access',
+
+        test: function(node) {
+            return node.type === 'MemberExpression';
```
dmajda: The check is too benevolent as it allows passing whatever you want in `node.object` and `node.property`. For example, these would be valid filters (assuming `f` is a function):

```juttle
f()[1] < 5
field[f()] < 5
```

We should also check that `node.property` is a simple literal and that `node.object` is either a field or another `MemberExpression` (so that we allow nested member expressions: `a.b.c`, `a[0][1]`, etc.).

lib/compiler/filters/static-filter-checker.js
```diff
@@ -53,6 +53,14 @@ var checks = {
         }
     },
 
+    isPropertyAccess: {
```
dmajda: Rename to `isMemberAccess` please.

lib/compiler/filters/static-filter-checker.js
```diff
@@ -68,46 +76,55 @@ var checks = {
 var ALLOWED_OP_FORMS = {
     '==': [
         { left: checks.isField, right: checks.isSimpleLiteral },
-        { left: checks.isSimpleLiteral, right: checks.isField }
+        { left: checks.isSimpleLiteral, right: checks.isField },
+        { left: checks.isPropertyAccess, right: checks.isSimpleLiteral }
```
dmajda: Why not allow `isPropertyAccess` everywhere `isField` is allowed? That sounds like a logical thing to do to me.

If we go that route we should probably fold the `isPropertyAccess` check into `isField` to avoid duplication. The `isField` check would then mean “field, either plain or nested”.

dmajda:
It would be good to also add something like this into `StaticFilterCompiler` (right above `visitUnaryExpression`) and test it in `static-filter-compiler.spec.js`:

```js
visitMemberExpression(node) {
    this.featureNotSupported(node, 'nested fields');
}
```

dmajda:
I think we should increase the minor part of adapter API version because this change can break adapters. Adding a `MemberExpression` handler to `StaticFilterCompiler` will help in cases where an adapter compiles filters using a class descending from it and carefully calls `this.visit` for every node where `MemberExpression` now could appear, but this is far from a requirement.

We may also delay this a bit and bundle it in the same release as https://github.com/juttle/juttle/issues/443 which will also require minor version increase. That way we won’t have to update adapters two times in a row. What do you think, @demmer?

demmer:
There's also #641 (standardizing the adapter config) and #123 / #523 (changing the times on batch reduce to beginning of batch + interval) which affects any adapters that optimize reduce, all of which will require a change to the adapters and thereby warrant a version bump.

If we can get this in as part of this week's juttle release (0.7.0) I'd vote for doing so since this was requested by users and I don't want it to sit on the shelf just to avoid a bit of version bumping on the adapter modules.

davidvgalbraith:
Updated.

dmajda:
I’d rather get rid of `isMemberAccess` completely now that it is unused by `ALLOWED_OP_FORMS`. The test itself can also be simplified a bit (no need to test for property existence). But the code should work, so in the interest of getting this into 0.7.0 I’ll just squash & merge. We can clean it up later.

dmajda:
I also added a commit which adds `MemberExpression` handling to `StaticFilterCompiler`.

PR #385, opened 2016-02-11
Consolidate Juttle runtime: Kill Juttle.channels

dmajda:
Move `Juttle.channels` to program (the object generated by compiler and passed to procs).

Part of work on #97.


bkutil:
Looks good, thanks! +1

demmer:
This is a fine change, but it seems like a strange thing for the compiler to be managing now that we no longer have publish / subscribe in the mix.

It seems somewhat better for the `view` proc itself to just manage the channels. This would require some convention for procs attaching state to the program object so the channel numbering would start at 0 for each program.

dmajda:
Agreed, this is meant as a temporary solution before I get to clean up `Program` & co.

PR #386, opened 2016-02-11
Kill Juttle.adapters

dmajda:
Just use the `adapters` module directly where it is needed.

Part of work on #97.


dmajda:
@demmer

demmer:
I'm ok with this change -- note that juttle-service will also need to be updated.

PR #398, opened 2016-02-11
package.json: change scripts/peg.js to run in prepublish not postinstall

demmer:
To workaround an npm problem in which it sometimes refuses to run the
postinstall script, change the package so that it generates the parser
in the prepublish phase and not postinstall.

This also required explicitly enumerating the files to be put into the
package since by default it omits anything that's in .gitignore, but we
want it to include the generated files in the package.


mstemm:
lgtm.

PR #400, opened 2016-02-14
program: don't delete this.graph in deactivate

davidvgalbraith:
There's no need to delete this.graph here (as far as I can tell), and by not deleting
it we enable tests that call deactivate() to validate graphs.
@juttle/developers @dmajda


dmajda:
+1

PR #412, opened 2016-02-16
Write table properly for multiple tables.

### LINKED ISSUE #350: With multiple table views, title is not displayed
When running programs on the CLI, if multiple table views have `-title` option, no title is displayed for any of the tables. This makes it hard to understand the output. (If there is only one table, the title option is honored.)

```
juttle> emit | (view table -title 'One'; view table -title 'Two')
┌──────────────────────────┐
│ time                     │
├──────────────────────────┤
│ 2016-02-06T01:30:21.313Z │
└──────────────────────────┘
┌──────────────────────────┐
│ time                     │
├──────────────────────────┤
│ 2016-02-06T01:30:21.313Z │
└──────────────────────────┘
```

mstemm:
The bug was actually that table titles were not being written when
-progressive was set to false. It just happens that when specifying
multiple tables, -progressive gets set to false by default.

Use the right option (self.options.title) instead of self.title.

Also add a test to cover this so we'll know if it breaks again.

This fixes #350.

@dmehra

### DIFF
--- a/lib/views/table.js
+++ b/lib/views/table.js
@@ -229,8 +229,8 @@ var TableView = View.extend({
                 table.push(values.map(self._valueFormatter.bind(self)));
             });
 
-            if (self.title) {
-                self.fstream.write(self.title + '\n');
+            if (self.options.title) {
+                self.fstream.write(self.options.title + '\n');
             }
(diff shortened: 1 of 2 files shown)


demmer:
+1 lgtm

PR #413, opened 2016-02-16
consolidated adapter api module

### LINKED ISSUE #367: add adapter api and explicit version
To solve some of the challenges we've had in keeping adapters up to date with respect to changes in the juttle runtime, the API used by adapter implementations should be both more clearly defined and explicitly versioned.

Currently adapters `require` various modules from inside the juttle project. This means that when things move around inside the repository (e.g. as happened to JuttleMoment in #312), adapters need to catch up to the changes.

Instead we should have an explicit adapter implementation API. The idea here
would be that the only way adapters would pull in portions of the juttle codebase would be via:

`var AdapterAPI = require('juttle').AdapterAPI(version);`

This module would compose all of the portions of the juttle compiler and runtime
that adapter implementations are allowed to use. This would include:

 * AdapterRead
 * AdapterWrite
 * getLogger
 * types => lib/runtime/types
 * FilterJSCompiler
 * reducerEmptyValue(reducer)
 * testUtils => juttle-test-utils

The `parseTime`, `compileError`, and `runtimeError` utilities would be folded
into methods on the adapter base classes.

Note that as we implement this we will likely find additional modules that
should be included.

In addition, this API should have an explicit semantic version with major and
minor components. Each adapter implementation would include the version of the
API to which it is written. This way the runtime can check if it is compatible
with the version that the implementation wants and throw a user-friendly error
when trying to load an incompatible adapter.

demmer:
Implement the consolidated adapter API proposal in which the juttle runtime exports a single module containing the full adapter API as a global, thereby making it so that adapters should not need to require('juttle').

Detailed changes are described in the commits, but a high-level summary includes:

- Add a consolidated adapter API module that exports all core juttle functionality needed to implement the adapters.
- Export this API as a global as part of loading the adapters and remove the now unneeded Module.load hijacking that was added in #134.
- Load the adapter's package.json and check for a `juttleAdapterAPI' entry that contains the version of juttle with which the adapter should be compatible.
- Add a `reducerDefaultValue` utility function to avoid exposing all the reducer implementation to the adapters just to get the empty value.
- Add a facade around juttle-test-utils so adapters can simply `require('juttle/test').utils` instead of the gory path down into runtime/specs.
- Add an `adapterDescribe` hook that functions just like mocha.describe but after setting the `JuttleAdapterAPI` global. This way adapter unit tests can still pull in portions of the implementation.

Resolves #367


demmer:
@davidvgalbraith @dmajda @bkutil @mstemm @VladVega

lib/adapters/api.js
```diff
@@ -0,0 +1,30 @@
...
+// The version field corresponds to the juttle version in which the latest
+// changes were made to the adapter API.
+//
+let AdapterAPI = {
+    version: '0.5.0',
+    AdapterRead: require('./adapter-read'),
+    AdapterWrite: require('./adapter-write'),
+    compiler: {
+        ASTVisitor: require('../compiler/ast-visitor'),
+        FilterJSCompiler: require('../compiler/filters/filter-js-compiler')
```
davidvgalbraith: Who needs FilterJSCompiler?

lib/adapters/api.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+    errors: require('../errors'),
+    getLogger: require('../logger').getLogger,
+    runtime: {
```
davidvgalbraith: The Elastic adapter also does `var reducers = require('juttle/lib/runtime/reducers').reducers;`, though it's just used to extract the right default value for a given reducer. Should we add that the reducers here or shred that dependency?

davidvgalbraith: Oh hey look it's line 25! Carry on.

lib/adapters/api.js
demmer: I could have sworn one of the adapters supported read filtering and implemented it in javascript using FilterJSCompiler but can't find it now.

davidvgalbraith:
+1

lib/adapters/api.js
```diff
@@ -0,0 +1,30 @@
...
+//
+let AdapterAPI = {
+    version: '0.5.0',
+    AdapterRead: require('./adapter-read'),
+    AdapterWrite: require('./adapter-write'),
+    compiler: {
+        ASTVisitor: require('../compiler/ast-visitor'),
+        FilterJSCompiler: require('../compiler/filters/filter-js-compiler')
+    },
+    errors: require('../errors'),
```
mstemm: Do we need errors at all, or can all adapters just throw using this.xxxError?

mstemm:
The rest of the changes look good. I think by time of 0.5.0 release we should document the versioning in a readme but it doesn't have to be this pr.

lib/adapters/api.js
demmer: For the most part adapters can just use the methods attached to the read and write base classes, but there may be cases in which the error needs to be thrown from a global context such as when validating config loading the adapter or from a utility module. 

-m

lib/adapters/api.js
bkutil: SQL adapter does: https://github.com/juttle/juttle-sql-adapter-common/blob/master/lib/read.js#L6

dmajda:
+1 with comments addressed

demmer:
Addressed all of @dmajda's comments and rebased on top of master. Will merge on green.

PR #415, opened 2016-02-17
adapters: support an explicit eof bit in the read return

demmer:
Instead of requiring the adapter to create an infinite JuttleMoment
to indicate that it's done reading, also check for an eof bit in the
read() response.


demmer:
@davidvgalbraith @VladVega @rlgomes

davidvgalbraith:
+1

PR #416, opened 2016-02-17
adapters: add helpers to trigger throttled warnings

demmer:
Add helper functions to the adapter read base class to emit warnings
when points are dropped because they either arrive too late (in the case of
true live) or get dropped due to buffer overflow.


demmer:
@davidvgalbraith @dmajda @bkutil

lib/runtime/adapter-read.js
```diff
@@ -114,6 +115,27 @@ class AdapterRead {
         return errors.runtimeError(code, info);
     }
 
+    // Helper to obtain a throttled function that warns on late arrivals.
+    throttledWarnLate(interval) {
```
davidvgalbraith: These seem a bit overly-specific, how about just a method `throttledWarn(errorCode, info)` that adapters could call with whatever error code and info they want?

demmer: Elegant and simple. I like it.

demmer:
@davidvgalbraith How's this look?

davidvgalbraith:
+1

lib/runtime/adapter-read.js
```diff
@@ -114,6 +115,15 @@ class AdapterRead {
...
     }
 
+    // Helper to obtain a throttled function to emit warnings.
+    throttledWarn(code, info, interval) {
+        interval = interval || 1000;
+        info = info || {};
+        return _.throttle(() => {
+            this.trigger('warning', this.runtimeError(code, info));
+        }, interval);
+    }
```
bkutil: We already have similar logic in `periodic_fanin` (also for triggering an error only once), maybe it would be worth it moving this functionality into a common module, say `throttled_errors` / support it in `errors` directly? @dmajda thoughts?

In fanin, the interval value is related to `interval` option - would it make sense to set the default to the value of `every` in adapters?

demmer: @bkutil That seems like it could be a useful follow-up to consolidate even further.

Regardless I think it would still make sense to have a `this.throttledWarning` utility on the adapter class itself, so I'm going to merge this as-is and then we can separately consider whether to refactor further.

Issue #52, opened 2015-12-23
Make “reduce -every” emit marks


dmehra:
`reduce -every <x>` should be an equivalent to `batch <x> | reduce` in that it emits marks on batch boundaries. Currently it doesn’t do that.

The impact is, for example, that sinks with logic for redrawing on batch boundaries (such as `view barchart`) must be fed from a `batch | reduce` not from `reduce -every` if the redrawing is to work correctly. If the batch marks were emitted down either path, then the juttle programmer doesn't need to be aware of these special cases.

henridf:
oh no @dmajda are you going to let that happen?! :)

bkutil:
@henridf care to elaborate?

henridf:
Whether or not `reduce -every` should emit marks was one of the (many) juttle design questions that kept coming up in recurring debate until it was settled (!?) that they  shouldn't.  @dmajda can probably fill you in on the history...

dmajda:
@henridf Yes, I am going to let that happen :-)

Some of the decisions we made got revisited as the set of people influencing the language design changed. In this particular case, the discussion was philosophical — do we think of the stream as the set of points flowing through the flowgraph which are merely *viewed* as batches occasionally, or do we *really* send little table of points through (at least conceptually)? This is coupled with practical issues such as duplicating (or distributing) `-every` to procs following `reduce`, views, etc. In the end, the view that “tables are real” prevailed, at which point it made sense unify `batch | reduce` and `reduce -every` behavior.

Note this didn’t kill the plan of adding `-every` to other batch-aware procs, in fact the thinking is that `batch` should be more or less deprecated (we will keep it mostly for symmetry with `unbatch`, which we need).

henridf:
Thanks for the context @dmajda!

dmehra:
Will the resolution of this issue make it possible to write this program without using `batch`? It currently must use it because `sort -every` and `head -every` do not exist. The program is simply outputting top 100 links for each day.

```
read elastic -from :2009-04-01: -to :2009-05-01: link != null 
| batch :1 day:
| reduce count() by link
| sort count -desc
| head 100
```

dmajda:
@dmehra Yes, it will be just:

```
read elastic -from :2009-04-01: -to :2009-05-01: link != null 
| reduce -every :1 day: count() by link
| sort count -desc
| head 100
```

bkutil:
Closing as fixed.

PR #417, opened 2016-02-17
Make reduce -every emit marks at the end of epoch

bkutil:
`reduce -every` emits marks at the end of each epoch. This also requires changes for sort and tests that depend on the old behavior.

### DIFF
--- a/lib/runtime/procs/periodic-fanin.js
+++ b/lib/runtime/procs/periodic-fanin.js
@@ -141,8 +141,7 @@ class periodic_fanin extends oops_fanin {
             if (!this.epoch) {
                 this.first_epoch(this.next_epoch);
             } else {
-                var out = this.advance_epoch(this.next_epoch);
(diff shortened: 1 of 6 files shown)


bkutil:
The only tests that fail now are out-of-order tests. The issue is that marks after reduce now move the 'ooo-treshold' in `oops_fanin` forward, and so e.g. where a sequence of times `:1: | :0: | :1: | :0:` (`|` is a reduce boundary) would previously result in `:1: :1:`, with marks carrying the timestamp of the boundary, everything after the first point is dropped as ooo.

bkutil:
cc: @dmajda

dmajda:
+1

bkutil:
Updated to use `super` instead of `prototype.x.call` and updated commit messages. Will merge on green.

bkutil:
Fixes #52

PR #418, opened 2016-02-17
Simplify Filter AST

dmajda:
Simplify filter expression AST before it is passed to a compiler. More specifically:

  1. Remove `ExpressionFilterTerm` nodes (which the compiler typically just passes-through).

  2. Replace `SimpleFilterTerm` nodes which contain `StringLiteral` (and thus represent FTS) with structurally simpler `FulltextFilterTerm` nodes.

  3. Replace `SimpleFilterTerm` nodes which contain `FilterLiteral` (and thus represent embedded filter expression coming from an input ) with the actual `FilterLiteral` AST (effectively inlining the literal).

These transformations are done using `FilterSimplifier`, which is an instance of newly introduced `ASTTransformer` — a special kind of visitor which transforms the tree.

Part of work on #320.

### DIFF
--- a/lib/compiler/ast-transformer.js
+++ b/lib/compiler/ast-transformer.js
@@ -1,0 +1,46 @@
+'use strict';
+
+let ASTVisitor = require('./ast-visitor');
+let _ = require('underscore');
+
+// Base class for implementing AST transformers.
+//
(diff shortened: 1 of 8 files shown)


dmajda:
@bkutil

bkutil:
Looks good. Especially the transformer will be useful... +1

PR #420, opened 2016-02-17
Implement StaticFilterCompiler

dmajda:
`StaticFilterCompiler` is a base class for compilers that transform filter expression ASTs into native queries, to be inherited and extended by adapters.

Part of work on #320.


dmajda:
@bkutil

bkutil:
+1

Issue #396, opened 2016-02-11
Make SemanticPass a visitor


dmajda:
Making `SemanticPass` a visitor will make the code cleaner and more consistent with other compiler passes which we also want to convert to visitors (#394, #395).

Issue #395, opened 2016-02-11
Make GraphCompiler a visitor


dmajda:
Making `GraphCompier` a visitor will make the code cleaner and more consistent with other parts of the compiler. Most importantly, it will allow code sharing with `Build` (once it’s also [converted to a visitor](https://github.com/juttle/juttle/issues/394)) and `FilterJSCompiler`.

Issue #394, opened 2016-02-11
Make Build a visitor


dmajda:
Making `Build` a visitor will make the code cleaner and more consistent with other parts of the compiler. Most importantly, it will allow code sharing with `GraphCompiler` (once it’s also converted to a visitor) and `FilterJSCompiler`.

PR #425, opened 2016-02-18
Rename the “ast” parameter to “node” in various compiler passes

dmajda:
The new name is more logical and it is in line with most other code which processes AST nodes.

This is a small step toward #394, #395, and #396.


dmajda:
@bkutil

bkutil:
Phew! :) +1

PR #428, opened 2016-02-18
oops-fanin.js: Fix args passing to warn_oops

bkutil:
warn_oops was called without arguments from emit_mark, which caused an
error masking the actuall out-of-order problem


bkutil:
@dmajda

dmajda:
+1

Issue #195, opened 2016-01-15
http adapter: add support for paging


dmehra:
Support paging in `http read` via -offsetField and -stepField parameters (placeholder names, maybe good enough). 

Rationale: There are many rest-enabled backends that require paging. For example, [orchestrate](https://orchestrate.io/docs/apiref#search) asks for 'offset' and 'limit' parameters in the search URL. The specific field names will vary across APIs so we need to support them through options that let you specify the name. 

The http adapter needs to handle the actual stepping, i.e. incrementing offset = last_offset + limit, and running until all results came back.

rlgomes:
Implemented the easy solution here which is to follow Link header when its provided https://github.com/juttle/juttle/pull/435

Now looking at various APIs (pager duty, elasticsearch, orchestra, github, twilio, new relic) and how they do pagination here is the set of arguments that we would need to provide in order to support the various implementations I've come across:

**offsetField**: name of the field used to keep track of the current offset in terms of points returned, some use offset and some use from.
**offsetFieldStart**: the actual starting value for the offset, may not be necessary as most offsets driven pagination starts at 0.
**pageField**: instead of offsets some APIs such as github, new relic,  likes to use the notion of pages
**pageFieldStart**: because some start at 0 (twilio) and others at 1 (github actually returns the same results for page=0 and page=1 which is quite odd)
**limitField**: the name of the field that contains the actual limit so we can send that along with the URL ? or do we assume the user is setting that themselves ?
**limitfieldValue**: the value for the limit to be passed (otherwise you can't do the math on the offsetField above)

Does this actually seem like something we want to do as I'm feeling that supporting the non [Web Link](https://tools.ietf.org/html/rfc5988) is going to be quite complex 

Links to a few of the paginated APIs I sampled:

https://developer.github.com/guides/traversing-with-pagination/
https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html
https://docs.newrelic.com/docs/apis/rest-api-v2/requirements/pagination-api-output
https://developer.pagerduty.com/documentation/rest/pagination
https://orchestrate.io/docs/apiref#graph-get-params
https://www.twilio.com/docs/api/rest/response

@davidvgalbraith @dmajda @dmehra @demmer what do you guys think ?

dmajda:
[...]

PR #434, opened 2016-02-18
split up http adapter tests

rlgomes:
The main http.spec.js was almost 1000 lines long and by splitting these
tests up it makes maintenance a little easier. In the process of
splitting up also moved to do things in a more ES6-y way.


rlgomes:
@demmer @davidvgalbraith was working on #195 and realized that I had to add a bunch of new tests that would definitely make the existing `http.spec.js` file over  1000 lines and so splitting these up seemed like the right thing to do before I'd add tests or that feature.

davidvgalbraith:
+1

PR #435, opened 2016-02-18
add followLinkHeader to read adapter

### LINKED ISSUE #195: http adapter: add support for paging

rlgomes:
partially fixes #195

added the option `-followLinkHeader` to http read adapter that allows
for automatic following of the HTTP Link header as described in
https://tools.ietf.org/html/rfc5988

### DIFF
--- a/docs/adapters/http.md
+++ b/docs/adapters/http.md
@@ -21,6 +21,10 @@ read http -url url
           -rootPath rootPath
           -includeHeaders true/false
           -format csv/json/jsonl
+          -followLinkHeader true/false
+          -pageParam parameter
+          -pageStart start
+          -pageUnit unit
 ```
 
 Parameter         |             Description          | Required?
@@ -34,6 +38,10 @@ Parameter         |             Description          | Required?
 `-includeHeaders` | When set to true the headers from the response are appended to each of the points in the response | No; default: `false`
 `-timeField`      | Name of the field in the data which contains a valid timestamp  | No; defaults to `time`
 `-format`         | HTTP data format, supports: `csv`, `json` and `jsonl` | No; defaults to HTTP response header `Content-Type`
+`-followLinkHeader` |  When set to true the [Web Linking](https://tools.ietf.org/html/rfc5988) feature is enabled which allows for automatic handling of paging as specified by the [RFC5988](https://tools.ietf.org/html/rfc5988). | No; default `false`
+`-pageParam`      | The name of the query parameter used to enable paging with multiple HTTP requests sent to the URL provided until response contains 0 records | No; defaults to paging behavior disabled
+`-pageStart`      | Initial value of the paging offset to set the `pageParam` to in the URL used to make the HTTP request | No; default: `0`
+`-pageUnit <request|record>` | If set to `request`, then each successive HTTP paging request increments the paging query parameter by one. If set to `record`, then each successive request increments the paging query parameter by the number of records returned in the previous request. | No; default: `request`
 
 Currently the `read http` adapter will automatically parse incoming data based off of the `content-type` header. Here are the currently supported content-types:
 
--- a/lib/adapters/http/read.js
+++ b/lib/adapters/http/read.js
@@ -5,15 +5,20 @@ var contentType = require('content-type');
 var AdapterRead = require('../adapter-read');
 var parsers = require('../parsers');
 var errors = require('../../errors');
+var parseLinkHeader = require('parse-link-header');
+var querystring = require('querystring');
 var Promise = require('bluebird');
 var request = require('request');
 var JuttleMoment = require('../../runtime/types/juttle-moment');
+var urlParse = require('url').parse;
+var urlFormat = require('url').format;
 
 class Iterator {
     constructor(adapter, res, from, to) {
         this.points = [];
         this.done = false;
         this.adapter = adapter;
+        this.total = 0;
 
         var format;
         try {
@@ -26,9 +31,18 @@ class Iterator {
             rootPath: this.adapter.rootPath,
             optimization: this.adapter.optimization_info
         });
-        this.adapter.logger.debug('starting parse');
+
+        if (res.headers['link']) {
+            var parsed = parseLinkHeader(res.headers['link']);
+            if (parsed.next) {
+                this.adapter.logger.debug('following link header', parsed.next.url);
+                this.linkHeader = parsed.next.url;
+            }
+        }
+
         this.parser.parseStream(res, (points) => {
             points = this.process(res, points, from, to);
+            this.total += points.length;
 
             // The current parser API doesn't offer any kind of backpressure,
             // so all we can do is manage our own buffer of points and return
@@ -41,7 +55,7 @@ class Iterator {
             this.adapter.trigger('error', err);
         })
         .then(() => {
-            this.adapter.logger.debug('done');
+            this.adapter.logger.debug('parse done, got', this.points.length, 'points');
             this.done = true;
             this.notify();
         });
@@ -78,12 +92,11 @@ class Iterator {
             this.waiter = null;
             var points = this.points;
             this.points = [];
-            var ret = {
-                state: this,
+            return {
                 points: points,
+                state: this,
                 readEnd: this.done ? new JuttleMoment(Infinity) : null
             };
-            return ret;
         });
     }
 
@@ -100,6 +113,30 @@ class ReadHTTP extends AdapterRead {
     constructor(options, params) {
         super(options, params);
 
+        if (params.filter_ast) {
+            throw new errors.compileError('ADAPTER-UNSUPPORTED-FILTER', {
+                proc: 'read http',
+                filter: 'filtering'
+            });
+        }
+
+        _.each(['pageParam', 'pageUnit', 'pageStart'], (option) => {
+            if (options.followLinkHeader && options[option] !== undefined) {
+                throw new errors.compileError('INCOMPATIBLE-OPTION', {
+                    option: 'followLinkHeader',
+                    other: option
+                });
+            }
+        });
+
+        if (options.pageUnit &&
+            !_.contains(['request', 'record'], options.pageUnit)) {
+            throw new errors.compileError('INVALID-OPTION-VALUE', {
+                option: 'pageUnit',
+                supported: 'request, record'
+            });
+        }
+
         this.url = options.url;
         this.method = options.method ? options.method : 'GET';
         this.headers = options.headers ? options.headers : {};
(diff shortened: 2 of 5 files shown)


rlgomes:
Initial implementation just to get some reviewing of the code before I'd implement the other offset/limit driven paging and all of the necessary tests. @davidvgalbraith @demmer @go-oleg see if the code changes make sense or if there's anything fishy.

davidvgalbraith:
I don't see anything terribly fishy so far.

rlgomes:
@davidvgalbraith added tests and then waiting to see what the response is to my comment [here](https://github.com/juttle/juttle/issues/195#issuecomment-186088795).

lib/adapters/http/read.js
```diff
@@ -41,6 +48,19 @@ class Iterator {
...
         })
         .then(() => {
+            if (this.adapter.followLinkHeader && res.headers['link']) {
+                var parsed = parseLinkHeader(res.headers['link']);
+                if (parsed.next) {
+                    this.adapter.logger.debug('following link header', parsed.next.url);
+                    return this.adapter.fetch(parsed.next.url)
+                    .then((res) => {
+                        this.parse(res, from, to);
+                        res.resume();
```
davidvgalbraith: What exactly is the story with the pause/resuming? Do the tests fail if you take it out?

rlgomes: The tests do fail and its for a valid reason. When you send back the `Response` object without first hooking up synchronously the `.on` `data`, `error`, etc. handlers that we do in the `new Iterator`-> `getParser` path then the resolving of the `Response` happens asynchronously up to where we do call the `new Iterator` and by then the `on('data'` for the first few bytes has already been called and that data is gone. Try removing the resume/pause and you'll see. I only had to shift around where the Iterator was being created so we could easily do new `fetch` calls without having to recreate the iterator each time but maybe there's a nicer way to achieve the same thing ?

rlgomes:
@davidvgalbraith @demmer updated to now use the `pageParam`, `pageStart` and `pageUnit` to handle a larger audience of different paginating REST APIs. I can find a way to fix that situation of using the `resume/pause` on the streams by pulling the parser initialization out of the iterator unless you guys have a better idea.

davidvgalbraith:
Let's see how the pulling out looks, still haven't quite wrapped my head around the `pause`/`restart` stuff.

rlgomes:
@davidvgalbraith got rid of the whole pause/resume business but it means incurring the cost of recreating the `Iterator` object and any supporting objects under it. Not a huge cost and probably a better option since this is "easier" to read and by creating a new `Iterator` we can tolerate changes from a new HTTP request possibly requiring a slightly different parser (can't think of a good reason to do that but it is supported now).

rlgomes:
I'll squash these commits down to a single one once I've gotten the +1's

demmer:
@rlgomes Last week I did a bit of hacking on http paging myself -- not ready for merge, but take a look at https://github.com/juttle/juttle/compare/http-read-paging?expand=1.

In this approach I kept all the paging logic in the adapter itself instead of the iterator. The Iterator is just responsible for parsing the results and in this case keeping track of whether or not it got any points.

I think that's cleaner overall and easier to follow, but take a look and let me know what you think.

rlgomes:
@demmer I think I get how to use that approach will have to also dangle more info on the `result.state` so that I have easy access to the resulting Link header for each request.

rlgomes:
@demmer @davidvgalbraith reworke with demmer's suggestion and squashed down to a single commit. I think you'll be happier with this solution.

lib/adapters/http/read.js
```diff
@@ -144,26 +180,81 @@ class ReadHTTP extends AdapterRead {
...
-            });
-        } else {
-            return iterator.next(to);
+        // The first read call should kick off the actual fetch request. This
+        // returns a promise that either rejects with an error or resolves with
+        // an iterator object that handles paging through the results in chunks
+        // that we return as the read state.
+        var url = this.url;
+
+        if (this.link) {
```
davidvgalbraith: This is a little funky, can we just overwrite `this.url` on line 228?

lib/adapters/http/read.js
```diff
@@ -144,26 +180,81 @@ class ReadHTTP extends AdapterRead {
...
+                result.readEnd = undefined;
+                result.state = undefined;
+            }
+
+            return result;
+        });
     }
 
-    fetch(from, to) {
+    fetch(thisURL, from, to) {
```
davidvgalbraith: `thisURL` is a pretty confusing variable name, looks like a misprint of `this.URL`. What's wrong with plain old `url`?

test/adapters/http/read.spec.js
```diff
@@ -71,6 +71,32 @@ describe('HTTP read tests', function() {
...
+            .catch(function(err) {
+                expect(err.toString()).to.contain('-followLinkHeader option should not be combined with -' + option);
+            });
+        });
+    });
+
+    _.each(['request', 'record'], (unit) => {
+        it('can handle pageUnit "' + unit + '"', () => {
+            return check_juttle({
+                program: 'read http -pageUnit "' + unit + '" -url "' + server.url + '/status/200"'
```
davidvgalbraith: Template strings!

test/adapters/http/read.spec.js
```diff
@@ -71,6 +71,32 @@ describe('HTTP read tests', function() {
...
+            })
+            .then((result) => {
+                expect(result.errors).deep.equals([]);
```
davidvgalbraith: Can we validate the results here too?

test/adapters/http/test-server.js
```diff
@@ -343,6 +343,74 @@ class TestHTTPServer {
                 }
             });
 
+            this.app.get('/pageByRequest', (req, res) => {
+                // returns 5 pages with 1 point at a time and provides the
+                // `Link` header needed to get the subsequent elements
```
davidvgalbraith: Where do you actually set the link header?

lib/adapters/http/read.js
rlgomes: Yes I think that is perfectly fine.

lib/adapters/http/read.js
rlgomes: url is the lib required up above `var url = require('url')` and so I wanted to avoid the name collision and what I'll do instead is simply `var urlParse = require('url').parse` and be able to use the url var here just fine.

test/adapters/http/read.spec.js
rlgomes: I find them to not be that great looking but I will apply here.

test/adapters/http/read.spec.js
rlgomes: Sure

test/adapters/http/test-server.js
rlgomes: This route if for paging using `-paramUnit 'request'` and not the one you're looking for which is [here](https://github.com/juttle/juttle/pull/435/files#diff-0835c78cfbeb374688a7590ad92c8564R401) and used when you use the `-followLinkHeader`.

rlgomes:
@davidvgalbraith updated

rlgomes:
@demmer looks good to you ? You had the suggestion from your other branch and I want to be sure things were done right.

rlgomes:
@dmehra check my English in the docs additions please :)

lib/adapters/http/read.js
```diff
@@ -100,6 +113,30 @@ class ReadHTTP extends AdapterRead {
...
+            if (options.followLinkHeader && options[option] !== undefined) {
+                throw new errors.compileError('INCOMPATIBLE-OPTION', {
+                    option: 'followLinkHeader',
+                    other: option
+                });
+            }
+        });
+
+        if (options.pageUnit &&
+            _.indexOf(['request', 'record'], options.pageUnit) === -1) {
```
davidvgalbraith: How about `!_.contains(['request', 'record'], options.pageUnit)`?

docs/adapters/http.md
```diff
@@ -34,6 +38,10 @@ Parameter         |             Description          | Required?
 `-includeHeaders` | When set to true the headers from the response are appended to each of the points in the response | No; default: `false`
 `-timeField`      | Name of the field in the data which contains a valid timestamp  | No; defaults to `time`
 `-format`         | HTTP data format, supports: `csv`, `json` and `jsonl` | No; defaults to HTTP response header `Content-Type`
+`-followLinkHeader` |  When set to true the [Web Linking](https://tools.ietf.org/html/rfc5988) feature is enabled which allows for automatic handling of paging as specified by the [RFC5988](https://tools.ietf.org/html/rfc5988). | No; default `false`
+`-pageParam`      | The name of the query parameter used to enable paging with multiple HTTP requests sent to the URL provided until response contains 0 records | No
```
dmehra: Needs default

docs/adapters/http.md
```diff
@@ -34,6 +38,10 @@ Parameter         |             Description          | Required?
...
+`-pageStart`      | Initial value of the paging offset to set the `pageParam` to in the URL used to make the HTTP request | No; default: `0`
+`-pageUnit`       | Can be either `request` or `record`. If set to `request`, then each successive HTTP paging request increments the paging query parameter by one. If set to `record`, then each successive request increments the paging query parameter by the number of records returned in the previous request. | No; default: `request`
```
dmehra: we usually specify the value choices in the "Parameter" column, let's do like this:
```
`-pageUnit <request|record>`    | If set to `request`, then ...
```

docs/adapters/http.md
rlgomes: @dmehra the default is its not set and not required since it would enable a "behavior".

dmehra: Ok, maybe say "No; defaults to paging behavior disabled".

docs/adapters/http.md
```diff
@@ -21,6 +21,10 @@ read http -url url
           -rootPath rootPath
           -includeHeaders true/false
           -format csv/json/jsonl
+          -followLinkHeader true/false
+          -pageParam paramter
```
dmehra: s/paramter/parameter or maybe parameterName

davidvgalbraith:
+1

Issue #419, opened 2016-02-17
Make member access ("foo.bar", "foo[bar]") work properly


dmajda:
## Specification

Non-computed member access (`foo.bar`) should work like this in rvalues:

  * If *object* is a module reference, it accesses specified property (const, function, etc.). If there is 
none, it produces an error.
  * If *object* is an expression evaluating to an `Object` and *property* is an expression evaluating to a `String`, it accesses specified property. If there is none, the result is `null`.
  * Otherwise, it produces an error.

Computed member access (`foo[bar]`) should work like this in rvalues:

  * If *object* is an expression evaluating to an `String` and *property* is an expression evaluating to a `Number`, it accesses specified character. If there is none, the result is `null`.
  * If *object* is an expression evaluating to an `Array` and *property* is an expression evaluating to a `Number`, it accesses specified element. If there is none, the result is `null`.
  * If *object* is an expression evaluating to an `Object` and *property* is an expression evaluating to a `String`, it accesses specified property. If there is none, the result is `null`.
  * Otherwise, it produces an error.

Non-computed member access (`foo.bar`) should work like this in lvalues:

  * If *object* is an expression evaluating to an `Object` and *property* is an expression evaluating to a `String`, it sets specified property (possibly overwriting existing value).
  * Otherwise, it produces an error.

Computed member access (`foo[bar]`) should work like this in lvalues:

  * If *object* is an expression evaluating to an `Array` and *property* is an expression evaluating to a `Number`, it sets specified element (possibly overwriting existing value) and adjusts array length (like in JavaScript).
  * If *object* is an expression evaluating to an `Object` and *property* is an expression evaluating to a `String`, it sets specified property (possibly overwriting existing value).
  * Otherwise, it produces an error.

## Checklist

  * [x] Preparatory refactorings
  * [x] Make the `[]` operator work properly in rvalues
  * [x] Make the `.` operator work properly in rvalues
    * [x] Fix handling in non-call expressions (`foo.bar`)
    * [x] Fix handling in call expressions (`foo.bar()`)
  * [x] Make the `[]` operator work properly in lvalues
  * [x] Make the `.` operator work properly in lvalues
  * [x] Make the `.` operator work properly in sub calls

dmajda:
Done, done, done!

PR #436, opened 2016-02-19
SemanticPass: Do not return nodes

dmajda:
The semantics pass was somewhat schizophrenic: its methods modified AST nodes in-place, but they also returned them, which would normally suggest a functional interface without in-place modifications.

This commit bends semantic pass internals firmly toward the in-place modification side, getting rid of node returns and fixing all places where returned nodes were used. The other alternative was to keep the returns and modify the code to work in functional style, always returning new nodes. This would complicate the code, cause performance problems, and provide little overall benefit, so I decided not to go this way, despite its apparent elegance.

One disadvantage of this approach is that `sa_Variable` and `sa_reducer_arg`, which created new nodes, now have to modify the existing ones, including changing their type. I suppose this will get solved once the semantic pass is split into multiple passes (which is a long-term goal) and pass containing these becomes an instance of `ASTTransformer`.

Note the schizophrenia was kept at the interface level (the `analyze` method). I may provide a functional interface at that level in the future by cloning the whole tree before processing, but I didn’t want to go into that now.

Also note this change reduces quite some code which was previously part of assignments to pure traversal (see e.g. `sa_BinaryExpression`). This means it will be possible to replace it by `super.*` calls when `SemanticPass` becomes an instance of `ASTVisitor`.

This is a prep work for #419.

### DIFF
--- a/lib/compiler/graph-builder.js
+++ b/lib/compiler/graph-builder.js
@@ -204,7 +204,12 @@ class GraphBuilder {
     value_ast(result) {
         // The _semantic_ast call is needed to correctly process filter
         // literals.
-        return this._semantic_ast(values.toAST(result));
+        var semantic = new SemanticPass({ now: this.now });
+        var ast = values.toAST(result);
+
+        semantic.sa_expr(ast);
+
+        return ast;
     }
     build_sub_args(sub_sig, callopts, subname, location) {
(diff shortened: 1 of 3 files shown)


dmajda:
@bkutil

bkutil:
+1

PR #439, opened 2016-02-20
Fanin readability

bkutil:
Improve readability of fanin's 'input with earliest point' selection.

Commit refactoring the `earliest` loop will be squashed - they were kept separate for easier review.


bkutil:
cc: @dmajda

bkutil:
Addressed the comments.

dmajda:
+1

PR #440, opened 2016-02-20
test-adapter-timeseries.js - support emitting more points during one tick

bkutil:
For testing two parallel timeseries reads where each reads different # of points.

### DIFF
--- a/test/runtime/test-adapter-timeseries/index.js
+++ b/test/runtime/test-adapter-timeseries/index.js
@@ -25,6 +25,7 @@ class TestTimeseriesRead extends AdapterRead {
         }
 
         this.every = options.every || this.defaultTimeOptions().every;
+        this.density = options.density || 1; // points per batch
         this.count = 0;
     }
 
@@ -39,6 +40,8 @@ class TestTimeseriesRead extends AdapterRead {
         });
     }
 
+    static allowedOptions() { return super.commonOptions().concat(['density']); }
+
     read(from, to, limit, state) {
         // Use the state to store the next timestamp to send.
         if (! state) {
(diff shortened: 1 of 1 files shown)


bkutil:
cc: @demmer

demmer:
I don't understand the point of this change -- if you want ten points per second, then you can accomplish the same thing with `-every :100ms:` as you can with `-every :1s: -density 10`.

On the other hand I think it would make sense to support an option for the adapter to emit multiple points with each timestamp. Maybe something like `-numSeries` which would default to 1. Then each point would include a `series` attribute, i.e. something like:

`read testTimeseries -last :2s: -every :1s: -numSeries 2`

would emit

```
[
{"time":"2016-02-22T15:04:51.422Z","series:0,"count":0},
{"time":"2016-02-22T15:04:51.422Z","series:1,"count":1},
{"time":"2016-02-22T15:04:52.422Z","series:0,"count":2}
{"time":"2016-02-22T15:04:52.422Z","series:1,"count":3}
]
```

bkutil:
What I wanted to have was a way to specify a `burst rate` - the number of points emitted into the flow graph during a single `every`, to see how different reads interact and how much memory is consumed in the process. E.g. 

`(read x -every :0.1s: -to :end:; read y -every :1s: -density 10 -to :end:) | write`

causes one of the `write`'s two input queues to fill differently from

`(read x -every :0.1s: -to :end:; read y -every :0.1s: -to :end:) | write`

The first example will cause the input queue of the first read to rise up to 10 points (waiting for points to come from the second read, then merge on timestamps), the second example will keep input queues max on 1, since `fanin` can do the merging on the go. So while they look equivalent in terms of output, the memory footprint is different.

bkutil:
@demmer bump. I can still use this locally, so I'll leave it up to you whether to merge this or not.

PR #442, opened 2016-02-22
SemanticPass: Kill unames (almost)

dmajda:
The semantic pass added two properties to `Variable` and `MemberExpression` nodes:

  * `uname` — unique name of the variable or module export, to be used in compiled code
  * `symbol` — symbol table entry for the variable or module export

Since the symbol table entry already contained the unique name, setting the `uname` property on the nodes was redundant, and I decided to remove the `uname` assignment. Note this makes a lot of sense conceptually — the unique name is a property of the referenced object (variable, function, reducer,…), not of the reference to it, so it belongs to its symbol table entry, not to the node referencing it.

During the work it turned out that `uname` is also set for nodes which *define* symbols, such as `FunctionDef` or `ReducerDef`. I decided to make these consistent with `Variable` and `MemberExpression` and use the `symbol` property there too. I also did few other small cleanups.

I left the `uname` property assignment for nodes corresponding to procs as there are no symbols defined for them.

Prep work for #419.

### DIFF
--- a/lib/compiler/build.js
+++ b/lib/compiler/build.js
@@ -293,14 +293,14 @@ class Build extends CodeGenerator {
     gen_FunctionDef(node) {
         var k, elems = node.elements;
         var args = _.map(node.args, arg => arg.symbol.uname);
-        this.emit('var ' + node.uname + '= function(' + args.join(', ') + ') {\n');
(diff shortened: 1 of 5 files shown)


dmajda:
@bkutil

bkutil:
+1

PR #445, opened 2016-02-23
Do not coerce identifiers in multi-value options to strings

### LINKED ISSUE #444: Identifiers in proc options get coerced to strings
Example:
```
$ bin/juttle -e 'emit -limit x, y, z'
<input>:1:1: 
   1:emit -limit x, y, z
     ^^^^^^^^^^^^^^^^^^^
Error: -limit wants a number, got [ "x", "y", "z" ] (NUMBER-ERROR)
```

The above program should complain about `x` not being defined.

Note the bug is present only when the option value is a list (represented by `ByList` AST node). When not, everything works as expected:

```
$ bin/juttle -e 'emit -limit x'
<input>:1:13: 
   1:emit -limit x
                 ^
Error: x is not defined (UNDEFINED)
```

dmajda:
Before this commit, option values were parsed using the `ByList` rule. This was an easy way to parse multi-value options, but it meant the parser returned multiple values as `ByList` nodes, which triggered
coercion rules for `by` clauses in the semantic pass. This was undesirable.

This commit changes the grammar to parse options using a new `OptionValue` rule which returns multiple option values as `ArrayLiteral` nodes. These don’t trigger coercion in the semantic pass.

Note the change also means options no longer auto-flatten their values (this was another side effect of using `ByList`). I think this is OK — the behavior was surprising and it probably needs some tweaking even in case of `by` clauses.

Fixes #444.

### DIFF
--- a/lib/parser/parser.pegjs
+++ b/lib/parser/parser.pegjs
@@ -1726,20 +1726,15 @@ ProcStatement
 ProcOptions = ProcOption*
 
 ProcOption
-    = ___ id:ProcOptionName ACOptionValueStart ___ ACOptionValueEnd by:ByList {
-          // If the list has only 1 element, pass the element itself.
-          if (by.elements.length === 1) {
-              by = by.elements[0];
-          }
-
+    = ___ id:ProcOptionName ACOptionValueStart ___ ACOptionValueEnd value:OptionValue {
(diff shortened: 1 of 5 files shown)


dmajda:
@bkutil

bkutil:
+1

PR #446, opened 2016-02-23
Add ASTTransformer and StaticFilterCompiler to adapter API

dmajda:
`ASTTransformer` can be used by adapters which transform filter AST before processing, e.g. the InfluxDB adapter.

`StaticFilterCompiler` is the new base class for filter compilers.

Related to #320.


dmajda:
@bkutil

bkutil:
+1

PR #447, opened 2016-02-23
CHANGELOG.md: Add 0.5.0 entries

dmajda:


dmajda:
@demmer

CHANGELOG.md
```diff
@@ -2,6 +2,85 @@
...
 
+## 0.5.0
+
+### Major Changes
+
+- Implemented consolidated, versioned adapter API. Adapters should use the `AdapterAPI` global variable to access the API instead of `require`-ing modules from the `juttle` package directly. [[#367](https://github.com/juttle/juttle/issues/367)]
+
+### Minor Changes
+
+- Made `reduce -every` emit marks at the end of epoch. [[#52](https://github.com/juttle/juttle/issues/52)]
```
demmer: This seems like a major change to me since it changes the semantics of programs.

CHANGELOG.md
```diff
@@ -2,6 +2,85 @@
...
+- Made debug logging controllable using the `DEBUG` environment variable. [[#373](https://github.com/juttle/juttle/issues/373)]
+- Removed `RT-` and `JUTTLE-` prefixes from error codes. Also removed the `-ERROR` suffix in cases where it was redundant. [[#356](https://github.com/juttle/juttle/pull/356)]
+- Disallowed `-from` to be equal to `-to` in sources. [[#322](https://github.com/juttle/juttle/issues/322)]
+- Cleaned up proc inheritance chain. [[#371](https://github.com/juttle/juttle/pull/371)]
+- Added allowed option validation logic into `base` and used it in built-in procs. [[#317](https://github.com/juttle/juttle/pull/317)]
+- Added required option validation logic into `base`. [[#353](https://github.com/juttle/juttle/pull/353)]
+- Changed `procName` from a prototype property on the procs into a function. [[#364](https://github.com/juttle/juttle/pull/364)]
+- Unique proc identifier (`pname`) is now passed to procs in `params` which are now also attached to proc instances. [[#365](https://github.com/juttle/juttle/pull/365)]
+- In views, changed `name` from a prototype property to an instance variable. [[#375](https://github.com/juttle/juttle/pull/375)]
```
demmer: It seems like the entries from "Cleaned up proc inheritance chain" down are just implementation details and not actually things that affect users. I'm ok either way but since this list is already long I'd also be fine to drop them.

CHANGELOG.md
```diff
@@ -2,6 +2,85 @@
...
+- Simplified AST representation of fields. [[#318](https://github.com/juttle/juttle/issues/318)]
+- Introduced `ASTTransformer` and `StaticFilterCompiler` classes. The latter is a base class the filter compilers in adapters shuld be generally derived from. [[#420](https://github.com/juttle/juttle/pull/420), [#446](https://github.com/juttle/juttle/pull/446)]
+- Converted the codebase to use ES6 classes instead of `extendable-base`. [[#341](https://github.com/juttle/juttle/issues/341)]
+- Replaced Backbone.js Events and built-in Node.js events with EventEmitter3. [[#49](https://github.com/juttle/juttle/issues/49)]
+- Moved classes representing Juttle types (currently `JuttleMoment` and `Filter`) to `lib/runtime/types`. [[#312](https://github.com/juttle/juttle/pull/312)]
+- Finished consolidation of Juttle runtime: [[#97](https://github.com/juttle/juttle/issues/97)]
+  - Dissolved the global `Juttle` variable.
+  - Made compiled programs independent on `JuttleMoment` and `Filter` variables.
```
demmer: Grammatically I think this should either be "independent of" or "not depend on". Same for the next one.

demmer:
A few comments but otherwise lgtm.

CHANGELOG.md
dmajda: OK, will move.

CHANGELOG.md
dmajda: I consider adapter creators and embedders as users too, so I also included entries relevant for them. There is certainly some fuzzy boundary there because we don’t document the API, so it was sometimes hard to say tell whether a change was just an internal thing or not.

I think I’ll just keep the list. For future releases, we may consider structuring the changelog differently (sections like “Language”, “Adapters”, “API”, etc.) and define the boundary more explicitly.

CHANGELOG.md
dmajda: Will fix.

dmajda:
Updated, merging.

Issue #58, opened 2015-12-26
Complete core library modules: String, Array, Object


dmehra:
Core library modules such as `String`, `Array`, and `Object` are mostly ad-hoc and incomplete. Finish them.

Specifically for Array, @dmajda proposed:

Implement the `Array` module with basic array-manipulating functions. A good start may be just wrapping [functions defined on arrays in ECMAScript 5.1](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.3).

We may also want to add `Array.zip` - the lack of it today requires the user to write juttle reducers like this:

```
reducer zip(a, b) {
  // return an array of pairs [[a0, b0], [a1, b1], ...]
  var A = [], B = [];
  function update() {
    A = *a; B = *b;
  }
  function zip_tail(A, B, C, n) {
    if (A[n] == null) { 
      return C;
    }
    C[n] = [A[n], B[n]];
    return zip_tail(A, B, C, n+1);
  }
  function result() {
    return zip_tail(A, B, [], 0);
  }
}
```

to be used like this:

```
 | put px = percentile(x, [0, .5, 1]), py = percentile(y, [0, .5, 1])
 | put qq = zip(px, py)
 | split qq
```

dmajda:
The high-level idea for this issue (which was lost by copying the original not-that-useful description) is to complete Juttle’s core library — more specifically, the `String`, `Array`, and `Object` modules where we are missing useful functionality. Some time ago I and Henri decided not to reinvent the wheel but look at what `String`, `Array`, and `Object` objects in ES5 have and more or less copy that into Juttle core modules. This approach has some advantages:

  1. The functionality will be familiar to our users who know JavaScript.
  2. We don’t have to invent our own core library (with all the inevitable bikeshedding).
  3. We already started on this course, so it’s natural to continue.
  4. The implementation will be simple.

The biggest disadvantage is that JavaScript core functionality is sometimes crappy.

---

Looking quickly at the [ES 5.1 spec](http://www.ecma-international.org/ecma-262/5.1/) I think it makes sense for Juttle to steal the following:

  * `Array.toString`
  * `Array.concat` (probably make it work just for arrays)
  * `Array.join`
  * `Array.pop`
  * `Array.push`
  * `Array.reverse`
  * `Array.shift`
  * `Array.slice`
  * `Array.sort` (use Juttle value comparison, not JavaScript’s crapy behavior here)
  * `Array.splice`
  * `Array.unshift`
  * `Array.length`
[...]

PR #449, opened 2016-02-23
string: add the rest of the functions

davidvgalbraith:
- [x] String.fromCharCode
- [x] String.charAt
- [x] String.charCodeAt
- [x] String.match
- [x] String.trim

relates to https://github.com/juttle/juttle/issues/58

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -171,9 +171,13 @@ class SemanticPass {
             exports: {
                 /* Functions (sorted alphabetically) */
                 length: { type: 'function', exported: true, uname: 'juttle.modules.string.length', arg_count: 1             },
+                charAt: { type: 'function', exported: true, uname: 'juttle.modules.string.charAt', arg_count: 2             },
+                charCodeAt: { type: 'function', exported: true, uname: 'juttle.modules.string.charCodeAt', arg_count: 2     },
                 concat: { type: 'function', exported: true, uname: 'juttle.modules.string.concat', arg_count: [0, Infinity] },
+                fromCharCode: { type: 'function', exported: true, uname: 'juttle.modules.string.fromCharCode', arg_count: [0, Infinity]},
                 indexOf: { type: 'function', exported: true, uname: 'juttle.modules.string.indexOf', arg_count: 2           },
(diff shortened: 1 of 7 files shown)


lib/compiler/semantic.js
```diff
@@ -172,6 +172,7 @@ class SemanticPass {
                 /* Functions (sorted alphabetically) */
                 length: { type: 'function', exported: true, uname: 'juttle.modules.string.length', arg_count: 1             },
                 concat: { type: 'function', exported: true, uname: 'juttle.modules.string.concat', arg_count: [0, Infinity] },
+                fromCharCode: { type: 'function', exported: true, uname: 'juttle.modules.string.fromCharCode', arg_count: [1, Infinity]},
```
rlgomes: Why not start from 0 ? Its a valid character code.

test/runtime/specs/juttle-spec/modules/string/from-char-code.spec.md
```diff
@@ -0,0 +1,28 @@
+The `String.fromCharCode` function
+============================
+
+Produces an error when passed an argument of incorrect type
+-----------------------------------------------------------
+
+### Juttle
+
+    emit -from Date.new(0) -limit 1 | put result = String.fromCharCode(null) | view result
```
rlgomes: I've seen this in a few other `juttle-spec` tests but is there a good reason why we tend to enumerate the negative tests and never add a positive test ? At least that way the coverage of that one line 27 above would happen unlike the current situation of this commit where that one line would be another "uncovered" line.

lib/compiler/semantic.js
davidvgalbraith: What? `arg_count` is the number of arguments, and `String.fromCharCode()` doesn't make sense.

test/runtime/specs/juttle-spec/modules/string/from-char-code.spec.md
davidvgalbraith: What? I have a `fromCharCode works as expected` test right there on line 15.

lib/compiler/semantic.js
davidvgalbraith: I guess it makes sense, we'd just return empty string.

davidvgalbraith:
@dmajda @bkutil

dmajda:
+1

PR #450, opened 2016-02-23
object: add the rest of the functions

davidvgalbraith:
- toString
- values

relates to https://github.com/juttle/juttle/issues/58

@dmajda @bkutil


dmajda:
+1 with the comment addressed

PR #451, opened 2016-02-23
array: add the rest of the functions

davidvgalbraith:
- [X] toString
- [x] concat
- [x] join
- [x] pop
- [x] push
- [x] reverse
- [x] shift
- [x] slice
- [x] sort
- [x] splice
- [x] unshift
- [x] lastIndexOf

### DIFF
--- a/lib/runtime/modules/array.js
+++ b/lib/runtime/modules/array.js
@@ -2,10 +2,20 @@
 
 /* Implementation of Juttle built-in Array module. */
 
+var _ = require('underscore');
 var errors = require('../errors');
 var values = require('../values');
 
 var array = {
+    concat: function() {
+        _.each(arguments, function(argument) {
+            if (!values.isArray(argument)) {
+                throw errors.typeErrorFunction('Array.concat', 'array', argument);
+            }
+        });
+
+        return Array.prototype.concat.apply([], arguments);
+    },
     length: function(array) {
         if (!values.isArray(array)) {
             throw errors.typeErrorFunction('Array.length', 'array', array);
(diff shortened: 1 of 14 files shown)


davidvgalbraith:
@dmajda @bkutil

lib/runtime/modules/array.js
```diff
@@ -28,7 +38,116 @@ var array = {
...
+
+        var spliceArgs = _.toArray(arguments).slice(1);
+        return array.splice.apply(array, spliceArgs);
+    },
+    toString: function(array) {
+        if (!values.isArray(array)) {
+            throw errors.typeErrorFunction('Array.toString', 'array', array);
+        }
+
+        return Array.toString(array);
```
bkutil: Do we want to use native representation? Maybe `values.toString` would be better (similarly for `Object`).

lib/runtime/modules/array.js
```diff
@@ -28,7 +38,116 @@ var array = {
...
+        }
+
+        return index;
+    },
+    pop: function(array) {
+        if (!values.isArray(array)) {
+            throw errors.typeErrorFunction('Array.pop', 'array', array);
+        }
+
+        return array.pop();
```
dmajda: On empty array, this would return `undefined`. We need to check for that and return `null` instead.

lib/runtime/modules/array.js
```diff
@@ -28,7 +38,116 @@ var array = {
...
+        }
+
+        return array.reverse();
+    },
+    shift: function(array) {
+        if (!values.isArray(array)) {
+            throw errors.typeErrorFunction('Array.shift', 'array', array);
+        }
+
+        return array.shift();
```
dmajda: On empty array, this would return `undefined`. We need to check for that and return `null` instead.

lib/runtime/modules/array.js
dmajda: Yes.

lib/runtime/modules/array.js
```diff
@@ -28,7 +38,116 @@ var array = {
...
+        }
+        return array.sort(function(a, b) {
+            return values.compare(a, b);
+        });
+    },
+    splice: function(array) {
+        if (!values.isArray(array)) {
+            throw errors.typeErrorFunction('Array.splice', 'array', array);
+        }
+        var begin = arguments[1];
```
dmajda: Any reason for not declaring `begin` in the argument list?

lib/runtime/modules/array.js
```diff
@@ -28,7 +38,116 @@ var array = {
...
+        if (begin !== undefined && !values.isNumber(begin)) {
+            throw errors.typeErrorFunction('Array.splice', 'number', begin);
+        }
+        var deleteCount = arguments[2];
```
dmajda: Same here.

lib/runtime/modules/array.js
dmajda: On empty array, this would return `undefined`. We need to check for that and return `null` instead.

dmajda:

PR #453, opened 2016-02-24
SemanticPass: coerce_var-related cleanups

dmajda:
Prep work for #419.


dmajda:
@bkutil

bkutil:
+1 - the test failure is unrelated, but could you double-check & re-run the tests?

dmajda:
Re-ran the tests (they are green now), merging.

PR #456, opened 2016-02-24
error: add toJSON so that message is included when serializing

demmer:
As a side-effect of changing from extendable-base to ES6 classes, and
due to the fact that the `message` property is not enumerable when
passed to the Error constructor, the message wasn't actually included
in the JSON representation of an Error.

To fix this, add a toJSON implementation that explicitly enumerates
message, code, and info. This does not include the name of the error
class, which makes things compatible with how they were in juttle 0.4
and earlier.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -12,15 +12,18 @@ var messages = require('./messages');
 // instead.
 class JuttleError extends Error {
     constructor(message, code, info) {
-        super(message, code, info);
+        super(message);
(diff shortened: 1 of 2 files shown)


demmer:
@dmajda: @go-oleg found this problem as part of updating juttle-service to be compatible with 0.5.0.
Once we merge this I think we should include in a 0.5.1 patch release.

dmajda:
+1

PR #457, opened 2016-02-25
validate options before init read or write proc

VladVega:
Lets validate options before initializing the the `read` or `write` proc.

Why? Because sometimes the `read` or `write` proc constructors need to use validated options.


VladVega:
@demmer @davidvgalbraith

davidvgalbraith:
+1

PR #460, opened 2016-02-25
Semantic: Kill “opts”

dmajda:
The high-level idea of this PR is to stop using the `opts` parameter to pass context around in the semantic pass and replace it by instance variables.

In general, passing context around using a parameter leads to unnecessarily verbose code and [can lead to performance problems](https://github.com/pegjs/pegjs/commit/cdeecf750f8a3f781459ed817815a537392993c1) when done properly (the context ends up being copied often so that child nodes don’t modify parents’ context).

In the particular case of semantic pass, the `opts` parameter was also used inconsistently, which made related code was brittle and bug-prone. Using an instance variable ensures the context is defined everywhere and is set in more controlled manner.

Note the instance variables and helper methods introduced in this PR probably aren’t the final form — there is a `Context` class hiding here :-) But we’ll see what we’ll need after the semantic pass gets converted to a visitor and split into multiple passes.

Prep work for #419.

### DIFF
--- a/lib/compiler/scope.js
+++ b/lib/compiler/scope.js
@@ -57,7 +57,7 @@ class Scope {
             return undefined;
         }
     }
-    is_mutable(name, opts, context) {
+    is_mutable(name, context) {
         var symbol;
 
         if (this.symbols.hasOwnProperty(name)) {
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
+1

Issue #433, opened 2016-02-18
Compiler can’t handle an empty sub


dmajda:
```
$ bin/juttle -e 'sub s() {}; s'
Error: Cannot read property 'head' of undefined
```

The correct result is an error message stating that a sub can’t be empty.

(This issue was tracked as PROD-7352 internally.)

dmajda:
@bkutil Assigning, you already looked into this at some point.

PR #470, opened 2016-02-26
semantic.js - disallow empty subs

bkutil:
Previously, defining an empty sub resulted in the following error from
the compiler:

```
$ bin/juttle -e 'sub s() {}; s'
Error: Cannot read property 'head' of undefined
```

As the sub didn't contain any nodes, it didn't return any which the
compiler could attach into the surrounding flowgraph.

A simple fix is to regard empty subs as an error. This is consistent
with handling of an empty program / flowgraph.

Fixes: #433

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -509,13 +509,17 @@ class SemanticPass {
         }
         this.exit_scope();
 
-        if (this.scope.type === 'top') {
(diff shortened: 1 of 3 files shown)


bkutil:
cc: @dmajda

dmajda:
+1

bkutil:
Fixed the error message. Will merge on green.

PR #471, opened 2016-02-26
Program cleanups

bkutil:
The high level goal of this patch set is to make `scheduler` accessible during procs construction.

Currently, procs access a `program` object and its properties, some of which (e.g assignments for `channels` and `now`) are generated by the compiler.

After the code produced by the compiler is evaluated, additional properties (`scheduler`) are assigned to `program`.

And as a part of `activate` step, the `program` object is further extended with `env` and `events` properties.

Instead of generating `program` in the compiler, then extending it in various ways, pass instance of `Program` class to the compiler and ensure that all properties required by procs are added during `Program`'s `constructor` call. This also makes them accessible during compilation phase.

### DIFF
--- a/lib/compiler/graph-compiler.js
+++ b/lib/compiler/graph-compiler.js
@@ -77,8 +77,7 @@ class GraphCompiler extends CodeGenerator {
         code = '(function(juttle, program) {\n';
         code += 'var views=[];\n';
         code += body;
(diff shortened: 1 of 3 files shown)


bkutil:
cc: @dmajda

dmajda:
Really nice set of cleanups, thanks!

bkutil:
Thanks, will merge on green.

PR #473, opened 2016-02-26
plugin runaway program detector

### LINKED ISSUE #245: Hook in runaway program detection
We have runaway program detection code in `lib/compiler/flowgraph/check_runaway.js` but we don’t use it. Let’s hook it in.

Note that because the absence of `-to` means different things in different sources now, the logic will need to be tweaked to accommodate that. We will need some way for an adapter to tell to the runaway detector what the absence of `-to` means for it. Alternatively, the sources could tell the runaway detector whether they will “run away” with given set of options (but this could become messy because sources are runtime concepts while the detector is a compile-time one).

rlgomes:
fixes #245

add in the runaway program detector and revive the unit tests along with
a small tweak to the pegjs grammar to correctly populate the sort proc
name

### DIFF
--- a/lib/compiler/flowgraph/check_runaway.js
+++ b/lib/compiler/flowgraph/check_runaway.js
@@ -3,58 +3,66 @@
 var _ = require('underscore');
 var errors = require('../../errors');
 
+function isRunaway(readNode, graph) {
 
-
-function is_runaway(read_node, g) {
-
-    function node_type(node) {
+    function nodeType(node) {
         // temp while grammar refactor is in progress; once complete we should be
         // able to always look up .name
         return node.name || node.type;
     }
 
-    function multiple_inputs(node) {
-        return g.node_get_inputs(node).length > 1;
+    function multipleInputs(node) {
+        return graph.node_get_inputs(node).length > 1;
     }
     // historical read
-    if (g.node_get_option(read_node, 'to') || g.node_get_option(read_node, 'last')) {
+    if (graph.node_get_option(readNode, 'to') || graph.node_get_option(readNode, 'last')) {
         return;
     }
 
     var outs;
-    var next_node = read_node;
+    var nextNode = readNode;
     while (true) {
-        outs = g.node_get_outputs(next_node);
+        outs = graph.node_get_outputs(nextNode);
 
         // Keeping things simple for now and only considering linear topologies
         // (which means we won't flag all possible runaways)
-        if (outs.length !== 1 || _.any(outs, multiple_inputs)) {
+        if (outs.length !== 1 || _.any(outs, multipleInputs)) {
             return;
         }
 
-        next_node = outs[0];
+        nextNode = outs[0];
 
-        switch (node_type(next_node)) {
+        switch (nodeType(nextNode)) {
             case 'batch':
                 return;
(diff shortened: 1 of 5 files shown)


rlgomes:
@demmer @dmajda runaway program detector plugged in! Now the question is if we want to somehow make this a default `fg_processor` ? Otherwise I have to fix juttle-engine to also add the `check_runaway` to its list of `fg_processors` when compiling.

rlgomes:
@davidvgalbraith you may also be familiar with this side of things.

lib/compiler/flowgraph/check_runaway.js
```diff
@@ -3,58 +3,70 @@
 var _ = require('underscore');
 var errors = require('../../errors');
 
+function isRunaway(readNode, g) {
```
davidvgalbraith: `g` is a lame variable name, let's use `graph`.

rlgomes: will do!

test/compiler/flowgraph/check_runaway.spec.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+var expect = require('chai').expect;
+var juttle_test_utils = require('../../runtime/specs/juttle-test-utils');
+var check_juttle = juttle_test_utils.check_juttle;
+
+describe('Runaway program detection', function() {
+
+    function check(juttle, is_runaway) {
+        it(`${juttle} // is ` + (is_runaway ? '' : 'not') + ' a runaway', () => {
```
davidvgalbraith: It's a little déclassé to put `it` inside a function like this, since it means you can't `it.skip` or `it.only` any of these tests.

rlgomes: this was the original way it was written and I was hoping someone else felt the same about it as I did. So I shall expand those tests out so it easier to skip/only and just read in general.

rlgomes:
@davidvgalbraith noicer ?

davidvgalbraith:
+1 and I think it should be a default processor as nobody really wants to run these scary runaways.

rlgomes:
@demmer @dmajda should we put this processor in the default list here instead: https://github.com/juttle/juttle/blob/master/lib/compiler/index.js#L86 ? That way its always on for all uses of the compiler (ie juttle bin, unit tests and juttle-service)

dmajda:
The commit in this PR does several things at once, please split it. Obvious groups of changes are:

  * Substantial changes in the detector (mostly `-to` handling)
  * Non-substantial changes in the detector
  * Hooking the detector in

As for the default processor list, let’s either go all the way (i.e. add all processors that make sense and don’t touch `fg_parameters` in the CLI and other places except tests) or keep the list empty. Anything else would be a bad API. The change should probably be done in a separate PR.

rlgomes:
@dmajda reworked things into separate commits looks better now ?

dmajda:
Thanks. Please further split the `-to` handling change and elimination of `node_type` form the first commit. These are the substantial changes I referred to and both should definitely stand alone. Elimination of `node_type` should probably be in one commit with adding `name` to `SortProc` because these are closely related.

With that, +1.

PR #474, opened 2016-02-26
location-stripper: make it an option in the compiler

demmer:
Instead of stripping locations from the AST only after the compiler
is completed, add a flag to the compiler to strip the locations right
after parsing.

This way locations can be stripped from all stages of the compiler,
including the parse tree, the flowgraph, as well as the build / compile
output.

When viewing the compilation output for these latter stages, the
generated code does contain `location: undefined`, but this is better
than containing the full location information by default.


demmer:
@dmajda This is an alternate approach to what I did in #183.

I personally prefer this one since it covers all the stages of compilation, including the Javascript code that's output from the build / compile stages.

dmajda:
Originally (when location information was introduced), I didn’t want to go this way in order to maintain invariant that locations are always present in the AST and other intermediate forms. The purpose was to avoid the possibility of introducing code conditional on their presence/absence.

I grepped through the code now and it looks like the only thing we do with locations is pass them around. I think it’s unlikely we’ll ever do anything else, so I guess there is no harm in stripping them in the compiler already. 

+1

demmer:
@dmajda I cleaned up the things you pointed out in the review, and also made another couple changes.

I made it so the module exports a function and not a class to simplify the call site, and updated the test to verify both `compile` and `compileSync`. Take another look when you get a chance but I assume this will be non-controversial.

dmajda:
+1

PR #475, opened 2016-02-26
Add new built-in adapters to top README.

mstemm:
Add "new" built-in adapters (stdio/http_server) to top level README.

@rlgomes @dmajda @mnibecker @go-oleg


demmer:
+1

PR #476, opened 2016-02-26
Adapter API: add parsers/serializers

davidvgalbraith:
https://github.com/juttle/juttle-s3-adapter/pull/8 `require`s these out of the Juttle lib, but the cool kids are pulling all their Juttle dependencies off the `JuttleAdapterAPI` these days.

@demmer @rlgomes

### DIFF
--- a/lib/adapters/api.js
+++ b/lib/adapters/api.js
@@ -10,7 +10,7 @@ let juttle_utils = require('../runtime/juttle-utils');
 // changes were made to the adapter API.
 //
 let AdapterAPI = {
-    version: '0.5.0',
+    version: '0.5.2',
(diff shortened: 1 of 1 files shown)


rlgomes:
+1 should have thought of this earlier

demmer:
+1 to the additions, but the API version should be `0.6.0`.

demmer:
Actually as we discussed in person, let's make this 0.5.2. Also can you re-open this PR against the 0.5.x branch since we'll release it as a bugfix since it's needed to enable the S3 adapter to be ported to the new API.

davidvgalbraith:
https://github.com/juttle/juttle/pull/478

PR #481, opened 2016-02-29
Make the “[]” operator work properly in rvalues

dmajda:
This PR adds tests that specify behavior of the `[]` operator in rvalues and adjusts the implementation to match. Specifically, it adds missing type checks into `ops.get`.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

PR #486, opened 2016-03-01
README: update to make juttle-engine more prominent

demmer:
Since the more complete installation of juttle will want to use
juttle-engine and not just the CLI, change the introduction and
installation instructions to refer to both options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,22 +3,31 @@
 [![Build Status](https://travis-ci.org/juttle/juttle.svg?branch=master)](https://travis-ci.org/juttle/juttle)
 [![Join the chat at https://gitter.im/juttle/juttle](https://badges.gitter.im/juttle/juttle.svg)](https://gitter.im/juttle/juttle?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-Juttle is an analytics system for developers built upon
-a stream-processing core and a
-[dataflow language](./docs/concepts/dataflow.md)
-that tightly integrates with
-[streaming visualizations](http://github.com/juttle/juttle-viz).
-The Juttle language lets you query,
-analyze, and visualize live and historical data from many different
-big data backends or other web services.
-You read data from a backend service, analyze it using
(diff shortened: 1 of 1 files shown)


demmer:
@dmajda @dmehra @mccanne What do you guys think about this change as an overall direction. I know it seems a bit strange to refer to juttle-engine here but it is the path we expect many users to go down so it seems appropriate to me.

README.md
```diff
@@ -30,7 +31,13 @@ To use Juttle as a command-line tool, the simplest approach is to install the ju
 $ npm install -g juttle
 ```
 
-You should now have a `juttle` executable in your path which you can use as follows:
+To use the full Juttle engine assembly, run:
```
dmajda: s/engine/Engine/

README.md
```diff
@@ -48,24 +55,15 @@ This produces:
...
 ```
 
-To use Juttle as a JavaScript library, install the juttle package locally as part of your project:
-
-```
-$ npm install juttle
-```
-
-See the [command line reference](./docs/reference/cli.md) for more information about how to configure use the Juttle CLI and how to configure.
+See the [command line reference](./docs/reference/cli.md) for more information about how to configure and use the Juttle CLI. See the [juttle engine README] for command line options and configuration instructions.
```
dmajda: [juttle engine README] — replace by real link?

dmajda:
Referring to Juttle Engine makes total sense to me.

dmehra:
+1 on making Juttle Engine prominent. When looking at the whole README, i noticed that we already had a section at the end "Visualizations with Juttle Engine", that has more detail. Let's link from the new upfront intro of Juttle Engine to that section for more info, so it's not kicking it there disconnected.

README.md
```diff
@@ -30,7 +31,13 @@ To use Juttle as a command-line tool, the simplest approach is to install the ju
 $ npm install -g juttle
 ```
 
-You should now have a `juttle` executable in your path which you can use as follows:
+To use the full Juttle Engine assembly, run:
```
demmer: I just tried and it actually (now) works fine.

demmer:
@dmajda @dmehra @mccanne I updated again, incorporating feedback from the first round. Please take another look.

dmajda:
+1

PR #488, opened 2016-03-01
Make the “.” operator work properly in rvalues (part 1)

dmajda:
Various preparations for making the `.` operator work properly in rvalues. Core changes are splitting lvalue and rvalue processing in the semantic pass (so they can be refactored independently) and preparing the compiler for the existence of dynamic non-computed member expressions (that is, expressions like `foo.bar` where `foo` is not a module).

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

dmajda:

PR #489, opened 2016-03-01
Fix function calls in filter expressions

### LINKED ISSUE #454: Function calls in filter expressions are optimized too much
The following Juttle always returns either 0 or 10 points:

```juttle
function f() {
    return Math.random();
}

emit -from :0: -limit 10 | filter f() < 0.5
```

This is because the `f()` call is considered compile-time optimizable and it is replaced by a constant value obtained by calling `f` at build time.

dmajda:
Fixes #454.


dmajda:
@bkutil

bkutil:
+1

PR #492, opened 2016-03-01
Make toString functions in Juttle modules use values.toString

dmajda:
Various `toString` functions in Juttle modules implemented their own string representation of Juttle values. Let’s replace that with `values.toString`, which is the canonical way.

(The reason why `toString` functions did their own thing is that there was no `values.toString` when they were written.)


dmajda:
@bkutil

bkutil:
+1

PR #493, opened 2016-03-01
Make the “.” operator work properly in rvalues (part 2)

dmajda:
In this part, `sa_MemberExpression` is cleaned up so that future change to support dynamic non-computed member expressions could be done easily. As a nice side effect, error messages in case of invalid module entity references (e.g. `m.c` when there is no module `m` imported) are improved.

Note that both lvalue and rvalue branches are affected by the changes (they are still kept identical). Splitting them in #488 was apparently slightly premature, but I realized that too late and didn’t want to go back and do a big rebase at the time.

Part of work on #419.

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -1436,11 +1436,12 @@ class SemanticPass {
     }
     sa_MemberExpression(node) {
         var symbol;
+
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
+1

PR #495, opened 2016-03-01
fix http error handling

### LINKED ISSUE #465: read http should not dump a full error
Entering a bogus stock identifier into the 'stock prices' demo results in the full 404 HTML page being appended into the demo (it is probably a part of the error message).

The http adapter read should guard against these errors flooding the output by replacing the message with a meaningful alternative. And/or for json, we could try parsing it first and if that fails, fallback to a pre-defined text.

rlgomes:
fixes #465

move the response to a `.body` field and placing the HTTP status message
into the actual `error.message` will make the UI's that display this
error not have to deal with large or ugly looking response payloads.

### DIFF
--- a/lib/adapters/http/read.js
+++ b/lib/adapters/http/read.js
@@ -271,8 +271,10 @@ class ReadHTTP extends AdapterRead {
                     res.on('end', function() {
                         var message = 'StatusCodeError: ' + res.statusCode +
                                       ' - ' + body.join('');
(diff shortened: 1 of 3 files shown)


rlgomes:
@bkutil

bkutil:
+1 - could you please also check that it fixes the demo page? Steps to reproduce are selecting `finance_data` example and entering an non-existent stock symbol.

dmajda:
+1

Issue #494, opened 2016-03-01
add notion of default processors used by various components by default


rlgomes:
This way we don't have to go plugging it into the compiler for the juttle cli, testing utilities and juttle-engine usage of the compiler.

rlgomes:
After some digging around this is a bit more complicated than I think we should fiddle with at this point. The `fg_processors` are a list of processors to check in order when compiling and if one of the uses differ a bit such as calling `implicit_views` with no argument vs an actual default implicit view then there's no easy way to merge those options with a default set of processors and keeping the order as it should. I'm thinking we should just leave things as they are and I can go add the `check_runaway` to the juttle-engine compiler options to make sure that gets checked. Opinions @dmajda @davidvgalbraith @demmer

dmajda:
@rlgomes Sorry but from your comment I can’t parse what is the problem.

I’d like to get rid of `fg_processors` as a compiler option (this is part of #61) and just have the same list of processors run every time. The only reason this wasn’t done already is the need to adapt tests which use `fg_processors`. I wrestled with them a bit in the past but gave up after some time.

As for `implicit_views`, we can either fix appended implicit view to `table` (which I’d prefer), or we can make it part of overall compiler options and make sure the processors receive these.

rlgomes:
@dmajda  Ok if we want to get rid of `fg_processors` then I can take care of that as the first commit to then set the fixed list of processors that run everytime. As for being able to change the implicit_view I'll figure out what can be done at that point.

dmajda:
@rlgomes That would be awesome!

I think the sequence of processors should be:

```javascript
[ implicit_views, check_runaway, views_sourceinfo, optimize ]
```

The idea is to first complete the graph, do checks, compute additional information, and finally optimize.

rlgomes:
+1 @dmajda that order of processors is what I was piecing together from seeing how these were being used across the board.  I think as for the tests that I'll make the other tests behave like the `check_runaway.spec.js` does where it doesn't run the compiler with the one and only processor in place and instead runs with the set of default processors in place verifying that specific processor is behaving correctly while running in the exact scenario it would along side other processors.

PR #496, opened 2016-03-01
remove unnecessary setting of table default

rlgomes:
Caught this through code inspection but we're already setting the default for an implicit view to be `table`  [here](https://github.com/juttle/juttle/blob/master/lib/compiler/flowgraph/implicit_views.js#L9) so no need for this code.


rlgomes:
@dmajda @demmer found while fixing #494

dmajda:
+1

PR #497, opened 2016-03-02
Make the “.” operator work properly in rvalues (part 3)

dmajda:
This commit adds tests that specify behavior of the `.` operator in rvalues and adjusts the implementation to match. Specifically, it allows using the `.` operator not only to access entities exported from
modules, but also properties of objects.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

dmajda:
Reworded per offline discussion with @bkutil, will merge on green.

PR #498, opened 2016-03-02
juttle-test-utils.js - ensure program is always deactivated after test run

bkutil:


bkutil:
cc: @dmajda @demmer

bkutil:
Fixed tests that relied on delayed deactivation but did not set it up and also the 'deactivate' tests themselves - they do not need to call deactivate explicitly.

dmajda:
+1

Issue #484, opened 2016-03-01
Optimized batch reduce -every doesn't emit marks


davidvgalbraith:
`reduce -every` recently changed to emit a mark after each point. Adapters that optimize `reduce -every` weren't changed accordingly, so optimized `reduce -every` doesn't emit any marks. It should emit marks. 

A proposal is to put a `batch` node into the flowgraph after a `read` that has optimized a `reduce`, since that will emit marks.

Issue #53, opened 2015-12-23
Support -every in all procs that care about marks


dmehra:
We should support -every in head, tail, sort, and any other proc that is mark-aware.

Prior conversation:

@dmajda: It would be nice to do this generically, so one would e.g. include some mixin into the proc class, implement few processing methods, and poof… the proc would be magically both batch-aware and -every-aware.

We're already a good part of the way there with periodic_funcs.js...

@welch: This may also require adding -reset, to control reaction to the epoch. 

We had a request for a simple way to `tail -f` a sparse stream. you could do that with `tail -every :10s: -reset false` (or maybe we want a -f option to tail that means these things?)
Alternatively, we'd want a tail reducer, to do all this with a single field, controlled by reduce -every.

@dmajda: POC was implemented in (private) https://github.com/jut-io/product/pull/8120

PR #499, opened 2016-03-02
optimizer: add a batch after reduce -every

davidvgalbraith:
Total hack! Best way to solve this problem though if we insist on the marks thing.

@demmer @dmajda @bkutil 

fixes https://github.com/juttle/juttle/issues/484

### DIFF
--- a/lib/graph.js
+++ b/lib/graph.js
@@ -11,21 +11,8 @@ class Graph {
         return this.built_graph.inputs;
     }
 
-    add_node(type, name) {
-        var index = this.get_nodes().length;
-        var new_pname = build_pname(index);
-
-        var node = { type: type,
-                     options: [],
-                     pname: new_pname,
-                     out: [],
-                     in: [],
-                     shortcuts: [] };
-        if (name) {
-            node.name = name;
-        }
-        this.built_graph.nodes.push(node);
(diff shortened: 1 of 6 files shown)


bkutil:
Q: wouldn't it be possible for the adapter, using the return value from its `read()` method, to tell the `read` that the output is batched and let `read` or `source` insert marks after each point? It might be cleaner than modifying the flow graph.

davidvgalbraith:
Yeah! Would prefer to do magic and have this be totally opaque to the adapters.

lib/graph.js
```diff
@@ -11,8 +11,8 @@ class Graph {
         return this.built_graph.inputs;
     }
 
-    add_node(type, name) {
-        var index = this.get_nodes().length;
+    add_node(type, name, index) {
```
dmajda: I’d slightly prefer to have a separate `insert_node(type, name, index)` method and to reimplement `add_node(type, name)` in terms of it. The idea is that appending and inserting at a specific position are two different things so they should be serviced by two different methods.

dmajda:
+1

Once #53 is resolved I think we should consider reverting this hack and using the same/similar method we’ll use to add `-every` to various other procs.

PR #503, opened 2016-03-02
Flowcontrol scheduler

bkutil:
This is a framework for future flow control scheduler. It is limited in several ways (which will be adressed in future PRs):

### DIFF
--- a/lib/runtime/flow-control.js
+++ b/lib/runtime/flow-control.js
@@ -101,7 +101,8 @@ class FlowControl {
 
     emitPoints(options) {
         _.each(this.procs, (proc, pname) => {
-            var points = proc.queue.splice(0, options.emitAll ? proc.queue.length : this.emitBatchSize);
+            var batchSize = this.getBatchSize(proc, options);
+            var points = proc.queue.splice(0, batchSize);
 
             if (points.length > 0) {
                 this.logger.debug(`${pname} emits ${points.length} points`);
@@ -113,6 +114,10 @@ class FlowControl {
                 }
             }
         });
+    }
+
+    getBatchSize(proc, options) {
+        return options.emitAll ? proc.queue.length : this.emitBatchSize;
     }
 }
 
--- a/lib/runtime/flow-control.js
+++ b/lib/runtime/flow-control.js
@@ -2,6 +2,7 @@
 
 var _ = require('underscore');
 var JuttleLogger = require('../logger');
+var JuttleMoment = require('./types/juttle-moment');
 
 class FlowControl {
     constructor(options) {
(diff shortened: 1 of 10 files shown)


bkutil:
Depends on https://github.com/juttle/juttle/pull/522

bkutil:
cc: @demmer @davidvgalbraith @dmajda 

I'll create/update a tracking issue with missing parts (as a checklist), then start working through them.

bkutil:
refs: #167

bkutil:
Rebased.

@demmer @davidvgalbraith bump. The state of the PR is equivalent to what we already have. And additional features of the flow control (see the checklist in #167) can be added gradually.

demmer:
@bkutil As we discussed, I added a high-level design writeup of the approach I had in mind to the top of the flow-control.js file. This represents what we talked about in person so please take a look and adjust / update.

bkutil:
@demmer I removed the logic of driving the flowcontrol via `setInterval`, could you please take a look?

I'm not sure about the correctness of ticks in the current implementation - they're driven by `enqueue`, while I think they should be independent of how often the source calls back (if their purpose is to act as a heartbeat and move the window forward while source is either lagging or silent for a long time).

Also, it is not clear how ticks will interact with emitting points when sources will be synchronized and emitting can be blocked until a read fills back the queue. But this is a separate problem to think about later.

lib/runtime/flow-control.js
```diff
@@ -1,5 +1,53 @@
...
+    to occur.
+
+    To accomplish this, the scheduler synchronizes the time range that's being
+    read from each source, potentially stalling faster sources in order for
+    slower sources to catch up. All processing in the middle of the flowgraph
+    occurs without interactions with the scheduler. To handle the case where a
+    source is faster than a sink, each sink is responsible for tracking the
+    amount of outstanding data it has buffered. If it exceeds a threshold, then
+    the sink signals the flow control scheduler to stop reading from all sources
+    and restarts things once the sink drainks.
```
davidvgalbraith: sink drainks

lib/runtime/flow-control.js
```diff
@@ -0,0 +1,136 @@
...
+        this.maxQueueSize = this.options.maxQueueSize || 100000;
+        this.defaultQueueSize = this.options.defaultQueueSize || 10000;
+
+        this.logger_name = 'flow-control';
+        this.logger = JuttleLogger.getLogger(this.logger_name);
+    }
+
+    register(proc) {
+        var pname = proc.params.pname;
+        if (!this.procs[pname]) { this.procs[pname] = proc; }
```
davidvgalbraith: `else` should we throw an error or is it OK?

PR #508, opened 2016-03-02
ast-visitor handle adapters

### LINKED ISSUE #507: error stripping locations from read / write procs
The following command produces an opaque error:

```
# bin/juttle --mode compile -e 'read file'
Error: Cannot read property 'apply' of undefined
```

The problem comes when the location stripper is trying to walk the AST, since the node generated for the read and write procs contains a child element (the adapter name) that has no AST node type but does have a location.

At the same time, the ASTVisitor should give more helpful errors when encountering problems in the AST.

demmer:
Update the grammar so that it generates AST nodes with `type: 'Identifier'` wherever we generate identifiers that contain locations.

This involved replacing instances of the `Identifier` rule with the functionally equivalent `IdentifierName`, and changing `IdentifierNameWithPos` to be the new `Identifier` which creates the appropriate AST nodes.

Fixes #507

### DIFF
--- a/lib/compiler/ast-visitor.js
+++ b/lib/compiler/ast-visitor.js
@@ -90,6 +90,7 @@ var NODE_CHILDREN = {
     ErrorStatement:           ['message'],
 
     /* Program Elements */
+    Identifier:               [],
     FunctionDef:              ['args', 'elements'],
     ReducerDef:               ['args', 'elements'],
     SubDef:                   ['args', 'elements'],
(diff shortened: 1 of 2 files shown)


demmer:
@dmajda

demmer:
@dmajda Building on your suggestion in https://github.com/juttle/juttle/commit/297fdd6ffc0a43686672bdd5b6e012e7ff30eaa7#commitcomment-16472568 I modified the grammar accordinfly. I updated the PR description with the details of what I changed.

I'll open a separate PR for the question of debug mode in the ASTVisitor.

dmajda:
+1

PR #511, opened 2016-03-03
read.js - emit initial tick immediately if from is specified

bkutil:
If there are parallel streams fed into fanin, the initial tick allows
fanin to start emitting points from a faster source earlier - it doesn't
have to wait until first points arrive.


bkutil:
cc: @demmer

bkutil:
Closing for now, will be handled by the flow control scheduler.

PR #512, opened 2016-03-03
values.spec.js - fix incorrect bind in "invalid comparison" test

bkutil:
The test verifying that typeErrorIncomparable was thrown for comparison
of two incomparable value types wasn't checking for an actual error
string.

This masked an incorrect bind and another problem with cyclical
dependenncy between `lib/runtime/errors.js` and `lib/runtime/values.js`.

Fix the bind and break the cycle by moving the error helper into
`values.js` - it is not used anywhere else.


bkutil:
cc: @dmajda

dmajda:
+1

PR #515, opened 2016-03-03
Refactor sa_CallExpression

dmajda:
Refactor `sa_CallExpression` in order to eliminate duplication and to prepare for moving bulk of the symbol-manipulating an checking code to `sa_Variable` and `sa_MemberExpression`.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

PR #522, opened 2016-03-03
Deactivate program after running it

bkutil:
Procs and scheduler depend on the program deactivation after juttle runs
and not doing that can lead to leaking timers.

Fix calls to `run_juttle` or `program.activate` that were missing a
corresponding `program.deactivate` call.

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -202,6 +202,9 @@ function perform_run(options) {
         })
         .then(function() {
             program.events.off();
+        })
+        .then(function() {
+            program.deactivate();
         });
     });
 }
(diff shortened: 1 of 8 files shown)


bkutil:
cc: @dmajda

dmajda:
Would it be possible to move the deactivation logic from `check_juttle` to `run_juttle` so that explicit deactivation is avoided? Or are there any cases where tests depend on programs *not* being deactivated?

bkutil:
Yeah, that would actually be a better approach. I'll patch `run_juttle` itself.

bkutil:
@dmajda done - it seems juttle spec was generating deactivation code. Is it still necessary? IMHO it is not needed anymore...

dmajda:
+1, but at some point we should move `deactivateAfter` to options.

bkutil:
Updated the `deactivateAfter` to be an option instead of a parameter.

PR #526, opened 2016-03-03
moved compiler tests into a more appropriate location

rlgomes:
noticed we had a duplicated runaway checker tests and that the location
of the view_sourceinfo.spec didn't match the correct lib/xxx path


rlgomes:
More work `en route` to fixing #494 @dmajda @demmer

rlgomes:
@davidvgalbraith not much to review other than eliminating a duplicate test and moving things into a better location.

dmajda:
+1 assuming you checked that the removed tests didn’t add any additional coverage

rlgomes:
@dmajda now that you said that they were using the `test` adapter and I think that's a better choice than the `stochastic` adapter I was using to generate "live" data. So I'll push a commit here to use the `test` adapter instead of the `stochastic` one.

PR #527, opened 2016-03-04
file: change read so that it emits points in batches

demmer:
Add an iterator so that the points are emitted in chunks instead
of waiting until the full file is parsed.

### DIFF
--- a/lib/adapters/file/read.js
+++ b/lib/adapters/file/read.js
@@ -7,6 +7,77 @@ var parsers = require('../parsers');
 var JuttleMoment = require('../../runtime/types/juttle-moment');
 var errors = require('../../errors');
 
+class Iterator {
+    constructor(adapter, from, to) {
+        this.logger = adapter.logger;
+
+        var stream = adapter.fetch();
+        this.adapter = adapter;
+        this.buffer = [];
+        this.eof = false;
+        this.error = null;
+        this.pending = null;
+
+        adapter.parser.parseStream(stream, (points) => {
+            points = adapter.parseTime(points, { timeField: adapter.timeField });
+            points = adapter.filterPoints(points, from, to);
+
+            this.logger.debug('parsed', points.length, 'points');
+            this.buffer = this.buffer.concat(points);
+
+            // go back to the event loop once before notifying out a response to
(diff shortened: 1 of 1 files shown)


demmer:
@davidvgalbraith witnessed this new code reading a file with 2GB of JSON data like a champ -- memory never even approached 100MB of RSS.

Needless to say the code on master shot up to 1.4GB of RSS and spun the core like mad.

rlgomes:
Would be nice if there was a way to test this had the same bug in the
original adapter write which I fixed and then we introduced the bug again.
Come to think of it I had a test that did something like verifying the
groups of points that were emitted. Did we remove that ? I'll have a look
later as I'm on my phone right now.

bkutil:
What do you think about benchmarking end-to-end performance for certain scenarios like `read -file '2GB.json' | (some proc like filter, reduce or even noop ) | write X`? The idea is to have something like `gulp perf` or `npm run-script -s perf` that would run a set of workload tests and measure cpu/memory consumption or iterations per second.

Even with stubbed out write method in `X` adapter, we could get a decent overview of how performant Juttle really is. I already did something similar in https://github.com/juttle/juttle-influx-adapter/pull/102 and used the data to optimise writes (the bottleneck was serialisation).

Ideally, the tests would be run nightly, but I'm not sure about Travis being a reliable platform for running these kind of tests and setting up our own dedicated box sounds fun, but would not probably be worth it (at the moment). So I think running them manually from time to time would be good enough for now.

demmer:
@rlgomes I played around with various ways of testing but I don't think there's going to be anything that's robust enough to put into Travis.

At a high level I like the direction that @bkutil suggested where we have a set of tests for large data sets that are run periodically.

However in checking the code coverage I realized we were missing some tests for how the adapter handles invalid files, so I added those in #534.

demmer:
@davidvgalbraith @mstemm Care to review?

mstemm:
I think it's ok. The structure of the Iterator shares a lot of properties with the http iterator so it'd be nice to find a way to share the overlap in some way.

davidvgalbraith:
+1

Issue #525, opened 2016-03-03
Field deference operator test failing


mstemm:
While working on a PR for changes to the http adapter, I noticed a probably unrelated test failure:

https://travis-ci.org/juttle/juttle/jobs/113442473

```1) Field dereference operator ambiguity Parses toplevel * as field dereference with whitespace before and no whitespace after:
      AssertionError: expected [ { count: 61 } ] to deeply equal [ { count: 62 } ]
      + expected - actual
       [
         {
      -    "count": 61
      +    "count": 62
         }
       ]
      
      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:485:19)
      at Assertion.ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at test/runtime/specs/juttle-spec/parser/field-dereference-operator-ambiguity.spec.js:40:50
      at bound (domain.js:287:14)
      at runBound (domain.js:300:12)
      at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
      at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
      at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
      at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
      at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/main/async.js:15:14)```

When manually re-running the test it passed, so I'd guess it's nondeterministic.

dmajda:
@bkutil Could you have a look?

bkutil:
Sure.

PR #529, opened 2016-03-04
Star operator ambiguity tests use test adapter

bkutil:
One of the tests was relying on stochastic adapter emitting a certain
number of points and was failing non-deterministically.

Either the bug is in reduce, or in stochastic adapter (which is known to
sometimes emit ooo points that are automatically dropped).

From the syntax POV, an adapter is an easy place to trigger the
ambiguity which we want to test, i.e. to get a number/duration literal
nearby a field dereference.

Change the adapter from stochastic to test and modify the irrelevant
test condition that was counting points to be stable.

Fixes: #525


bkutil:
cc: @mstemm @dmajda

bkutil:
The test failures are being fixed in #530.

demmer:
Looks good. I'm going ahead and merging.

Issue #528, opened 2016-03-04
Q: program deactivation logic


bkutil:
The tests currently delay the deactivation using `deactivateAfter` parameter, which shuts down the program after a certain period of time has passed.

If I get things correctly, the program should deactivate either after all reads/writes finish, or when the juttle process receives a signal.

How does the deactivation logic currently work?

bkutil:
cc: @davidvgalbraith @demmer

demmer:
I think the correct behavior is implemented in #532.

bkutil:
Makes sense - should/does the same logic apply when the program is not exercised via tests?

demmer:
When outside of tests, deactivate should happen when the user hits control-c or pushes a stop button or something like that.

PR #530, opened 2016-03-04
read.spec.js - fix http adapter tests requiring delayed deactivate

bkutil:


bkutil:
cc: @rlgomes

rlgomes:
Humm this doesn't seem right if the program doesn't finish on its own and needs to be deactivated after 250ms when it should simply stop and eof itself. Also the 250ms is just the right amount of time for the existing HTTP requests to finish up before the program is stopped which means if they take a bit longer these tests will fail since the deactivation would happen earlier than it should. So what is the adapter itself not doing to finish things up correctly aside from issuing the eof when it should ?

bkutil:
I think the problem is that program finishes too early, i.e. without waiting for the adapter to return the results - which causes the tests to fail. (The other tests in http read used the delayed activation as well, maybe for the same reason?)

But I don't know how the correct logic should look like - I asked about it in #528.

rlgomes:
@demmer @davidvgalbraith maybe you guys can shed some light on this, as the `deactivateAfter` feature was for live programs that wouldn't terminate and you wanted to run a juttle program for a period of time knowing that it would generate some data in that time frame that you could validate. Now `deactivateAfter` is being used on a program that isn't a long running program and it would be hard for anyone even writing tests to know if you should set the `deactivateAfter` or not ? Also this is going to generate a lot of flaky test behavior given the nature of guessing how much time should you wait to see data.

demmer:
This proposal is misguided -- the underlying problem is that unless `deactivateAfter` is passed as an option, we shouldn't be calling deactivate until the program completes. See #532 for the proper fix.

PR #524, opened 2016-03-03
Add support for new paging method using after query params.

mstemm:
While working on pulling TravisCI build statuses, I found I needed an
additional pagination method.

The pagination format TravisCI uses is:

https://api.travis-ci.org/repos/juttle/juttle/builds?after_number=<id>

after_number is related to a number field in the results:

{..., number: <id>, ...}

The idea is that you fetch the first page, find the last build number,
and then ask for another page with all builds after <id>.

To support that, add a new option for -pageUnit, "field". When
-pageUnit=field, an additional new option pageField is required, which
controls the field to use from the last returned point as the value to
put in the subsequent request's pageParam.

When pageUnit = field, the initial request does not contain a
pageParam query parameter.

Within the read adapter, the iterator keeps track of the last parsed
point. When it's time to move to a new page, the specified field from
that last point is used for the next page's pageParam.

Also require that when pageUnit "field", pageField is specified.

Add an example to the docs using TravisCI build status.

@rlgomes @VladVega

PR #531, opened 2016-03-04
More improvements to field based paging.

mstemm:
Address feedback from Mike on https://github.com/juttle/juttle/pull/524 after I merged:

 - Make pageField and pageStart exclusive, adding a new unit test to check.
 - Change the docs to only mention the exclusivity for pageField.

There's still a bad interaction with
https://github.com/juttle/juttle/commit/167935e3f5e66cc3862e890fc707efd4ab076a8e. Test
programs are calling deactivate immediately and relying on active
reads to signal that the program should be torn down. That doesn't
work for any of these paging based methods that could involve multiple
calls to read.


demmer:
looks great

PR #532, opened 2016-03-04
juttle-test-utils: by default deactivate after the program is done

demmer:
The change in #522 called deactivate immediately after running the
program unless deactivateAfter was set, which is the wrong semantics
since it will cause the programs to terminate early.

Instead, when deactivateAfter isn't set, wait until the program is
done before calling deactivate.

Also fix the call to Promise.delay -- the delay time is supposed to
be the only argument in bluebird promises.

### DIFF
--- a/test/adapters/http_server.spec.js
+++ b/test/adapters/http_server.spec.js
@@ -20,7 +20,7 @@ describe('read http_server', function() {
     });
 
     it('ingests simple json array', function() {
-        var programFinish;
+        var programFinish, program;
         var body = [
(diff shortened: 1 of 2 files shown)


demmer:
@bkutil @rlgomes @mstemm 

Note I cherry-picked this change and confirmed this fixes the problems on #531.

mstemm:
lgtm.

rlgomes:
+1

demmer:
@rlgomes @mstemm Just pushed a commit to fix the http server test failures.

PR #534, opened 2016-03-04
file: add missing tests for corrupt files and fix json parser

demmer:
I noticed that when reading a corrupt JSON file, the error message
contains [Object object]. Fix the parser to reject with err.thrown
and add a test to verify that we get expected error messages when
reading from corrupt files.


demmer:
The test actually depends on error semantics which changed in #527 -- In that branch, I now just propagate the underlying error (e.g. INVALID-JSON) rather than wrapping everything.

PR #535, opened 2016-03-04
default processors for compiler

### LINKED ISSUE #494: add notion of default processors used by various components by default

rlgomes:
fixes #494

set the list of default processors while making sure to pass the
`options.implicit_view` which allows the underlying `implicit_view`
processor to change its default behavior at runtime.

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -35,9 +35,6 @@ var URLResolver = require('../lib/module-resolvers/url-resolver');
 var FileResolver = require('../lib/module-resolvers/file-resolver');
 var resolver_utils = require('../lib/module-resolvers/resolver-utils');
 var compiler = require('../lib/compiler');
-var optimize = require('../lib/compiler/optimize');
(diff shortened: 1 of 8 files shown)


rlgomes:
@dmajda @demmer hopefully not to controversial as I couldn't figure out a nicer way to allow for setting of the `implicit_view` and maybe overloading the options passed to the compiler isn't right ? It does seem like a compile time type of setting. If this goes in then I'd have to also update the juttle-service to no longer set the default for the next release to be compatible with juttle.

rlgomes:
Also created #536 becuase I noticed we don't have actual spec tests for most of those processors.

dmajda:
I agree with the overall approach. Obviously this needs more work (fixing tests + there are some `fg_processors` instances left).

rlgomes:
@dmajda ok see if this works better

dmajda:

PR #540, opened 2016-03-05
reducers: change sigma to remove the bessel correction

### LINKED ISSUE #539: sigma reducer shouldn't use bessel's correction
The current implementation of the sigma reducer uses [Bessel's correction](https://en.wikipedia.org/wiki/Bessel%27s_correction) by using n - 1 in the divisor as opposed to n.

This can cause a discrepancy with other implementations of standard deviation, as noted in https://github.com/juttle/juttle-elastic-adapter/pull/114.

demmer:
Bessel's correction can remove bias in the standard deviation by
dividing by n - 1 instead of n. This causes discrepancies with other
calculations of standard deviation, so change the default behavior
to not apply the correction, but keep it as an optional second
argument to the reducer.

Also reduce potential error due to cancellation by shifting the computed
samples by the value of the first point.

Fixes #539


demmer:
@dmajda @davidvgalbraith

demmer:
With this change in place, the tests get the same results for optimized vs non-optimized standard deviation added in https://github.com/juttle/juttle-elastic-adapter/pull/114.

welch:
good call, that's the default behavior for numpy's std function (with an optional argument ddof -- delta degrees of freedom -- that  you can set to 1.0 if you want the bias correction)

still a shame this is named sigma instead of std....

demmer:
I originally thought about adding an optional reducer argument to control whether or not to apply bessel's correction -- i.e. `reduce sigma(value)` would be the normal behavior but `reduce sigma(value, true)` would do the correction. I decided it was a bit overkill so I just changed the default.

As to the name change, I'd be all for it. Will open another PR for that one.

lib/runtime/reducers.js
```diff
@@ -278,6 +278,7 @@ reducers.sigma = {
     arg_count: 1,
     fn: function(field) {
         var fld = field;
+        var K = null;
```
davidvgalbraith: `K` is not a very good variable name.

lib/runtime/reducers.js
```diff
@@ -288,14 +289,21 @@ reducers.sigma = {
...
                 }
             },
             update: function(pt) {
                 var v = pt[fld];
                 if (v !== undefined) {
-                    sum = ops.add(sum, v);
-                    ssum = ops.add(ssum, ops.mul(v, v));
+                    // Reduce error using the assumed mean method by shifting
+                    // the data based on the first value as suggested by:
+                    // https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
```
davidvgalbraith: Can we move this comment to just above `reducers.sigma = {` as part of a larger comment that describes the whole algorithm we use including that Bessel thing?

davidvgalbraith: Also does this relate to the Bessel thing or is it a separate change?

davidvgalbraith: Also, that link doesn't talk about the assumed mean, and the one that does, https://en.wikipedia.org/wiki/Assumed_mean, says
```
Its interest today is chiefly historical but it can be used to quickly estimate these statistics. There are other rapid calculation methods which are more suited for computers which also ensure more accurate results than the obvious methods.
```
which isn't exactly a ringing endorsement.

dmajda:
+1 to removing Bessel’s correction.

As for changing the algorithm — are the cancellations an actual problem with our typical datasets and the behavior of `double`? If so, I agree with @davidvgalbraith that the assumed mean algorithm might not be the best choice. Overall, this change feels like it needs a bit more research.

demmer:
@davidvgalbraith @dmajda @welch I removed the poorly justified changes related to the assumed mean which addressed most of the comments. I also noticed a bunch of places in the stats.juttle stdlib module where we were also using bessel's correction and I fixed those as well.

davidvgalbraith:
+1

PR #542, opened 2016-03-05
reducers: rename sigma to stdev

### LINKED ISSUE #541: rename the sigma reducer to stddev
It would be clearer and more consistent to use `stddev` as the reducer name that calculates standard deviation instead of `sigma`.

Noted in the discussion on #540.

demmer:
Keep an alias in place so existing juttle programs will still work,
and add a test to verify as such. We can deprecate that in some
future release.

Resolves #541.


demmer:
@dmajda @welch What do you guys think about this?

welch:
I like sd (R), std (numpy), stdev (matlab)
I didnt find any standard libraries using the stddev spelling

speaking of standard libraries, we can safely reject the 
apache.commons.math name getStandardDeviation as being too java.

dmajda:
+1 technically

As for the name, I don’t have a strong preference except that I think short names like `sd` or `std` are more suitable in mathematical/statistical context where the meaning would be immediately clear and the function would be used often. For our purposes I’d say longer names like `stdev` or `stddev` would work better. Of these two, the latter sounds more logical to me (I tend to parse `stdev` as `std` + `ev`), so I’d slightly prefer that.

welch:
if there is anything to be learned from the names out there, it's that there is no standard (and plenty of deviation). anything not involving camelcase is better than sigma.

demmer:
In the interest of being consistent I changed the PR to rename to `stdev` in the interest of using the more prevalent name.

I also added another commit that does similar renaming in the stdlib stats module.

@davidvgalbraith FYI

davidvgalbraith:
+1

demmer:
Rebased to fix conflicts. Will merge on green.

Issue #538, opened 2016-03-04
error in join after reduce -every


demmer:
Joining the output of `reduce -every` produces marks with no timestamps that are dropped and then crashes:

```
emit -from :0: -limit 10
| (
    reduce -every :5s: max(time);
    reduce -every :5s: min(time);
   )
| join
| view text -format 'raw' -times true

Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
Error: Cannot read property 'gt' of undefined [TypeError: Cannot read property 'gt' of undefined]
/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:774
        if (time.gt(this.last_mark)) {
                ^

TypeError: Cannot read property 'gt' of undefined
    at join.maybe_emit_mark (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:774:17)
    at join.emit_result (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:761:18)
    at join.advance_input (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:597:18)
    at join.mark_from (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:366:14)
    at join.consume_mark (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/join.js:349:14)
    at reduce_every.emit_mark (/usr/local/lib/node_modules/juttle-engine/node_modules/juttle-service/node_modules/juttle/lib/runtime/procs/base.js:216:25)
[...]

PR #543, opened 2016-03-07
reduce.js - emit initial mark on first batch start

### LINKED ISSUE #537: reduce -every should emit a mark at the beginning of the batch
The following two programs should produce identical output but they don't.

The `batch...reduce` version produces an initial mark:

```
emit -from :0: -limit 10
| batch :5s:
| reduce last(time)
| view text -format 'raw' -times true

--------------------------------------------------------------1970-01-01T00:00:00.000Z
{"time":"1970-01-01T00:00:05.000Z","last":"1970-01-01T00:00:04.000Z"}
--------------------------------------------------------------1970-01-01T00:00:05.000Z
{"time":"1970-01-01T00:00:10.000Z","last":"1970-01-01T00:00:09.000Z"}
--------------------------------------------------------------1970-01-01T00:00:10.000Z
==============================================================
```

While the `reduce -every` version does not:

```
emit -from :0: -limit 10
| reduce -every :5s: last(time)
| view text -format 'raw' -times true

{"time":"1970-01-01T00:00:05.000Z","last":"1970-01-01T00:00:04.000Z"}
--------------------------------------------------------------1970-01-01T00:00:05.000Z
{"time":"1970-01-01T00:00:10.000Z","last":"1970-01-01T00:00:09.000Z"}
--------------------------------------------------------------1970-01-01T00:00:10.000Z
==============================================================
```

We should be consistent here -- my instinct is that both should produce the initial mark.

bkutil:
To be consistent with `batch | reduce`, `reduce -every` should emit an
initial mark at the start of first batch.

Fixes #537, #538

### DIFF
--- a/lib/runtime/procs/reduce.js
+++ b/lib/runtime/procs/reduce.js
@@ -317,7 +317,8 @@ class reduce_every extends reduce_base {
         return 'reduce';
     }
     first_epoch(epoch) {
-        // don't run on the first epoch (called by advance)
+        // emit initial mark (before any points arrive)
+        this.emit_mark(epoch);
     }
     //
     // mark() and process() are from periodic. tick() from reduce_runner_funcs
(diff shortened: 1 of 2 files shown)


bkutil:
cc: @demmer @dmajda

dmajda:
+1

bkutil:
Fixed the comment, will merge on green.

PR #544, opened 2016-03-07
Move variable checks from Build and GraphCompiler to SemanticPass

dmajda:
`Build` and `GraphCompiler` checked that entities such as module references or functions were not used as variables. This check was semantic in nature so this PR moves it to the semantic pass. See the commit messages for more details.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1 on green

Issue #516, opened 2016-03-03
Inconsistency between “reduce -every ...” and “batch ... | reduce” on empty flows


dmajda:
```
$ bin/juttle -e 'emit -from :0: -limit 0 | reduce -every :s: count()'      
└───────────┴───────────┘
```

vs.

```
$ bin/juttle -e 'emit -from :0: -limit 0 | batch :s: | reduce count()'
┌──────────┐
│ count    │
├──────────┤
│ 0        │
└──────────┘
```

(This issue was tracked as PROD-6317 internally.)

dmajda:
@bkutil Something for you to look at.

bkutil:
IMHO negative, or non-positive value of `limit` should be disallowed.

PR #545, opened 2016-03-07
reduce.js - reduce -every returns aggregates for empty set of points

bkutil:
`reduce_batch` advances on `eof`, thus emitting the aggregate on empty flow,
i.e. when there aren't any upstream points, only an `eof`.

Make `reduce -every` follow the same pattern.

Fixes: #516

### DIFF
--- a/lib/runtime/procs/reduce.js
+++ b/lib/runtime/procs/reduce.js
@@ -311,6 +311,7 @@ class reduce_every extends reduce_base {
             // if -from or -to are specified, align on them by default.
             this.on = options.from || options.to ;
         }
+        this.after_first_epoch = false;
         this.init_warnings();
     }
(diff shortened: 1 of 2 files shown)


bkutil:
cc: @dmajda

dmajda:
Looks good modulo small comments.

However, it feels to me that with both `reduce` and `reduce -every` classes sharing the same base there should be a way to enforce the behavior on a higher level, rather than to override `eof` and duplicate the logic which already has to be present in `periodic_fanin` / `reduce_batch`. But I didn’t dig too deeply into this and it may also turn out that the only way is to generalize `periodic_fanin` to handle batches (which will likely happen as part of resolving #53).

dmajda:
+1

Issue #42, opened 2015-12-23
variable or sub redefinition does not produce an error


dmehra:
variable or sub or function redefinition should produce an error, but does not. 

When fixed, we can enable the two skipped tests here:

https://github.com/juttle/juttle/blob/master/test/runtime/basic-language.spec.js

dmajda:
This is slightly more general problem —beside subs it also concerns reducers and functions. Any of these can redefine a symbol of any type. Redefining any symbol by a const or a variable produces the expected result (an error).

Updating the title accordingly.

PR #549, opened 2016-03-07
throw an error when redefining a module import

### LINKED ISSUE #520: Importing modules under the same identifier twice should cause an error
The following code does not cause an error (assuming `m.juttle` exists and is a valid Juttle module):

```
import "m.juttle" as m;
import "m.juttle" as m;

emit -limit 1
```

(This issue was tracked as PROD-8442 internally.)

rlgomes:
fixes #520

quick fix and test to verify that we throw an error when the user
accidentally imports a module with the same locally scoped identifier.

### DIFF
--- a/lib/compiler/scope.js
+++ b/lib/compiler/scope.js
@@ -95,6 +95,21 @@ class Scope {
         }
         return varName;
     }
+    check_not_redefined(name, type, location) {
+        if (this.symbols.hasOwnProperty(name)) {
+            throw errors.compileError('REDEFINED', {
+                type: type,
+                name: name,
+                location: location
+            });
+        }
+    }
+    define_import(name, exports, location) {
+        this.check_not_redefined(name, 'import', location);
+        var symbol = { type: 'import', exported: false, exports: exports, d: true };
+        this.put(name, symbol);
+        return symbol;
+    }
     define_var(name, exported) {
         var runtimeVar = this.alloc_var(name);
         var symbol = { type: 'var', exported: exported, uname: runtimeVar };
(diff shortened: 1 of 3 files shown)


rlgomes:
@demmer @dmajda @davidvgalbraith easy enough fix to throw an error if you attempt to redefine an import. There doesn't seem to be any location information available at this point should that be a separate bug ?

dmajda:
@rlgomes Somewhat more systematic approach would be:

  1. Extract the redefinition check from `Scope#define_variable` into a separate method, e.g. `check_not_redefined(name, type)`.
  2. Introduce `Scope#define_import` (which would use the check) and use it instead of `Scope#put` in `SemanticPass#sa_ImportStatement`.

This way, it would be possible to reuse the code to solve #42 (by adding calls to `check_not_redefined` into other `define_*` methods).

(As a side note, the `Scope` class is not in the best shape and it needs some refactoring to get rid of repetitions. But I think it makes sense to do it only after #419 is finished because quite some methods will be dropped or simplified by then.)

rlgomes:
@dmajda looks better ? I can fix #42 in a follow up PR using the check_not_redefined. Also figured out how to pass in the location information so the CLI does the right thing and points at the correct location when reporting an error:

```
rlgomes@x230> ./bin/juttle -e "import 'm.juttle' as m1; import 'm.juttle' as m1; emit -limit 1"
<input>:1:26: 
   1:import 'm.juttle' as m1; import 'm.juttle' as m1; emit -limit 1
                              ^^^^^^^^^^^^^^^^^^^^^^^^
Error: redefined import: m1 (REDEFINED)
```

lib/compiler/scope.js
```diff
@@ -30,6 +30,12 @@ class Scope {
         return this.next ? this.next.get(name) : null;
     }
     put(name, info) {
+        if (name in this.symbols) {
```
davidvgalbraith: Can we just call `check_not_redefined` here?

rlgomes: Removing this as w'ere not suppose to do any checking here.

dmajda:
There is no need to pass `exported` and `d` to `define_import` as these have always the same value so they can be just set directly in `define_import`.

+1 with that change

PR #552, opened 2016-03-08
Fix moved variable checks

dmajda:
Variable checks in the semantic were incomplete. For example, the compiler correctly reported that `f` can’t be used as a variable in cases like this:

    function f() {
    }

    const c = f[0];

but not in cases like this:

    function f() {
    }

    const c = (f + 1)[0];

This was because the `in_member_expression` flag wasn’t reset back to `false` when processing expressions deeper in `MemberExpression`’s object.

This PR fixes that and also generalizes the flag a bit in anticipation of upcoming refactorings.

Part of work on #419 (really a fix to #544).

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -58,7 +58,7 @@ class SemanticPass {
         this.context = 'build';
         this.coercion_mode = 'none';
         this.lhs = false;
-        this.in_member_expression = false;
+        this.expr_mode = 'result';
         this.reducers = null;
         this.index = null;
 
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
Could you please add a test for the fixed case? +1 otherwise.

dmajda:
Right. Added few regression tests to `test/runtime/specs/juttle-spec/expressions/variable.spec.md` (the added cases always mirror the ones above them).

@bkutil Looks good?

bkutil:
Yep, thanks! :)

PR #554, opened 2016-03-08
Disallow filtering on time in static filter expressions

### LINKED ISSUE #551: Read filters on `time` shouldn't be allowed
We have `from` and `to` to set the time bounds of a program. Further filtering on the `time` field in a `read` filter is not very useful and suggests that somebody should be using `from` and `to` but doesn't know about them. Let's reject programs that filter on `time` in `read`.

dmajda:
Resolves #551.

### DIFF
--- a/lib/messages.json
+++ b/lib/messages.json
@@ -125,5 +125,6 @@
   "OPTION-VALUE-EXCEEDED": "option {option} exceeded limit of {value}{extra}",
   "INVALID-OPTION-COMBINATION": "option {option} can only be used with {rule}",
   "FILTER-FEATURE-NOT-SUPPORTED": "Filters do not support {feature} in this context.",
-  "HTTP-ERROR": "HTTP request failed with {status}: {message}"
+  "HTTP-ERROR": "HTTP request failed with {status}: {message}",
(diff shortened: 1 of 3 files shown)


dmajda:
@davidvgalbraith or @demmer

lib/messages.json
```diff
@@ -125,5 +125,6 @@
   "OPTION-VALUE-EXCEEDED": "option {option} exceeded limit of {value}{extra}",
   "INVALID-OPTION-COMBINATION": "option {option} can only be used with {rule}",
   "FILTER-FEATURE-NOT-SUPPORTED": "Filters do not support {feature} in this context.",
-  "HTTP-ERROR": "HTTP request failed with {status}: {message}"
+  "HTTP-ERROR": "HTTP request failed with {status}: {message}",
+  "CANNOT-FILTER-ON-TIME": "Cannot filter on \"time\". Use -to, -from, or -last instead."
```
demmer: I'd make it clear that this only applies to `read`... i.e. "Cannot filter on \"time\" in read. Use -to, -from, or -last instead."

demmer:
+1 with the message change.

davidvgalbraith:
+1

dmajda:
Updated the message, will merge on green.

PR #555, opened 2016-03-08
adapter config doc init

VladVega:
A forum for editing https://github.com/juttle/juttle/wiki/adapter-config-design/_edit

### DIFF
--- a/docs/adapters/configuration.md
+++ b/docs/adapters/configuration.md
@@ -1,5 +1,7 @@
 ## Goal
-Most simply and consistently across adapters, allow users to configure one or multiple connections in addition to a single path to the adapter module. 
+Allow users to configure one or multiple connections for an adapter in addition to a single path to the adapter module. Make this process most simple for the user and consistent across all adapters.
+
+Additionally when a Juttle program specifies `read adapter` or `write adapter`, a default connection must be selected to read from / write to. If there are multiple configured connections for the adapter, the user can select which one to read/write from by specifying the connection id as an option to read/write source.
 
 ## Options and Problems
 
@@ -15,13 +17,13 @@ Most simply and consistently across adapters, allow users to configure one or mu
 ```
 
 ### 2. An array of connection objects
-- Description: Each member of array has a unique `id` field so the user can indicate which connection to use.
+- Description: Each member of array has a unique `id` field so the user can indicate which connection to use. The first connection object is treated as the default one, regardless of the name of its `id` field (ie: it is not necessary to name it `id: "default"`). A juttle program `read <adapter>` that doesn't specify `-id 'key'` will read from this first instance of the adapter config.
 - Eval: Simple but we cannot assign `path` to the array unless we enforce some hacky rule to always insert it into the first element.
 - Example: 
 ```
 [
 {
-  id: "default",
+  id: "local",
   host: 'local',
   port: 5432
 },
@@ -72,6 +74,9 @@ path: "path"
   path: 'path'
 }
 ```
+
+## Another Default connection solution:
+Instead of the requiring the first connection object in the array to be the default, another option is requiring a `default` flag in one of the connection objects to signify the default selection.
 
 ## Adapter implementation
 Adapters will receive a single format (always an array, always with `id`s) and will use the first by default.
--- a/docs/adapters/configuration.md
+++ b/docs/adapters/configuration.md
@@ -1,0 +1,105 @@
+## Goal
+Most simply and consistently across adapters, allow users to configure one or multiple connections in addition to a single path to the adapter module. 
+
+## Options and Problems
+
+### 1. Single connection object.
+- Eval: Simple, can include `path`, but does not allow for several connections/environments (e.g. prod/staging/dev).
+- Example: 
+```
+{
+  host: 'local',
+  port: 5432,
+  path: 'optional'
+}
+```
+
+### 2. An array of connection objects
+- Description: Each member of array has a unique `id` field so the user can indicate which connection to use.
+- Eval: Simple but we cannot assign `path` to the array unless we enforce some hacky rule to always insert it into the first element.
+- Example: 
+```
+[
+{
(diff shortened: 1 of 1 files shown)


VladVega:
@dmehra

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+### 4. An object with with unique key `id`s
+- Description: An object with a key `path` and connection objects as values behind unique `id` keys.
+- Eval: Less simple due to nesting and can be ambiguous if you allow a single connection object as well.
+- Example:
+```
+{
+  default: {
```
dmehra: in this option, would one be required to have a nested object named "default", or would you still pick the top object as default one, even if it's named "bananas"?

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+```
+{
+  host: 'local',
+  port: 5432,
+  path: 'optional'
+}
+```
+
+### 2. An array of connection objects
+- Description: Each member of array has a unique `id` field so the user can indicate which connection to use.
```
dmehra: Let's add: "The first instance is treated as the default one, regardless of the name of its `id` field (ie: it is not necessary to name it `id: "default"`). A juttle program `read <adapter>` that doesn't specify `-id 'key'` will read from this first instance of the adapter config."

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
+## Goal
+Most simply and consistently across adapters, allow users to configure one or multiple connections in addition to a single path to the adapter module. 
+
```
dmehra: Let's clarify by adding something like this:

When a Juttle program specifies `read adapter` or `write adapter`, a default instance of that adapter will be selected to read from / write to. The default instance is the top one as specified in the juttle config file for that adapter type (or the only one, in case of one configured instance). If there are multiple configured instances of the adapter, the user can select which one to read/write from by specifying the instance id as an option to read/write source, in some way. A few alternative approaches exist.

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+## Options and Problems
+
+### 1. Single connection object.
+- Eval: Simple, can include `path`, but does not allow for several connections/environments (e.g. prod/staging/dev).
```
dmehra: what's the special thing about `path`? I don't know it.

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+    port: 5432
+  },
+],
+path: "path"
+}
+```
+
+### 4. An object with with unique key `id`s
+- Description: An object with a key `path` and connection objects as values behind unique `id` keys.
+- Eval: Less simple due to nesting and can be ambiguous if you allow a single connection object as well.
```
dmehra: What's the ambiguous thing? Afaict this is the approach used by https://github.com/juttle/juttle-elastic-adapter which has worked fine for me as a user so far. I didn't have to specify `path` because I don't know what that is, and the need never came up. The nesting did not strike me as difficult - I mean, our config is a JSON file, so nesting is expected... I personally find it simpler if it's just objects, not arrays of objects (one type of braces to align).

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+    port: 5432
+  }
+  prod: {
+    host: 'prod',
+    port: 5432
+  }
+  path: 'path'
+}
+```
+
```
dmehra: One more option was to explicitly specify "default: true" on an instance. Let's include it for completeness of discussion, but it's not my favorite, I'd rather use the top instance as default.

docs/adapters/configuration.md
```diff
@@ -0,0 +1,105 @@
...
+  {
+    id: "prod",
+    host: 'local',
+    port: 5432
+  },
+]
+```
+
+## Suggested Solution
+All options 1-3 should be accepted allowing the user to use the simplest form they need.
```
dmehra: As noted above, i think option 4 is what elastic adapter has been using, and it's been simple for me as the user.

docs/adapters/configuration.md
VladVega: `path` tells us where the module is. Doesn't have to do with the connection information. It is also `one` path per adapter where connection is `n` per adapter.

docs/adapters/configuration.md
VladVega: ES adapter uses option 1 and 2. So does sql.

docs/adapters/configuration.md
VladVega: Thats the problem. Choosing the default since there is no first option. You'd basically have to specify `id` in your juttle.

VladVega:
Tried to answer/include all of your concerns. Take another look? @dmehra

docs/adapters/configuration.md
dmehra: ah yes, you're right, i aligned the square brackets in my config in such a way that i stopped noticing them :)

docs/adapters/configuration.md
dmehra: let's add that into the doc, as it's not otherwise obvious that we have this per-adapter config option that is not instance specific.

VladVega:
@bkutil @demmer @davidvgalbraith good to go? I was going to work on this soon.

dmehra:
I was good with the doc, and proposed support for options 1-3.

bkutil:
+1

davidvgalbraith:
Let's have logic in juttle/juttle that transforms any of these to form 3 before passing it to the adapter, so all the adapters don't have to worry about all the formats.

VladVega:
@davidvgalbraith see line 81 on this doc. Adapters will receive one format - and yes there will be logic in juttle to convert all 3 formats available to the user into a standard format that adapters will receive.

PR #556, opened 2016-03-08
Allow ctrl-c to stop programs (politely)

### LINKED ISSUE #31: CLI can’t be interrupted by Ctrl+C in some situations

mstemm:
Allow ctrl-c to stop programs by adding a signal handler for SIGINT
that calls program.deactivate(). bin/juttle must also set standard
input to non-raw mode while the program is running so ctrl-c isn't
read by the cli directly.

Read already handles teardown properly in a limited way--after any
pending read has completed, the adapter is torn down and emits an
eof. An in-progress read is not stopped, though.

Emit required changes, though--add a teardown method that sets
a flag. The next time run() is called, it checks the flag and emits
eof() if set.

With these changes, a program won't be stopped until the next emit
happens or a read in progress has completed. If we want immediate
program cancellation, we'd have to make some additional changes:

 - Modify the scheduler to return ids from all calls to schedule so
   any pending callbacks could be cancelled, or do a wholesale
   teardown itself.

 - Start cancelling the read promise in read's teardown.

I think the current polite teardown is fairly nice as it results in a
relatively clean set of data and stil mostly meets the goal of
allowing program teardown, so I think we should try it first.

This fixes #31.

@demmer @bkutil @dmajda

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -18,6 +18,14 @@ process.stdout.on('error', function( err ) {
     }
 });
 
+var current_program = undefined;
+process.on('SIGINT', function() {
+    if (current_program) {
+        logger.debug('ctrl-c received, stopping current program');
+        current_program.deactivate();
+    }
+});
+
 var _ = require('underscore');
 var fs = require('fs');
 var path = require('path');
@@ -193,6 +201,13 @@ function perform_run(options) {
             console.warn('Warning: ' + msg);
         });
 
+        current_program = program;
+
+        // Change stdin to non-raw mode
+        if (process.stdin.isTTY) {
+            process.stdin.setRawMode(false);
+        }
+
         // Start the program.
         program.activate();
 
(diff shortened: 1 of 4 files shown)


bkutil:
Looks good. +1

dmajda:

Also, could you file issue(s) for the outstanding work necessary for immediate cancellation? Probably not of highest priority, but it should still be done at some point.

dmajda:
@mstemm Could you please squash review changes into original commits next time? Not doing so unnecessarily introduces “bad” code only to be immediately replaced by “good” one. Also, having commits with messages like “Address feedback from PR.” in the history is no good because one can’t tell anything about such commits and has to dive into the diff. This really hurts e.g. when preparing a release change log.

Thanks!

PR #559, opened 2016-03-09
parsers: add a columns parser for formatted text

demmer:
UNIX commands like ps, lsof, etc format their output as column
aligned text with certain quirks, like the fact that each line
doesn't always have every column, and sometimes the last column has
text that contains whitespace (e.g. the process args in ps).

To handle this, add a 'columns' parser type which implements the
various heuristics needed.


demmer:
@rlgomes @go-oleg @davidvgalbraith

lib/adapters/parsers/columns.js
```diff
@@ -0,0 +1,129 @@
...
+                    return;
+                }
+
+                if (!headers) {
+                    headers = this.findColumns(line);
+                    return;
+                }
+
+                let columns = this.findColumns(line);
+                let point = this.getValues(headers, columns, line);
```
rlgomes: When `getValues` hits an error on line 69, we'd end up getting an uncaught error here as we're not doing any try/catch around this to reject the promise so upstream the error can be reported correctly.

rlgomes: Also a test that triggers the error on line 69 would show the problem.

rlgomes:
One small issue in the uncaught error from `getValues` and a few more suggestions:
 1. I would push for not naming variables `c` and `h` as it makes things a bit harder to read
 2. lets add a test that validates an input with just headers emits nothing correctly (I think from code inspection this is the case but a test would be the best way to know for sure).
 3.The regex used to parse is `\S+` which is everything but `[ \t\r\n\f]` do we really want to split on carriage return, line feed and new line characters ? Maybe the regex `[^\t ]+` would be a better fit and We could use some tests with tabs themselves.
 4. I think this parsing lib could eventually be its own `npm` package that others may find useful as parsing the columned output of shell commands is a common thing and I couldn't find a good library on `npm`.

lib/adapters/parsers/columns.js
davidvgalbraith:
```new Promise(function(resolve, reject) {
    throw new Error('this is broken now')})
.then(function() { 
    console.log('then'); 
}).catch(function(err) { 
    console.error('caught error ' + err); 
});
```
prints `caught error Error: this is broken now`. So, this is fine.

davidvgalbraith: No, I'm totally wrong, didn't notice we're in an event handler here. So, this is not fine.

lib/adapters/parsers/columns.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        return result;
+    }
+
+    parseStream(stream, emit) {
+        let buffer = [];
+
+        function drain() {
+            let points = buffer;
```
davidvgalbraith: Do we want to check `if (points.length > 0)` here?

lib/adapters/parsers/columns.js
```diff
@@ -0,0 +1,129 @@
...
+            if (cidx === columns.length) {
+                result[h.value] = null;
+                hidx++;
+
+            // For the last column, grab the full text from the original line to
+            // handle cases like ps output where there can be spaces in the
+            // final column
+            } else if (hidx === headers.length - 1) {
+                result[h.value] = line.substr(c.start);
+                break;
```
davidvgalbraith: By my calculations you can put `hidx++` after all the `if/else`s here and get rid of the three `hidx++`s and this `break.

demmer:
@rlgomes @davidvgalbraith Updated in response to PR::

- I couldn't actually find a way to trigger the error case, so I just removed it altogether. Now if the columns don't line up as expected it just skips the column in all cases. I think this is guaranteed to terminate and to parse something.
- Added an error when the file is totally empty and verified that a file with only headers emits no points.
- Added the check for 0 length buffer and removed the unnecessary emits in those cases.
- Cleaned up the variable names and other stylistic suggestions.
- Along the way I consolidated the INVALID-JSON, INVALID-JSONL, and INVALID-CSV error codes to INVALID-DATA.

davidvgalbraith:
Can you trigger the error case with columns that overlap? Like
`abc def ghi`
`banana porcupine`

lib/adapters/parsers/columns.js
rlgomes: @demmer you didn't fix this one ;)

demmer: But it can't throw an error now so there's nothing to catch.

rlgomes: In my defense I haven't had a single cup of coffee yet, my bad.

demmer:
@davidvgalbraith the latest code parses this as:
```[
{"abc":"banana","def":"porcupine","ghi":null}
]
```

davidvgalbraith:
+1

PR #561, opened 2016-03-09
SemanticPass: Call sa_Variable in sa_CallExpression

dmajda:
When `callee` of a `CallExpression` node is a `Variable` node, process it using `sa_Variable` first (via `sa_expr`) and then investigate whether it represents a reducer/function or not. This means we lean on `sa_Variable` processing instead of doing our own in `sa_CallExpression`. This makes `sa_CallExpression` code more generic and the language more regular.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

PR #566, opened 2016-03-09
SemanticPass: Call sa_MemberExpression in sa_CallExpression

dmajda:
This is basically the same thing as #561, just for `MemberExpression`. This concludes refactoring of `sa_CallExpression`, which now doesn’t contain various symbol-manipulating and symbol-checking code that doesn’t belong there and relies fully on `sa_Variable` and `sa_MemberExpression` to handle this.

A nice side effect of the change is fixing several incorrect error messages.

Part of work on #419.

### DIFF
--- a/lib/compiler/scope.js
+++ b/lib/compiler/scope.js
@@ -184,12 +184,6 @@ class Scope {
     lookup_module_sub(mod, name) {
         return this.lookup_module_any(mod, name, 'sub');
     }
-    lookup_module_func(mod, name) {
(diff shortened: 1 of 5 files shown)


dmajda:
@bkutil 

The tests fail because of insufficient coverage, will figure out what to do with that (but it’s strange, this is refactoring and it *removes* lines).

dmajda:
Rebased to fix a small conflict, tests magically went green.

@bkutil, could you have a look?

bkutil:
+1

dmajda:

PR #567, opened 2016-03-09
Represent marks as objects

bkutil:
So far, marks were represented as a simple scalar value - time.

As a preparation for adding `interval` info into marks, change marks to
be objects, with `time` property containing the time stamp.

Modify all procs to create and access the `time` property, rename `time`
parameter of `consume_mark/emit_mark/mark` methods to `mark` and update
views to receive the complete mark object via the `view:mark` event, not
just time.

### DIFF
--- a/lib/runtime/procs/base.js
+++ b/lib/runtime/procs/base.js
@@ -163,8 +163,8 @@ class base {
             this.trigger('error', err);
         }
     }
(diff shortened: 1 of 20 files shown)


bkutil:
cc: @dmajda. Part of work on #123.

dmajda:
+1

PR #570, opened 2016-03-09
add env variable to control logging level

### LINKED ISSUE #569: cli should allow you to set the logger level
This would avoid seeing warnings/info's in the output when using the CLI to pull data that you wanted to further process downstream and outside of the juttle runtime.

rlgomes:
fixes #569

with the env variable LOGLEVEL you can pick the exact logging level
you'd like ie. `LOGLEVEL=off`, `LOGLEVEL=debug`, etc.

### DIFF
--- a/lib/cli/log-setup.js
+++ b/lib/cli/log-setup.js
@@ -10,7 +10,7 @@ function logSetup() {
     JuttleLogger.getLogger = log4js.getLogger;
 
     let levels = {
-        '[all]': 'info'
+        '[all]': process.env.LOGLEVEL === undefined ? 'info' : process.env.LOGLEVEL
     };
 
     // Handle node.js style debug configuration where DEBUG is a


rlgomes:
@mstemm @demmer @go-oleg good bad ? Our adapters are sometimes chatty writing info logs and what not and the CLI has no nice way to turn these off completely so I thought with this exposed it would be easier to not get info logs in the middle of your `view text` output on the CLI.

demmer:
I don't have any fundamental problem with this -- but if we are going to have chatty adapters at info level then I think the default cli level should be changed to warning.

demmer:
I also think a command line argument for -l <level> would be preferable to the environment variable.

rlgomes:
@demmer I'd be ok with simply making the default level `warning` instead and I can't do the `-l` thing easily since the logger is initialized way up at the top of the CLI bin file.

mstemm:
I'd say we should clean up the chatty adapters. I did that for aws/cloudwatch.

rlgomes:
I'll close this and file issues for the adapters that are chatty.

PR #571, opened 2016-03-10
compiler: Fix fg_processors processing when an empty array is passed

dmajda:
Currently, passing an empty array as a value of the `fg_processors` option causes the compiler to run the default processors. That shouldn’t be the case, it should simply run no processors. This commit fixes the compiler to do just that.


dmajda:
@rlgomes Not sure why you implemented the original behavior in #535, but I think this makes more sense.

rlgomes:
+1 yes you're right

PR #578, opened 2016-03-10
change the runaway error message to something more useful

### LINKED ISSUE #577: rework the runaway detector messages to be more useful
from #572 

@dmehra:
> By the way, the helpful message "add -to option to read?" is not so helpful when -to is already supplied, can we distinguish between the two runaway cases and emit different error messages?

rlgomes:
fixes #577

the old message was way off topic suggesting to use a `-to` when the
fact is that you simply need to make your read range time bound or batch
the underlying stream

### DIFF
--- a/lib/compiler/flowgraph/check_runaway.js
+++ b/lib/compiler/flowgraph/check_runaway.js
@@ -17,6 +17,13 @@ function isRunaway(readNode, graph) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


rlgomes:
@dmehra

rlgomes:
+1 for `"Make the read time bounded, or batch the stream"`

PR #573, opened 2016-03-10
SemanticPass: Refactor sa_FunctionProc

dmajda:
The idea is the same as in refactoring of `sa_CallExpression` (#515, #561, #566) — rely on `sa_Variable` and `sa_MemberExpression` for symbol processing and avoid duplicating it in `sa_FunctionProc`.

A nice side effect of the change is fixing an incorrect error message.

Part of work on #419.

PR #533, opened 2016-03-04
updating the coverage numbers

rlgomes:
Noticed that our coverage numbers had gone up a tiny bit so lets  not lose the coverage.

PR #579, opened 2016-03-11
gulpfile.js: Lower functions coverage threshold to 91%

dmajda:
Recent increase of functions coverage threshold to 92% (#533) broke at least one refactoring-only PR (#573) with another one suspected (#566). Let’s revert the change.


dmajda:
@bkutil

@rlgomes FYI

bkutil:
+1 for now to unblock.

PR #580, opened 2016-03-11
gulpfile.js: Lower statements & lines coverage threshold to 92%

dmajda:
Turns out that #579 wasn’t enough and #573 still [fails on Node 5.x](https://travis-ci.org/juttle/juttle/jobs/115269310). Let’s lower more numbers.


dmajda:
@bkutil 

@rlgomes FYI

bkutil:
+1

Issue #523, opened 2016-03-03
Change time produced by “reduce” to beginning-of-batch


dmajda:
Right now, `reduce` in a batched stream and `reduce -every` set the `time` field of emitted points to end-of-batch (minus epsilon). In practice, it would often be more useful to set `time` to beginning-of-batch. This issue covers changing `reduce` to do exactly that. In addition, emitted points should include an `interval` field containing a duration of the batch.

Note that with this idea implemented we should be able to get rid of epsilons.

bkutil:
Refs: #123 

If we want to pass the duration via `interval`, what should be its units?

In case of `reduce -every`, they can be the same as in the `-every` option value. The solution is not as clear for `batch | reduce`.

If we want `reduce -every` and `batch | reduce` expressions to be equivalent, we'd have to resort to some basic unit like milliseconds. But that seems to go against the purpose of introducing `interval` - IIUC, it is meant for the views to be able to display the 'natural' or 'calendar' intervals for batches.

cc: @demmer

demmer:
@bkutil `interval` should be the juttle duration used in `reduce -every` or `batch`. In the latter case, I think we would just need to include the interval along with the marks so it'll be present when reduce needs to generate the points.

bkutil:
> include the interval along with the marks so it'll be present when reduce needs to generate the points.

The marks are currently simple scalars (time). Does this mean we should extend them to be actual objects/hashes (similar to points), that would have `time` and `interval` properties? Alternatively, we could also modify the `consume_mark/emit_mark/mark` methods so they pass along `interval` as well as `time` as an argument and keep the values scalar.

The former is more dataflow-y and consistent, the latter is more ad-hoc, but would be easier to implement (procs using `time` argument wouldn't have to be modified to access `mark.time` instead).

In both cases, we should also consider:

1. Views can still receive `null` interval for artificially created marks (e.g. `join` proc) and should be able to handle this case. Or we could set a default interval, e.g. `1ms` in this case. Using the default is trivial to implement, as long as this is a reasonable value and views can handle it well.

2. When two parallel flows each contain marks and are joined via`fanin`, it de-duplicates them based on time. With intervals, it could either take them into account (e.g. use shortest/longest one), fallback to default (see above), or drop it.

dmajda:
First, the most important thing is to realize that marks are just “brackets” for batches. I believe we should not think in terms of manipulation of marks but in terms of manipulation of the underlying batches and *only then* translate that into operations on marks.

[...]

PR #581, opened 2016-03-11
Add interval marks

bkutil:
Aggregated points coming from `reduce -every` or `batch` operations
represent a quantity over a certain time interval.

This PR modifies `reduce_every` and `reduce_batch` processors so they
include this info into the points as an `interval` field.

This allows views to display these points differently, e.g. by including
the time interval.

Batch and reduce also insert the interval info into marks, so they
can be consumed by reduces further in the flowgraph - e.g. when a `batch`
is followed by several `reduce first()` processors.

Other processors simply forward the marks, including the interval info,
with the notable exception of join.


bkutil:
cc: @dmajda @demmer 

refs: #123

dmajda:
What is the plan for `join`?

Overall, I think it would help to split changes needed to add intervals to marks form the ones needed to add intervals to points.

bkutil:
The join logic is somewhat complicated, so the plan is:

1. figure out where and why join emits marks
2. figure out the correct `interval` value, if there is one.

Maybe Will would be able to help here.

We can track this as a separate issue, to unblock further work on handling interval info in the viz views.

bkutil:
Split the changes between points/marks updates, renamed `mark_interval` to `interval` and omitted the incorrect addition of `interval` info into `witness_timestamp`.

dmajda:
Looking at the change now, it seems like the `interval` parameter in `first_epoch` and `other_epoch` isn’t really necessary as in the periodic case the value is always `this.interval`. I’d remove it unless you have some deeper reason to keep it.

> omitted the incorrect addition of `interval` info in `witness_timestamp`

I think the right thing to do here is actually to add an interval with a duration between the last emitted mark and this one. Basically, we are emitting a shortened batch. Note we can bump into the same problems with calendarness as in fanin.

> We can track this as a separate issue, to unblock further work on handling interval info in the viz views.

Well, I think we want to get into the state where all marks (except the very first one) have a correct interval — only then will points coming from a batched reduce always have a correct interval. I’d avoid knowingly pushing any code which can potentially generate incorrect intervals. 

AFAIK, to get into that target state we need to:

 * resolve the `witness_timestamp` mark problem above
 * implement correct semantics for merging streams in fanin
 * fix `join`

I think this should be tracked together, either in #123 (because without this the logic won’t work in all cases) or separately.

bkutil:
> implement correct semantics for merging streams in fanin

Lets continue the discussion from https://github.com/juttle/juttle/issues/523#issuecomment-195288571 here:

> output stream of fanin contains marks from all input streams, combined. This creates batches of various sizes in the output stream
>> As for 2) (, the result of discussion with @demmer is that batches of varying length are OK, but we’ll forbid merging streams containing calendar-based batches with streams containing non-calendar-based ones. This will prevent any calendarness loss in batch interval lengths.

>> If this semantics proves to be too restrictive in practice, we can safely revert this position, allow the mixing, and accept possible calendarness loss.

(Calendarness loss means combining streams with calendar (i.e. interval where `interval.duration.months != 0`) and non-calendar intervals. E.g. `:month:` and `:5 days:`)

There are two additional things to consider:

1. In addition to the scenario above, calendarness loss can also occur if `-on` option is used with non-aligned calendar intervals. For example: `reduce -every :month: -on :day 10: ...; reduce -every :month: -on :day 20: | view...` can result in interval of days, which are not calendar. So we'd have to disallow this case as well.

2. It seems we'd have to distinguish batched & reduced streams from batched & non-reduced:

   * In streams coming out of a reduce, both marks and points contain the same interval value. When the interval fragments after `fanin`, the proc recomputes the interval for an incoming mark. However, now the interval may no longer correspond with the value in the reduced point. In this case, it seems that marks should be dropped - `fanin` would emit unbatched stream of reduced points.
  * If the input stream contains only non-reduced points and marks, `fanin` should merge incoming batches in the way we discussed, so the points can be reduced in those (possibly smaller/non-uniform) batches.
  
   As @dmajda pointed out, this approach creates two semantically different modes in the flow,but basically reflects the fact that we now have 'points in time' and 'points as intervals'.

@demmer thoughts?

dmajda:
We discussed the problems with @demmer. The results are:

  * We will just allow calendarness loss for marks in situations when it is unavoidable rather than forbid these situations. At the same time we will guarantee that fanin doesn’t lose calendarness at least in the following cases:

    * One input stream with calendar batches and zero or more unbatched streams.
    * One or more input stream with exactly aligned calendar batches.

    This work is covered by #123 and this PR (which will be updated to make sure all procs handle marks with intervals correctly and that fanin behaves as described above).

  * We won’t allow points with intervals to be “split” by marks. This is covered by #617.
  * We will make the `interval` field special. This is covered by #618.

bkutil:
The fanin now behaves as described. Also updated pace to forward marks, which I missed.

bkutil:
@dmajda can I have another review, please?

dmajda:
The commit message of 79353f5665b3ee594c6dd2f94f61ccee24872da8 (“periodic-fanin.js, reduce.js - emit marks with interval information”) no longer corresponds to changes in the commit, which now only contains changes that prepare for passing marks with intervals around.

It would also be good to move this commit before adapting `batch`.

lib/runtime/procs/reduce.js
```diff
@@ -152,8 +152,16 @@ class reduce_base extends periodic_fanin {
...
                 var fns = row.fns;
                 var pt = { };
-                // new points get the batch end timestamp
-                if (eob) { pt.time = eob; }
+                // new points get the batch end timestamp...
+                if (eob) {
+                    pt.time = eob;
+                }
+                // ...and interval info from marks coming from batch,
+                // every of `reduce -every`, or undefined for unbatched
```
dmajda: The comment is slightly misleading — in case of unbatched stream, the interval isn’t set to `undefined` but not set at all. This is especially confusing because `undefined` is not a valid Juttle value.

dmajda:
Please squash f8b8e6949034bc2e1d11dd6d132d605b307f697c to 79353f5665b3ee594c6dd2f94f61ccee24872da8.

dmajda:
I also think both `reduce` changes can just go in one commit. That is, squash f081b6ad02a76ae2a7bccd0ff77f249d37f2bd52 into c8d9c74107994cfcdfd2a2059b5bf11161779b5f.

lib/runtime/procs/join.js
```diff
@@ -801,28 +801,44 @@ class join extends base {
...
     emit_result(points) {
-        // emit the points, possibly bracketed by marks, and update emitter state.
+        // Emit the points, possibly bracketed by marks, and update emitter state.
+        // Mark interval will be the distance between group start and complete times.
         var leader = this.batched_leader() ;
         var join_time = this.outer ? this.ins[this.outer - 1].live_time() : this.max_live_time();
+        var interval;
+
+        if (leader) {
+            interval = leader.live().complete_time.subtract(leader.live().start_time);
```
dmajda: This means we will lose calendarness even in cases mentioned in my [earlier comment](https://github.com/juttle/juttle/pull/581#issuecomment-197864965). Can we avoid that?

lib/runtime/procs/join.js
```diff
@@ -801,28 +801,44 @@ class join extends base {
...
+        }
+
         if (leader) {
-            this.maybe_emit_mark(leader.live().start_time);
+            this.maybe_emit_mark(leader.live().start_time, interval);
         }
         if (points.length) {
+            // Marks around points, add interval info
+            if (interval) {
+                points.map(function(pt) { pt.interval = interval; });
```
dmajda: I don’t think it makes sense to add time to points in `join`. Unlike in `reduce`, they are not synthetic and they don’t cover the whole interval.

lib/runtime/procs/fanin.js
```diff
@@ -130,11 +131,22 @@ class Input {
 
         if (timelessInput) { return timelessInput; }
 
+        function hasInterval(a) {
+            return a.mark && a.interval;
+        }
```
dmajda: The name of the function is misleading — it checks more than it says. Either rename to something like `isMarkWithInterval` or just inline it.

lib/runtime/procs/fanin.js
```diff
@@ -130,11 +131,22 @@ class Input {
...
+
         return _.reduce(pendingInputs, (earliest, input) => {
             var inPt = input.first();
             var earliestPt = earliest.first();
 
             if (inPt.time.lt(earliestPt.time) || (inPt.mark && inPt.time.eq(earliestPt.time))) {
+                if (hasInterval(inPt) && hasInterval(earliestPt)) {
```
dmajda: It would be good to add a comment explaining why is this done and a test verifying the behavior.

lib/runtime/procs/pace.js
```diff
@@ -116,15 +116,17 @@ class pace extends fanin {
         }
     }
     mark(mark) {
+        mark.time = this.offset_time(mark.time);
```
dmajda: You should make a copy before modifying in order to avoid influencing other procs working with the mark (e.g. in a parallel stream).

dmajda:
Squash 54ac58e0c933ae1ee40a147e8db58e4b5b5a4d9c into 223294d0054ac2b26dbb5d946077784254a1fbdb? That would also resolve part of my [comment](https://github.com/juttle/juttle/pull/581/commits/223294d0054ac2b26dbb5d946077784254a1fbdb#r56816565) there.

dmajda:
Beside the squashing, it would be nice to regroup the PR so that changes that enable sending marks with intervals around go in first, followed by changes causing generating marks with intervals, followed by the change that causes generating points with intervals in `reduce`. That would follow the sequence of what is necessary for what, which would be easier to follow compared to current somewhat zig-zaggy approach.

lib/runtime/procs/join.js
bkutil: Thats a good question. I think it might be possible to copy information from mark into the join groups and then extract them after joining.

bkutil: Ok, that worked. Added it into `join` code and added a few tests.

bkutil:
@dmajda updated according to review suggestions

lib/runtime/procs/pace.js
```diff
@@ -116,15 +116,17 @@ class pace extends fanin {
         }
     }
     mark(mark) {
+        mark.time = this.offset_time(mark.time.clone());
```
dmajda: I meant to clone the mark, not the time.

lib/runtime/procs/join.js
```diff
@@ -801,11 +802,20 @@ class join extends base {
...
+        // Mark interval will be the distance between group start and complete times.
         var leader = this.batched_leader() ;
         var join_time = this.outer ? this.ins[this.outer - 1].live_time() : this.max_live_time();
+        var interval;
+
         if (leader) {
-            this.maybe_emit_mark(leader.live().start_time);
+            var live_leader = leader.live();
+            interval = live_leader.interval;
+            interval = interval || live_leader.complete_time.subtract(live_leader.start_time);
```
dmajda: When exactly does the situation that we don’t have `live_leader.interval` set arise?

test/runtime/specs/juttle-spec/juttle-procs-join.spec.md
```diff
@@ -1079,3 +1079,65 @@ marked complete and participates in early joins. Later joins get the second tabl
...
+    (
+        emit -limit 2 -every :1M: -from :0: | put first=true, m=count() | batch :1M:;
+        emit -limit 4 -every :1M: -from :0: | put second=true, n=count() | batch :2M:
+    )
+    | join -outer 1
+    | view result -times true -marks true
+
+### Output
+
+    { "time": "1970-01-01T00:00:00.000Z", "interval": "1M", "mark": true }
```
dmajda: This is odd, shouldn’t the first mark be interval-less?

test/runtime/specs/juttle-spec/juttle-procs-batch.spec.md
```diff
@@ -180,12 +180,28 @@ no bare numbers with the new -every option
...
-    emit -limit 2 -every :2s: -from Date.new(0)
-    | filter time >= Date.new(2)
+    emit -limit 2 -every :2s: -from :0:
+    | filter time >= :2:
     | batch :2s:
     | view result -marks true -times true
 
 ### Output
-    { "time": "1970-01-01T00:00:02.000Z", "mark": true }
+    { "time": "1970-01-01T00:00:02.000Z", "interval": "00:00:02.000", "mark": true }
```
dmajda: Same here — I think the first mark should be interval-less.

lib/runtime/procs/pace.js
bkutil: Oops, my bad. Will fix.

lib/runtime/procs/join.js
bkutil: When the mark in a live batched input does not contain an interval. This could happen for the first mark (group). But it should be possible to avoid this case by taking the interval from the 'end-of-batch' mark of the first batch, instead of the initial one.

test/runtime/specs/juttle-spec/juttle-procs-join.spec.md
bkutil: Ok, if we want to do this, we should be consistent across procs. I'll modify batch, reduce, fanin and join to emit the first mark without interval info. I'll put it into a separate commit, so it can be reverted later, if we decide against it.

PR #583, opened 2016-03-11
read: include the full stack trace in the debug log on a read error

demmer:
If an error occurs as part of the read implementation, it's useful
to have the full stack trace as context when logging at debug level.


demmer:
@go-oleg @mstemm @dmajda

go-oleg:
:+1:

PR #584, opened 2016-03-11
exposes the `-separator` option in the stdio and file adapter

### LINKED ISSUE #582: expose csv parser options at builtin adapter level
lets expose the csv parsing options such as delimiter, newline, etc.

rlgomes:
fixes #582

exposed the `separator` option for the csv parser from stdio and file
adapter and added necessary tests to verify we can read a TSV file (tab
separated value file)


test/adapters/parsers/csv.spec.js
```diff
@@ -12,7 +13,7 @@ describe('parsers/csv', function() {
     var pointsFile = path.resolve(__dirname, 'input/csv/points.csv');
     var invalidFile = path.resolve(__dirname, 'input/csv/invalid.csv');
 
-    it('can instantiate a csv parser', function() {
+    it('can instantiate a csv parser with', function() {
```
demmer: with what?

test/adapters/parsers/csv.spec.js
```diff
@@ -119,4 +120,56 @@ describe('parsers/csv', function() {
         });
     });
 
+    var pointFileTSV = path.resolve(__dirname, 'input/tsv/point.tsv');
+    var pointsFileTSV = path.resolve(__dirname, 'input/tsv/points.tsv');
+    var invalidFileTSV = path.resolve(__dirname, 'input/tsv/invalid.tsv');
+
+    _.each({
```
demmer: `_.each` of one entry seems like overkill, no?

For completeness we should test that it works fine when you explicitly give a `,` as the separator. Then you could parameterize the extension.

demmer:
In addition to the above comments this needs docs and an error check if the user passes `separator` when format is not csv.

rlgomes:
@demmer another quick look, addressed everything including errors when using separator with anything other than csv and testing the separator with `\t` and `,`.

demmer:
looks good. +1

rlgomes:
@demmer forgot to handle the docs will do another PR right now

PR #585, opened 2016-03-11
updated stdio and file adapter documentation

rlgomes:
with #584 we added the ability to specify a separator so lets document
how to use this option


rlgomes:
@dmehra missing documentation for changes in #584

dmehra:
+1 merging

PR #586, opened 2016-03-11
use node-grok from npm

### LINKED ISSUE #213: update node-grok dependency in juttle once maintainer pushes out npm package
issue opened on node-grok side: https://github.com/Beh01der/node-grok/issues/10

rlgomes:
fixes #213

we no longer have to depend on our own fork containing a fix and can
correctly rely on node-grok from npm


rlgomes:
@demmer we can finally use an official npm build here.

demmer:
+1

Issue #426, opened 2016-02-18
Allow expression statements


dmajda:
In Juttle grammar, we currently have the `AssignmentStatement` rule, which allows any expression at its RHS:

```juttle
foo = f() + 5 + bar++;
```

We should generalize this into `ExpressionStatement`, which would allow to use any expression as a statement (without a need to assign anywhere):

```juttle
f() + 5 + bar++;
```

The `AssignmentStatement` rule would disappear with this change and assignment statements would be handled by `AssignmentExpression` (which would be invoked via `Expression` from `ExpressionStatement`).

The purpose of `ExpressionStatement` is to allow execution of expressions purely for their side effects.

(This issue was tracked as PROD-6260 internally.)

## Steps

  * [x] Extend `AssignmentExpression` to support assignments like `a[0] = 5`
  * [x] Extend `AssignmentStatement` to support more operators
  * [x] Replace `AssignmentStatement` with `ExpressionStatement`

dmajda:
It turns out that `AssignmentStatement` and `AssignmentExpression` are two different kinds of assignments:

  * `AssignmentExpression` supports all the usual assignment operators like `+=`, but it doesn’t support assignments like `a[0] = 5`, only assignment into variables.
  * `AssignmentStatement` supports assignments like `a[0] = 5`, but it doesn’t support the usual assignment operators like `+=`, only `=`.

Thus, the steps to resolve this issue are:

  1. Extend `AssignmentExpression` to support assignments like `a[0] = 5`.
  2. Extend `AssignmentStatement` to support more operators (by implementing it in terms of `AssignmentExpression`).
  3. Replace `AssignmentStatement` with `ExpressionStatement`.

I’m adding these to the description.

dmajda:
I implemented steps described in my previous comment in #587, #588 and #589 (except that in step 2 I implemented multiple operator support for `AssignmentStatement` without rewriting it in terms of `AssignmentExpression` as this would effectively mean step 3 and I wanted the steps to be separate for clarity and easy referencing).

This is now done.

PR #587, opened 2016-03-14
Support indexed assignment in assignment expressions

dmajda:
This change brings assignment expressions closer to assignment statements which already support indexed assignment.

Part of work on #426.


dmajda:
@bkutil

bkutil:
+1.

Maybe it'd be better to expand the GH issue reference in the test to include

a) a full link
b) name and/or a brief summary of the referenced feature.

The number alone doesn't provide the required context.

dmajda:
@bkutil A reference to an issue number seems good enough to me but I added some more context. Better?

bkutil:
Ok.

PR #589, opened 2016-03-14
Support expression statements

dmajda:
Expression statements replace more limited assignment statements.

Part of work on #426.


dmajda:
@bkutil

bkutil:
+1 on green

PR #590, opened 2016-03-14
Generalize function call grammar

dmajda:
Instead of treating function calls specially in the grammar, handle them similarly to `[]` and `.` operators — that is, allow any expression before `()` and leave it to the semantic pass to decide whether this expression is callable or not.

Part of work on #419.

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -1261,8 +1261,7 @@ class SemanticPass {
                     });
             }
         } else {
-            throw errors.compileError('NOT-A-FUNCTION-OR-REDUCER', {
-                name: this.function_name(node),
+            throw errors.compileError('CANNOT-USE-AS-FUNCTION-OR-REDUCER', {
                 location: node.location
             });
         }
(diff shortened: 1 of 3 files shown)


dmajda:
@bkutil and @demmer please both let me know what you think about this.

The gist of the change is making the decision about what is a “callable” not in the grammar (by limiting function/reducer call syntax) but in the semantic pass (by analyzing the expression and the symbols if refers to).

The change seems logical to me because some level of semantic checking on callables is necessary anyway (e.g. in `foo.bar()` one must check that `foo` is a module and that it exports a function/reducer `bar`) and actually all the necessary code was already in the semantic pass (introduced during last week’s refactorings). The change also makes the grammar more regular and it unifies handling of all entities exported by modules. Moreover, it will make introduction of first-class functions/reducers easier one day.

The disadvantage is that the grammar represents what is allowed less exactly and that one error message ends up a tiny bit worse because it can’t make assumptions about the expression structure (see the first commit).

bkutil:
IMHO sounds good I can't think of any examples where checking for 'callability' in the semantic pass instead of the grammar would cause trouble and the regularity of handling `.`, `[` and `(` is nice...

bkutil:
(The comment above was meant as a +1)

dmajda:
OK, will merge now.

@demmer If you care and don’t think this is a good idea (I think it is), let me know.

PR #591, opened 2016-03-14
Add support for importing using relative paths

### LINKED ISSUE #513: Distinguish system and user module imports
Issues like #438 reveal a need for two kinds of file-based module imports:

  * **System imports** — would load predefined modules from standard locations (think stdlib or possibly adapters)
  * **User imports** — would load user modules from wherever users put them

This distinction corresponds to `#include <foo.h>` vs. `#include "foo.h"` in C/C++, or `require('foo')` vs. `require('./foo')` in Node.js.

Our module system is similar to the one used in Node.js in that it identifies modules using strings (as opposed to more restricted ways, such as identifiers in Python or hierarchical packages in Java). It feels natural to adopt Node.js’s method of distinguishing system and user imports by module path prefix.

`FileResolver` should be adapted to treat modules whose path begins with `/`, `./`, or `../` as user modules and load them directly without using module load path. Other modules should be treated as system modules and handled in the same way as today.

Once this is implemented, we should make sure that the current directory and stdlib directory are no longer added to the module load path.

See also [Node.js module resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together).

mstemm:
See commits for specific changes.

I'd like feedback on the technique of modifying the ast to handle the case of potentially duplicate import statements with the same relative pathname. Not sure whether this is bad practice or not.

One open question--do we want imports of standard library modules to have a 'stdlib' prefix? Currently, the stdlib directory is in the set of search paths so one imports those files with a 'import "random" as random" rather than 'import "stdlib/random". I think I'd have to move the lib/stdlib directory somewhere as a part of this change--we don't want lib/ as a search path.

@demmer @dmajda 

This fixes #513.

### DIFF
--- a/docs/concepts/programming_constructs.md
+++ b/docs/concepts/programming_constructs.md
@@ -159,10 +159,27 @@ location, then in another program you can import it as a module, using the
     import "<module_path>" as local_name;
 ```
 
-How the `<module_path>` is resolved depends on the specific capabilities of the
-application in which Juttle is running. For the Juttle CLI, it can be either a
-relative or absolute filesystem path or a URL to a remote Juttle file containing
-exported code.
+`<module_path>` can either refer to a *user import* or a *system
+import*.
+
+User imports are references to individual modules. User imports can
+either be a url, a full pathname starting with `/`, or a relative
+pathname starting with `./` or `../`. All other paths are considered
+system imports.
+
+When resolving user imports with relative pathnames, the pathname is
+always applied relative to the file doing the import. See the below
+examples for details.
+
+System imports load predefined modules from standard
+locations. Examples of system imports are the
+[Juttle Standard Library](./juttle_standard_library.md). For system
+imports, module_path is a path suffix that is combined with one of the
+standard locations to result in a full pathname.
+
+For both user and system imports, when referring to files, a `.juttle`
+extension will be added if not present, and when referring to
+directories, an `/index.juttle` will be added.
 
 Once it is resolved, the import statement pulls all exported symbols from the
 specified Juttle module and makes them available under the specified local
@@ -216,18 +233,45 @@ emit
 | view table;
 ```
 
-_Example: import by filename_
+_Example: user import by filename_
 
 This example assumes that the program with exports is saved at path
-`docs/examples/concepts/export_module.juttle`.
+`docs/examples/concepts/export_module.juttle`, and that the program
+being run is at `docs/examples/concepts/import_module.juttle`.
 
 ```juttle-no-syntax-check
 // This program is runnable from CLI from the juttle repo root,
 // juttle docs/examples/concepts/import_module.juttle
 
-import 'docs/examples/concepts/export_module.juttle' as my_module;
+import './export_module.juttle' as my_module;
 emit
 | my_module.stamper -mark 'test'
+```
+
+_Example: modules importing modules with relative paths_
+
+This example uses the following files:
+
+`docs/examples/concepts/import_module_nested.juttle`:
+```juttle-no-syntax-check
+// This program is runnable from CLI from the juttle repo root,
+// juttle docs/examples/concepts/import_module_nested.juttle.
+
+import "./shared/module.juttle" as mod;
+
+emit -limit 1 | put val=mod.myvalue | view table;
+```
+
+`docs/examples/concepts/shared/module.juttle`:
+```juttle-no-syntax-check
+import './second_module.juttle' as mod2;
+
+export const myvalue=mod2.value;
+```
+
+`docs/examples/concepts/shared/second_module.juttle`:
+```juttle-no-syntax-check
+export const value = 10;
 ```
 
 _Example: import by URL_
@@ -244,6 +288,22 @@ import 'https://gist.githubusercontent.com/jut-test/273396ac4efcc838687b/raw/dde
   as this_module;
 emit
 | this_module.stamper -mark "test"
+```
+
+_Example: using system import_
+
+This example loads the file `random` from the juttle standard
+library. Note the use of the implied `.juttle` extension that is added
+when resolving.
+
+```juttle-no-syntax-check
+// This program is runnable from CLI from the juttle repo root,
+// juttle docs/examples/concepts/import_module_from_stdlib.juttle
+
+import "random" as random;
+emit -limit 3 -from :0:
+| put pure = random.normal(0, 1)
+
 ```
 
 Subgraphs
--- a/docs/concepts/programming_constructs.md
+++ b/docs/concepts/programming_constructs.md
@@ -192,7 +192,7 @@ You can reference the imported symbol(s) like this:
 ```
 
 -  local_name is the local name from the import command above.
--  symbol_name is the name of the exported subgraph, constant, function, 
+-  symbol_name is the name of the exported subgraph, constant, function,
    or other code fragment from the imported program file.
 
 Module Behavior
@@ -215,7 +215,7 @@ When defining or importing modules:
 
 _Example: export from module_
 
-This juttle module exports a function, a subgraph, and a constant. 
+This juttle module exports a function, a subgraph, and a constant.
 
 ```juttle
 export const pi = 3.14;                                // exported constant
@@ -223,8 +223,8 @@ const not_exported = 2;                                     // not exported
 export function double(x) { return x * not_exported; } // exported function
 
 export sub stamper(mark) {                             // exported subgraph
-	put stamp = mark
-	| put stamp2 = mark
+    put stamp = mark
+    | put stamp2 = mark
 }
 
 // top-level flowgraph is not exported
--- a/lib/module-resolvers/file-resolver.js
+++ b/lib/module-resolvers/file-resolver.js
@@ -8,8 +8,7 @@ var logger = require('../logger').getLogger('file-resolver');
 
 function search_paths() {
     var env = process.env;
-    var cwd = process.cwd();
-    var paths = [ cwd ];
+    var paths = [];
     if (env.JUTTLE_MODULE_PATH) {
         paths = paths.concat(env.JUTTLE_MODULE_PATH.split(':'));
     }
--- a/lib/module-resolvers/file-resolver.js
+++ b/lib/module-resolvers/file-resolver.js
@@ -4,6 +4,7 @@
 var fs = require('fs');
 var path = require('path');
 var Promise = require('bluebird');
+var logger = require('../logger').getLogger('file-resolver');
 
 function search_paths() {
     var env = process.env;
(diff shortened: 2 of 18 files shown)


lib/parser/index.js
```diff
@@ -82,7 +83,14 @@ function parse(juttle, options) {
...
         return Promise.resolve(module.source)
             .then(function(juttle) {
-                return doParse(juttle, _.extend(options, { filename: module.name }));
+                // If the module has a filename property, it was
+                // passed back from the resolver as the full path to
+                // the file. Use it instead.
+                let filename = module.name;
+                if (_.has(module, 'filename')) {
+                    filename = module.filename;
+                }
```
dmajda: I think it would be cleaner if the protocol was that a resolver always sets `module.filename`. Then we could rely on that here.

lib/parser/index.js
```diff
@@ -97,10 +105,9 @@ function parse(juttle, options) {
 
     options = processOptions(options, defaultResolver);
 
-    return parseModule({ source: juttle, name: 'main' })
+    return parseModule({ source: juttle, name: 'main', filename: (options.filename ? options.filename : 'main') })
```
dmajda: `options.filename` is always set in `processOptions` so it should be enough to do `filename: options.filename` here.

lib/parser/index.js
```diff
@@ -97,10 +105,9 @@ function parse(juttle, options) {
...
         .then(function() {
             asts.main.modules = buildModuleDefs(asts);
-
```
dmajda: Unrelated change. (Moreover one I disagree with — in general I like to always have an empty line before a `return`.)

lib/parser/index.js
```diff
@@ -159,7 +167,15 @@ function parseSync(juttle, options) {
...
         }
 
-        var ast = doParse(module.source, _.extend(options, { filename: module.name }));
+        // If the module has a filename property, it was
+        // passed back from the resolver as the full path to
+        // the file. Use it instead.
+        let filename = module.name;
+        if (_.has(module, 'filename')) {
+            filename = module.filename;
+        }
```
dmajda: Same as above — rely on the resolver.

lib/parser/index.js
```diff
@@ -171,7 +187,7 @@ function parseSync(juttle, options) {
 
     options = processOptions(options, defaultResolver);
 
-    parseModule({ source: juttle, name: 'main' });
+    parseModule({ source: juttle, name: 'main', filename: options.filename ? options.filename : 'main' });
```
dmajda: Same as above: `options.filename` should be enough.

lib/module-resolvers/file-resolver.js
```diff
@@ -38,36 +40,75 @@ class FileResolver {
...
-            var filename = paths[k] + '/' + module_path;
-            if (self._is_file(filename)) {
-                try {
-                    return fs.readFileSync(filename, 'utf8');
-                }
-                catch (e) {
-                    // continue looking for a module we can successfully load
+
+        if (module_path.indexOf('./') === 0 ||
+            module_path.indexOf('../') === 0) {
```
dmajda: `module_path.startsWith('./') || module_path.startsWith('../')`

lib/module-resolvers/file-resolver.js
```diff
@@ -38,36 +40,75 @@ class FileResolver {
...
+
+            // module_path is a relative path and resolve it relative
+            // to the base directory of the file doing the import, or
+            // the current working directory.
+            let dir = process.cwd();
+            if (path.isAbsolute(importer_path)) {
+                dir = path.dirname(importer_path);
+            }
```
dmajda: Why is this logic needed? When, in practice, can `importer_path` be relative?

lib/module-resolvers/file-resolver.js
```diff
@@ -38,36 +40,75 @@ class FileResolver {
...
+            try {
+                return {
+                    source: fs.readFileSync(filename, 'utf8'),
+                    filename: filename
+                };
+            }
+            catch (e) {
+                return false;
+            }
+        } else if (path.isAbsolute(module_path)) {
```
dmajda: I think this branch could be merged with the previous one as `path.resolve` should handle an absolute path in its arguments.

lib/module-resolvers/file-resolver.js
```diff
@@ -38,36 +40,75 @@ class FileResolver {
...
+                if (self._is_file(filename)) {
+                    try {
+                        return {
+                            source: fs.readFileSync(filename, 'utf8'),
+                            filename: filename
+                        };
+                    }
+                    catch (e) {
+                        // continue looking for a module we can successfully load
+                    }
```
dmajda: It would be nice to split the relative module lookup and system module lookup into separate branches.

docs/concepts/programming_constructs.md
```diff
@@ -175,7 +191,7 @@ You can reference the imported symbol(s) like this:
 ```
 
 -  local_name is the local name from the import command above.
--  symbol_name is the name of the exported subgraph, constant, function, 
+-  symbol_name is the name of the exported subgraph, constant, function,
```
dmajda: Put any formatting/whitespace changes into different commit(s) please.

dmajda:
I only reviewed the first commit as it is obvious some non-trivial changes will be needed there already.

I don’t like the `name` vs. `filename` duality and related conditions scattered around the code base at all. Basically, I think the resolver should return resolved path in the `name` property and that is what the module should be known/indexed as everywhere. Renaming this to `filename` would make it more clear, but it should be done in a separate commit as a prep change.

Overall, it would be good to split the first commit into multiple ones. At least modifying the resolver interface, adapting parser internals to handle returned module paths correctly, and teaching the file resolver to do relative imports seem like separate steps. Having them all intermingled makes the review quite hard (I went thought the changes, but I’m not sure if something wasn’t missed since the commit is not small enough to be obviously correct).

lib/module-resolvers/file-resolver.js
mstemm: importer_path can be non-absolute when the program has a generic name like 'main'. That in turn happens for programs in the cli repl or when provided via -e.

mstemm:
Ok I've split the work into more smaller commits and also changed the resolver interface to return the full pathname in the name: property of the returned object. The filename: property is gone. I think name: is more appropriate than filename: as filename isn't a good fit for the url resolver.

I've also addressed all the smaller changes suggested.

Please take another look?

docs/concepts/programming_constructs.md
```diff
@@ -159,10 +159,26 @@ location, then in another program you can import it as a module, using the
     import "<module_path>" as local_name;
 ```
 
-How the `<module_path>` is resolved depends on the specific capabilities of the
-application in which Juttle is running. For the Juttle CLI, it can be either a
-relative or absolute filesystem path or a URL to a remote Juttle file containing
-exported code.
+`<module_path>` can either refer to a *User Import* or a *System
+Import*.
```
dmajda: Do we really want these to be Capitalized Terms? Makes ours docs look a bit like legalese…

docs/concepts/programming_constructs.md
```diff
@@ -159,10 +159,26 @@ location, then in another program you can import it as a module, using the
...
+
+User Imports are references to individual modules. User Imports can
+either be a url, a full pathname starting with `/`, or a relative
+pathname starting with `./` or `../`. All other paths are considered
+System Imports.
+
+When resolving User Imports with relative pathnames, the pathname is
+always applied relative to the file doing the import. See the below
+examples for details.
```
dmajda: “See the examples below for details.”?

lib/module-resolvers/file-resolver.js
```diff
@@ -39,22 +39,58 @@ class FileResolver {
...
-        logger.debug(`Resolving module_path=${module_path} module_name=${module_name} importer_path=${importer_path}`);
+        logger.debug(`searching local module_path=${module_path} module_name=${module_name} importer_path=${importer_path}`);
+
+        // Resolve module_path relative to importer_path or the
+        // current working directory.
+        let dir = process.cwd();
+        if (importer_path !== undefined &&
+            path.isAbsolute(importer_path)) {
+            dir = path.dirname(importer_path);
+        }
```
dmajda: Would be good to add a comment on when the importer path is non-absolute (basically your [answer here](https://github.com/juttle/juttle/pull/591#discussion_r56397960)).

lib/module-resolvers/file-resolver.js
```diff
@@ -69,7 +105,14 @@ class FileResolver {
     _resolve(module_path, module_name, importer_path) {
         var self = this;
 
-        var found = self._search(module_path, module_name, importer_path);
+        var found = false;
+        if (module_path.startsWith('./') ||
+            module_path.startsWith('../') ||
+            path.isAbsolute(module_path)) {
```
dmajda: Maybe put the condition into a helper method like `_is_local_import` for more clarity?

dmajda:
On list item is wrongly indented in the commit message of https://github.com/juttle/juttle/pull/591/commits/bd0d57fd2235261401cefe90925f63709dd77963.

dmajda:
For the record, I don’t like the AST modification done in https://github.com/juttle/juttle/pull/591/commits/bd0d57fd2235261401cefe90925f63709dd77963 much, but I’d say it’s good for the time being. (The whole parser-level module resolution thing needs rethinking anyway.)

dmajda:
+1, thanks!

> One open question--do we want imports of standard library modules to have a 'stdlib' prefix?

After thinking it through, I don’t think so. All library code will be in stdlib and the prefix would just be unnecessary bloat.

rlgomes:
@dmajda @mstemm @demmer  we need to pull this module resolution into `juttle-spec.js` so that the juttle spec files can import those stdlib modules without having to do "predict.juttle" as it works fine from the CLI like so: 

```rlgomes@x230> ./bin/juttle -e "import 'predict' as predict ; emit -limit 10 -from :0: | put value = count() | predict.predict"
┌────────────────────────────────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────┐
│ time                               │ value    │ E        │ P        │ S        │ T        │ Y        │ Z        │
├────────────────────────────────────┼──────────┼──────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│ 1970-01-01T00:00:00.000Z           │ 5        │ 0        │ 5        │ null     │ 5        │ 0        │ null     │
└────────────────────────────────────┴──────────┴──────────┴──────────┴──────────┴──────────┴──────────┴──────────┘
``` 

but if you modify one of the imports in https://github.com/juttle/juttle/tree/master/test/runtime/specs/juttle-spec/stdlib the test utilites can't resolve the standard library that way. I tried figuring out if there was a way to fix the test utilities but at the same time shouldn't this just work ? otherwise we have to also fix the module resolution within juttle-engine ?

mstemm:
juttle-test-utils needs to change from using its own module resolver to using the standard file resolver. Not sure why it has a custom module resolver, do you know?

rlgomes:
@mstemm lets wait for @dmajda he probably knows that code better than I do.

dmajda:
The custom resolver in `juttle-test-utils` is needed because we want the ability to specify modules inline in JuttleSpec tests (grep `### Module` to see examples) and also some JavaScript tests.

It looks like the best course of action would be to chain the inline module resolver with the file resolver and possibly the URL resolver using `resolver_utils.multiple`.

PR #593, opened 2016-03-14
csv errors with location information

### LINKED ISSUE #592: csv parser should indicate line and column when hitting errors
originally mentioned in https://github.com/juttle/juttle/issues/562, we should either find a better parser or fix the existing one.

rlgomes:
fixes #592

moving to use the `fast-csv` parser gives us the exact line that failed
to parse and we're just appending which line it was at


rlgomes:
@dmehra @davidvgalbraith @dmajda so this would make the errors reported by the csv parser a bit easier to read since it includes the exact line that failed (with a suffixed `at line X`). The other nice thing is this parser would allow us to handle blank lines, incomplete rows and lines with comments: https://www.npmjs.com/package/fast-csv#parsing which we could expose just like the `-separator` option.  Let me know what you think and I'll the minimally required tests here with a separate commit to expose the `-comments '#'`, `-ignoreEmpty` and `-allowIncompleteRows` options which would make the parser super flexible in the real world where the csv "standard" has been highly abused.

dmajda:
Add a test for the line reporting?

> Let me know what you think and I'll the minimally required tests here with a separate commit to expose the `-comments '#'`, `-ignoreEmpty` and `-allowIncompleteRows` options which would make the parser super flexible in the real world where the csv "standard" has been highly abused.

Makes total sense to me — be strict by default but offer flexibility when needed.

Please implement these options in separate PR(s) in order to keep this one focused to line number reporting.

test/adapters/http/read.spec.js
```diff
@@ -470,7 +470,7 @@ describe('HTTP read tests', function() {
                 return check_juttle({
                     program: 'read http -headers { Accept: "' + contentType + '"} ' +
                              '          -url "' + server.url + '/objects?foo=' +
-                             Array(2048).join('X') + '&count=100" | head 1'
+                             Array(8*1024).join('X') + '&count=100" | head 1'
```
davidvgalbraith: What's this?

test/adapters/http_server.spec.js
```diff
@@ -276,7 +276,7 @@ describe('read http_server', function() {
             return finish;
         })
         .then(function(results) {
-            expect(results.errors).to.deep.equal(['Invalid CSV data: Error: Row length does not match headers']);
+            expect(results.errors).to.deep.equal(['Invalid CSV data: Red Dragon, 24.04 at line 2']);
```
davidvgalbraith: It seems like we lost something useful in the `Row length does not match headers`, can we have both that and the new message?

test/adapters/http/read.spec.js
davidvgalbraith: Maybe if I read the commit message

rlgomes: I'll do my best to explain and then put a comment here for future readers of the code. Some of the CSV parsers read a ahead a small amount 1-2KB and then parse that and fire events, then read again ahead. So when you tell the parser to `end,close,stop` it will in fact spew out a few more entries (as many as can fit in its internal buffer). When I switched the csv parser here the new one must have a larger buffer and therefore we were getting more entries. I made the test lines much longer resulting in filling the internal buffer quicker and therefore not exceeding the expected 50 entries of reading ahead.

lib/adapters/parsers/csv.js
```diff
@@ -33,7 +33,7 @@ class CSVParser extends base {
                     self.doneParsing = true;
                     emit(buffer);
                     buffer = [];
-                    csvStream.end();
+                    csvStream.emit('end');
```
davidvgalbraith: Does `end()` not emit `'end'`? Do you not have to call `end()` at some point to clean the stream up properly?

rlgomes: I think this was from fiddling to understand why that other tests was failing due to the "internal buffer size changes"

rlgomes: Yup this was completely accidental change and great catch, now removed from this PR.

davidvgalbraith:
+1

test/adapters/http_server.spec.js
rlgomes: These darn csv parser all have their own magical error messages. The new one gives me the full line on an error while the old one would indicate the message you liked `"Row length does not match headers"`. Then the messages diverge even more once certain flags are turned on and off: https://github.com/C2FO/fast-csv/blob/master/lib/parser/parser_stream.js#L127

Not sure how to easily reconcile this... I'm thinking I could make the message instead look like so:

```
Invalid CSV data: "Red Dragon, 24.04" at line 2 does not match header line "product,amount,saleTime"
```

@davidvgalbraith what do you think ? its a bit more verbose but allows the user to compare the problematic line with the header line ?

davidvgalbraith: That sounds good.

rlgomes: Not as easy to get the headers as I had thought so if the error happens on the second line I don't have the headers (since no valid point was processed till that point) and would simply write out 

```
Invalid CSV data: "Red Dragon, 24.04" at line 2 does not match header line
```

So I'm thinking I'll just print that instead (without the actual header line)

lib/adapters/parsers/csv.js
rlgomes: @davidvgalbraith yes this is necessary now since `fast-csv` doesn't really care if you call the `end()` method and only responds to the `end` event. Figured this out through some code inspection and trial and error.

rlgomes:
I think it won't be controversial to merge this at this point and I can make any follow up PR's quickly. Just want to get this in so I can work on quickly adding the options for #582

PR #595, opened 2016-03-15
Add google analytics tracking

mnibecker:
From: http://www.mkdocs.org/user-guide/configuration/#google_analytics

@demmer @dmehra


dmehra:
+1, what will this let us track - popularity of individual docs articles?

demmer:
Yep. That it should. 

Let's do the same for the juttle-viz docs. 

-m

PR #599, opened 2016-03-15
Add comments & simplify conditions join

bkutil:
While reading through `join` code, I annotated the code with a few extra comments and did minor simplifications around conditions.

Putting them up to see if they might be useful to future readers.


bkutil:
cc: @dmajda

dmajda:
+1

bkutil:
Yeah, the redundancy stems from the logic being complicated in certain places. A refactoring would certainly help :)

PR #601, opened 2016-03-15
Fix processing of argument of the “*” operator

### LINKED ISSUE #598: Member expressions inside the “*” operator in lvalues are not processed correctly
Example:

```
$ bin/juttle -e 'function f() { }; emit -limit 1 | put *f["x"] = 5'
Error: f_3 is not defined [ReferenceError: f_3 is not defined]
└───────────┴───────────┘
```

The expected output is:

```
$ bin/juttle -e 'function f() { }; emit -limit 1 | put *f["x"] = 5'
<input>:1:40: 
   1:function f() { }; emit -limit 1 | put *f["x"] = 5
                                            ^
Error: Cannot use a function as a variable (CANNOT-USE-AS-VARIABLE)
```

dmajda:
Consider an expression like this:

```juttle
*field = "foo"
```

The `field` variable is evaluated and its value is used to select a point field. It is not itself assigned to. In semantic pass, this corresponds to processing with `this.lhs == false`.

Before this commit, code in `sa_UnaryExpression` wrongly processed `*` arguments with the same `this.lhs` value as the operator itself. This PR fixes it to always process them with `this.lhs == false`.

Fixes #598.

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -1270,39 +1270,44 @@ class SemanticPass {
         var op = node.operator;
         var name;
 
-        if (op === '++' || op === '--') {
-            this.with_lhs(true, () => {
+        switch (op) {
+            case '*':
                 this.with_expr_mode('result', () => {
                     this.sa_expr(node.argument);
                 });
-            });
-        } else {
-            this.with_expr_mode('result', () => {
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
+1

Issue #430, opened 2016-02-18
Juttlespec should verify the number of warnings/errors raised


bkutil:
Currently, the generated code verifies that the warnings/errors are present (and in a given order): 

`expect(res.warnings[0]).to.include(...)`.

However, if less than expected warnings are emitted, the tests pass. Ideally, juttlespec should also do something like

`expect(warnings.length).to.equal(emitted_warnings.length)`.

so that the tests fail if some warnings/errors are missing.

dmajda:
> However, if less than expected warnings are emitted, the tests pass.

JuttleSpec tests indeed fail when some expected warnings/errors are missing, which can be demonstrated with the following example (produces 1 warning, expects 2):

```markdown
# Test spec

## Emits warnings

### Juttle

    emit -from :0: -limit 1 | put time = time + null

### Warnings

  * Invalid operand types for "+": date (1970-01-01T00:00:00.000Z) and null.
  * Invalid operand types for "+": date (1970-01-01T00:00:01.000Z) and null.
```

On the other hand, JuttleSpec tests pass when there are more than expected warnings/errors, which can be demonstrated with the following example (produces 2 warnings, expects 1):

```markdown
# Test spec

## Emits warnings

### Juttle

    emit -from :0: -limit 2 | put time = time + null

### Warnings

  * Invalid operand types for "+": date (1970-01-01T00:00:00.000Z) and null.
```

@bkutil Was the latter case what you really meant?

bkutil:
@dmajda - yes, the latter case is correct - I mistakenly wrote 'less' instead of 'more' in the original description.

dmajda:
OK. Marking as a bug as the intention was always to match the errors/warnings exactly.

bkutil:
Fixed in #602

PR #602, opened 2016-03-15
Stronger juttle spec assertions

bkutil:
Require tests to emit # of warnings / errors equal to the number of errors / warnings expected.

This also changes the semantics of juttle-spec's assertions. A missing errors/warnings section no longer means 'any number of errors', but 'no errors'.

Fix tests that were broken by this change.

### DIFF
--- a/test/spec/juttle-spec.js
+++ b/test/spec/juttle-spec.js
@@ -152,10 +152,12 @@ class SpecRenderer {
             parts.push('            expect(res.sinks.result).to.deep.equal(' + util.inspect(test.output, { depth: null }) + ');');
         }
 
+        parts.push('            expect(res.errors.length, "Expected test to emit ' + test.errors.length + ' errors: " + JSON.stringify(res.errors)).to.equal(' + test.errors.length + ');');
         for (var erri=0 ; erri < test.errors.length ; erri++) {
             parts.push('            expect(res.errors[' + erri + ']).to.include(' + JSON.stringify(test.errors[erri]) + ');');
         }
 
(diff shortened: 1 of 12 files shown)


bkutil:
refs: #430
cc: @dmajda

test/spec/juttle-spec.js
```diff
@@ -155,10 +155,12 @@ class SpecRenderer {
         for (var erri=0 ; erri < test.errors.length ; erri++) {
             parts.push('            expect(res.errors[' + erri + ']).to.include(' + JSON.stringify(test.errors[erri]) + ');');
         }
+        parts.push('            expect(res.errors.length, "Expected test to emit ' + test.errors.length + ' errors: " + JSON.stringify(res.errors)).to.equal(' + test.errors.length + ');');
```
dmajda: I’d move this test above the tests validating individual errors. This will prevent non-sensical message comparisons in cases where the code produces less errors than expected.

test/spec/juttle-spec.js
```diff
@@ -155,10 +155,12 @@ class SpecRenderer {
...
 
         for (var warni=0 ; warni < test.warnings.length ; warni++) {
             parts.push('            expect(res.warnings[' + warni + ']).to.include(' + JSON.stringify(test.warnings[warni]) + ');');
         }
+        parts.push('            expect(res.warnings.length, "Expected test to emit ' + test.warnings.length + ' warnings: " + JSON.stringify(res.warnings)).to.equal(' + test.warnings.length + ');');
```
dmajda: Same here.

dmajda:
+1

bkutil:
Moved the # assertions before the loops. Thanks! Will merge on green.

PR #604, opened 2016-03-15
Inputs register

### LINKED ISSUE #148: Registering juttle inputs with the compiler
In Jut 1.0, the compiler API that the web client used to get input definitions out of a juttle program accepted the default values for each input type. More specifically, the client passed in a map that mapped input type to a function that returned the default value for that type. The compiler used these functions when an input didn't already have a value and no `-default` option was specified. This all happened on the browser-side.

It is important that the right default value gets applied because otherwise we may run into operations on incompatible types. Also, since an input's value can be used as an option into a later input, its important that the compiler computes the right default value.

In the new outrigger world, we send a bundle (program and modules) and current input values to the server-side `/prepare` API and get back the input definitions (and values).

There is currently no way to tell the compiler what the default values are for each input type...this is what this ticket is about.

@demmer and I think a good way to achieve this would be to make inputs (i.e. their default values, valid option lists, etc.) something you can register with the compiler (similar to how you can register adapters). Giving the compiler this input-specific knowledge is definitely a departure from Jut 1.0, but we believe that `input`s can/should have an API that is closer to adapters than views (which the compiler knows nothing about).

Other options that we considered but decided against:
- Require users to specify `-default` for all inputs. This is too much of a burden to put on users.
- Send the per-input-type defaults with each `/prepare` call. This makes the API awkward and doesn't allow you to specify default value functions which is necessary for the `select` input which should default to the first item in the list.

@dmajda, we wanted to get your feedback on this direction.

@demmer, please add if I missed anything.

### LINKED ISSUE #558: Error on invalid input values (with revert option)
resulting from issue with changing inputs in juttle/juttle-viewer#34
Expands upon #148

Currently juttle is indifferent to the value of inputs in regards to their type. For instance if I have this program:
```
input my_text: text;
```

and I provide the compiler with the following input value `{ my_text: ['a', 'b', 'c'] }`, the compiler will not complain and simply return the obviously invalid value as the inputs value and a runtime error will get thrown if you try to do any stringy things with the input value (correct me if I'm wrong).

Following what was defined in #148 it would be nicer if juttle were more aware  of what value type each input_type expected and either threw an error or revert to input default for invalid values.

As currently implemented we push this responsibility to any code that wants to implement juttle-inputs, which increases the complexity to implementing any client that wants to use juttle.

By default the compiler should throw an error if an invalid input_value is provided for an input. There should be a compiler option that would discard any invalid options and revert to the input default. Perhaps an option called `input_mismatch_revert` (defaults to false), that when set to true will not throw and error and instead revert invalid values to the input default.

mattnibs:
Add input value and param validation and normalization in the compiler. 

Essentially I've created a registry of input types that the compiler supports. The input classes will validate and morph params and values passed into it and throw errors accordingly. 

If an input-type is not supported an error is thrown.

Resolves #148 
Resolves #558 

Will eventually help juttle/juttle-viewer#34

TODO:
- [ ] Add support for multi-select input
- [ ] Add support for date input
- [ ] Add support for duration input

### DIFF
--- a/lib/compiler/graph-builder.js
+++ b/lib/compiler/graph-builder.js
@@ -3,6 +3,7 @@
 var _ = require('underscore');
 var values = require('../runtime/values');
 var procs = require('../runtime/procs');
+var inputs = require('../inputs');
 var errors = require('../errors');
 var SemanticPass = require('./semantic');
 
@@ -173,10 +174,23 @@ class GraphBuilder {
         }
         return key + '[' + this.input_seqnos[key]  + ']';
     }
-    define_input(bname, input_name, options, isStatic) {
-        var input = {id: bname, name: input_name, options: options, static: isStatic};
-        input.value = this._compute_input_val(input);
-        this.inputs.push(input);
+    define_input(bname, input_type, params, isStatic) {
+        let InputConstructor = inputs[input_type];
+
+        if (!InputConstructor) {
+            throw errors.compileError('UNKNOWN-INPUT', {
+                input_type: input_type
+            });
+        }
+
+        let input = new InputConstructor(bname, params);
+
+        if (_.has(this.input_values, input.id)) {
+            input.setValue(this.input_values[bname]);
+        }
+
+        this.inputs.push(input.toObj());
+
         return input.value;
     }
     add_function(ast) {
--- a/lib/inputs/base-input.js
+++ b/lib/inputs/base-input.js
@@ -1,20 +1,20 @@
 'use strict';
 
 class BaseInput {
-    constructor(type, id, options, inputDefaultValue) {
+    constructor(type, id, params, inputDefaultValue) {
         this.type = type;
         this.id = id;
-        this.options = this.normalizeOptions(options);
+        this.params = this.normalizeParams(params || {});
 
-        this.value = this.options.default ? this.normalizeValue(this.options.default) : this.getInputDefaultValue();
+        this.value = this.params.default || this.getInputDefaultValue();
     }
 
     getInputDefaultValue() {
         throw new Error('getInputDefaultValue must be implemented');
     }
 
-    normalizeOptions(options) {
-        return options || {};
+    normalizeParams(params) {
+        return params || {};
     }
 
     normalizeValue() {
@@ -30,7 +30,7 @@ class BaseInput {
             id: this.id,
             type: this.type,
             value: this.value,
-            options: this.options
+            params: this.params
         };
     }
 }
--- a/lib/inputs/base-input.js
+++ b/lib/inputs/base-input.js
@@ -1,0 +1,38 @@
+'use strict';
+
+class BaseInput {
+    constructor(type, id, options, inputDefaultValue) {
+        this.type = type;
+        this.id = id;
+        this.options = this.normalizeOptions(options);
+
+        this.value = this.options.default ? this.normalizeValue(this.options.default) : this.getInputDefaultValue();
+    }
+
+    getInputDefaultValue() {
+        throw new Error('getInputDefaultValue must be implemented');
+    }
+
+    normalizeOptions(options) {
+        return options || {};
+    }
+
+    normalizeValue() {
+        throw new Error('normalizeValue must be implemented');
+    }
+
+    setValue(value) {
+        this.value = this.normalizeValue(value);
+    }
+
+    toObj() {
+        return {
+            id: this.id,
+            type: this.type,
+            value: this.value,
+            options: this.options
+        };
+    }
+}
+
+module.exports = BaseInput;
--- a/lib/inputs/number-input.js
+++ b/lib/inputs/number-input.js
@@ -2,16 +2,38 @@
 
 var _ = require('underscore');
 var BaseInput = require('./base-input');
+var errors = require('../errors');
 
 const DEFAULT_VALUE = 0;
 
 class NumberInput extends BaseInput {
-    constructor(id, options) {
-        super('number', id, options);
+    constructor(id, params) {
+        super('number', id, params);
     }
 
     getInputDefaultValue() {
         return DEFAULT_VALUE;
+    }
+
+    normalizeParams(params) {
+        if (params.default) {
+            try {
+                params = Object.assign({}, params, {
+                    default: this.normalizeValue(params.default)
+                });
+            } catch (err) {
+                if (!(err instanceof errors.CompileError)) throw err;
+
+                throw errors.compileError('INPUT-INVALID-PARAM', {
+                    input_id: this.id,
+                    param: 'default',
+                    param_value: params.default,
+                    message: err.message
+                });
+            }
+        }
+
+        return params;
     }
 
     normalizeValue(value) {
@@ -19,7 +41,11 @@ class NumberInput extends BaseInput {
             return value;
         }
 
-        throw new Error('invalid input value');
+        throw errors.compileError('INPUT-INVALID-VALUE', {
+            input_id: this.id,
+            value: value,
+            message: 'Value must be a number'
+        });
     }
 }
 
--- a/lib/inputs/number-input.js
+++ b/lib/inputs/number-input.js
@@ -1,0 +1,26 @@
+'use strict';
+
+var _ = require('underscore');
+var BaseInput = require('./base-input');
+
+const DEFAULT_VALUE = 0;
+
+class NumberInput extends BaseInput {
+    constructor(id, options) {
+        super('number', id, options);
+    }
+
+    getInputDefaultValue() {
+        return DEFAULT_VALUE;
+    }
+
+    normalizeValue(value) {
+        if (_.isNumber(value)) {
+            return value;
+        }
+
+        throw new Error('invalid input value');
+    }
+}
+
+module.exports = NumberInput;
--- a/lib/inputs/select-input.js
+++ b/lib/inputs/select-input.js
@@ -1,0 +1,40 @@
+'use strict';
+
+var _ = require('underscore');
+var BaseInput = require('./base-input');
+
+class SelectInput extends BaseInput {
+    constructor(id, options) {
+        super('select', id, options);
+    }
+
+    getInputDefaultValue() {
+        return this.options.items[0].value;
+    }
+
+    normalizeOptions(options) {
+        options = Object.assign({}, options) || {};
+
+        if (!options.items || !_.isArray(options.items)) {
+            throw new Error('invalid options');
+        }
+
+        options.items = options.items.map((item) => {
+            return _.isString(item) ? { value: item, label: item } : item;
+        });
+
+        return options;
+    }
+
+    normalizeValue(value) {
+        let retrievedValue = _.findWhere(this.options.items, { value });
+
+        if (!retrievedValue) {
+            throw new Error('invalid value');
+        }
+
+        return retrievedValue.value;
+    }
+}
+
+module.exports = SelectInput;
--- a/lib/inputs/select-input.js
+++ b/lib/inputs/select-input.js
@@ -2,35 +2,87 @@
 
 var _ = require('underscore');
 var BaseInput = require('./base-input');
+var errors = require('../errors');
 
 class SelectInput extends BaseInput {
-    constructor(id, options) {
-        super('select', id, options);
+    constructor(id, params) {
+        super('select', id, params);
     }
 
     getInputDefaultValue() {
-        return this.options.items[0].value;
+        return this.params.items[0].value;
     }
 
-    normalizeOptions(options) {
-        options = Object.assign({}, options) || {};
+    normalizeParams(params) {
+        params = Object.assign({}, params);
 
-        if (!options.items || !_.isArray(options.items)) {
-            throw new Error('invalid options');
+        if (!params.items) {
+            throw errors.compileError('INPUT-REQ-PARAMS-MISSING', {
+                input_id: this.id,
+                missing_params: 'items'
+            });
         }
 
-        options.items = options.items.map((item) => {
-            return _.isString(item) ? { value: item, label: item } : item;
+        if (!_.isArray(params.items)) {
+            throw errors.compileError('INPUT-INVALID-PARAM', {
+                input_id: this.id,
+                param: 'items',
+                param_value: params.items,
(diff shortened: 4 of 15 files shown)


mnibecker:
@dmajda @go-oleg - Because I'm a noob to this side of things a review would be appreciated before I wrap up the rest of the work for this pr.

lib/inputs/number-input.js
```diff
@@ -0,0 +1,52 @@
...
+
+var _ = require('underscore');
+var BaseInput = require('./base-input');
+var errors = require('../errors');
+
+const DEFAULT_VALUE = 0;
+
+class NumberInput extends BaseInput {
+    constructor(id, params) {
+        super('number', id, params);
```
go-oleg: Instead of having a different constructor signature between the `BaseInput` and subclasses, I'd have the subclasses implement a

```
getType() {
  return 'number';
}
```

or explicitly set `this.type` in the subclass constructor.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+
+    getInputDefaultValue() {
+        return this.params.items[0].value;
+    }
+
+    normalizeParams(params) {
+        params = Object.assign({}, params);
+
+        if (!params.items) {
+            throw errors.compileError('INPUT-REQ-PARAMS-MISSING', {
```
go-oleg: I'd spell out `REQUIRED` here to make it clear what it stands for.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+                missing_params: 'items'
+            });
+        }
+
+        if (!_.isArray(params.items)) {
+            throw errors.compileError('INPUT-INVALID-PARAM', {
+                input_id: this.id,
+                param: 'items',
+                param_value: params.items,
+                message: 'Item param must be an array'
```
go-oleg: "Item" doesn't need to be in `message` because its already in `param` above, right?

If it is there, it should be lowercased and quoted.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+            });
+        }
+
+        params.items = params.items.map((item) => {
+            // if value is string, number or array expand to obj with value and label
+            if (_.isString(item) || _.isNumber(item) || _.isArray(item)) {
```
go-oleg: How about changing this to:

If its not an object that contains a `value` and `label`, expand it like you below.

With that, I think the following `!item.value || !item.label` should become unnecessary.

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
...
+
+    normalizeValue() {
+        throw new Error('normalizeValue must be implemented');
+    }
+
+    setValue(value) {
+        this.value = this.normalizeValue(value);
+    }
+
+    toObj() {
```
go-oleg: Suggest renaming to `toJSON` (which is expected to return the value that should be serialized), or something like `valueOf`. I don't particularly like either of these though, @dmajda would prolly have a better suggestion.

test/inputs/text-input.spec.js
```diff
@@ -0,0 +1,84 @@
...
+            let input = new TextInput('id', testObj.params);
+            expect(input.toObj()).to.deep.equal(testObj);
+
+            input.setValue('other value');
+            expect(input.toObj()).to.deep.equal(updateTestObj({
+                value: 'other value'
+            }));
+        });
+
+        it('not setting value or default has value of emptry string', () => {
```
go-oleg: `s/emptry/empty/`

go-oleg: I'd also rephrase the description to be: "builtin default value is an empty string".

test/inputs/text-input.spec.js
```diff
@@ -0,0 +1,84 @@
...
+var expect = require('chai').expect;
+var TextInput = require('../../lib/inputs/text-input');
+var expectError = require('./input-test-utils').expectError;
+
+describe('text input', () => {
+    describe('valid cases', () => {
+        const testObj = {
+            id: 'id',
+            type: 'text',
+            value: 'default value',
```
go-oleg: Change `value` to be different from `params.default` to make sure `params.default` doesn't override `value`.

test/inputs/select-input.spec.js
```diff
@@ -0,0 +1,130 @@
...
+
+        it('properly transforms items with shortened syntax', () => {
+            let select = new SelectInput('id', {
+                default: '1',
+                items: [ '1', '2', '3']
+            });
+            expect(select.toObj()).to.deep.equal(testObj);
+        });
+
+        it('propery transforms array items with shortened syntax', () => {
```
go-oleg: `s/propery/properly/`

test/inputs/select-input.spec.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+};
+
+const updateTestObj = (updateObj) => {
+    return Object.assign({}, testObj, updateObj);
+};
+
+describe('select-input', () => {
+    describe('valid cases', () => {
+        it('handles normal text input with default and value', () => {
```
go-oleg: `s/text/select/`

test/inputs/select-input.spec.js
```diff
@@ -0,0 +1,130 @@
...
+            {value: '3', label: '3'}
+        ]
+    }
+};
+
+const updateTestObj = (updateObj) => {
+    return Object.assign({}, testObj, updateObj);
+};
+
+describe('select-input', () => {
```
go-oleg: consistency nit: the other spec files don't have a dash here.

test/compiler/inputs.spec.js
```diff
@@ -79,14 +83,26 @@ describe('Juttle inputs', function() {
...
 
-    describe('gadgets producing a filter', function() {
+    describe('inputs depending on inputs', function() {
+        it('label gets populated with view of input', () => {
+            return check_juttle({
+                program: 'input a: text; input b: text -label a;' +
+                         'emit | view sink'
+            })
+            .then((res) => {
+                // console.log(res);
```
go-oleg: add verification.

lib/inputs/base-input.js
dmajda: Actually, `toJSON` sounds fine to me. Or maybe `serialize`.

lib/compiler/graph-builder.js
```diff
@@ -173,10 +174,23 @@ class GraphBuilder {
         }
         return key + '[' + this.input_seqnos[key]  + ']';
     }
-    define_input(bname, input_name, options, isStatic) {
-        var input = {id: bname, name: input_name, options: options, static: isStatic};
-        input.value = this._compute_input_val(input);
```
dmajda: The `this._compute_input_val` method is no longer used, it should be deleted together with the “input value” section of `test/compiler/inputs.spec.js`.

lib/inputs/text-input.js
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        return params;
+    }
+
+    normalizeValue(value) {
+        if (_.isString(value)) {
+            return value;
+        } else if (_.isNumber(value)) {
+            return value.toString();
```
dmajda: I’d slightly prefer using `values.toString(value)` (a helper form `values.js`) because it doesn’t require creating an object wrapper and it is how we do these conversions elsewhere in the compiler.

A bigger point: Why do we need this conversion? And why just numbers, not other Juttle types?

lib/inputs/text-input.js
```diff
@@ -0,0 +1,56 @@
...
+    getInputDefaultValue() {
+        return DEFAULT_VALUE;
+    }
+
+    normalizeParams(params) {
+        if (params.default) {
+            try {
+                params = Object.assign({}, params, {
+                    default: this.normalizeValue(params.default)
+                });
```
dmajda: I think `_.extend(_.clone(params), { default: ... })` would better capture what you are doing here.

lib/inputs/text-input.js
```diff
@@ -0,0 +1,56 @@
...
+            } catch (err) {
```
dmajda: Putting the `try`/`catch` around the whole `Object.assign` thing makes if non-obvious from where you expect the error to come from. I’d just do `let default = this.normalizeValue(params.default)`, wrap it into `try`/`catch`, and then do `Object.assign`.

lib/inputs/number-input.js
```diff
@@ -0,0 +1,52 @@
...
+                    input_id: this.id,
+                    param: 'default',
+                    param_value: params.default,
+                    message: err.message
+                });
+            }
+        }
+
+        return params;
+    }
```
dmajda: Same notes apply here as for `TextInput#normalizeParams`.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+                    message: err.message
+                });
+            }
+        }
+
+        return params;
+    }
+
+    normalizeValue(value, items) {
+        items = items || this.params.items;
```
dmajda: Rather than overload the function to do two slightly different things I’d create an internal helper and make `normalizeValue` call it.

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+class BaseInput {
+    constructor(type, id, params, inputDefaultValue) {
```
dmajda: Just `default` or `defaultValue`? The `input` part is redundant.

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+class BaseInput {
```
dmajda: Just `Input`?

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
...
+    constructor(type, id, params, inputDefaultValue) {
+        this.type = type;
+        this.id = id;
+        this.params = this.normalizeParams(params || {});
```
dmajda: Is there any case when `params` wouldn’t be set? If not, I’d just make it a required parameter.

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
...
+
+        this.value = this.params.default || this.getInputDefaultValue();
+    }
+
+    getInputDefaultValue() {
```
dmajda: Same as with the constructor parameter: `getDefault` or `getDefaultValue`.

lib/inputs/base-input.js
```diff
@@ -0,0 +1,38 @@
...
+        throw new Error('getInputDefaultValue must be implemented');
+    }
+
+    normalizeParams(params) {
+        return params || {};
```
dmajda: Same thing as in the constructor — is there any case when `params` wouldn’t be set? If not, I’d just make it a required parameter.

Here, the issue is even more pronounced as a method called `normalizeParams` really shouldn’t invent its own params when not supplied any.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+
+        let retrievedValue = _.find(items, (item) => {
+            return _.isEqual(value, item.value);
```
dmajda: Use `values.equal`— a generic Juttle value equality function which will work for all Juttle values.

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+
+        if (!retrievedValue) {
+            throw errors.compileError('INPUT-INVALID-VALUE', {
+                input_id: this.id,
+                value: value,
+                message: 'Value must be a member of -items'
+            });
+        }
+
+        return retrievedValue.value;
```
dmajda: Why not return just `value`?

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+var BaseInput = require('./base-input');
+var errors = require('../errors');
+
+class SelectInput extends BaseInput {
+    constructor(id, params) {
+        super('select', id, params);
+    }
+
+    getInputDefaultValue() {
+        return this.params.items[0].value;
```
dmajda: What happens if the items are empty?

lib/inputs/select-input.js
```diff
@@ -0,0 +1,92 @@
...
+                    input_id: this.id,
+                    param: 'default',
+                    param_value: params.default,
+                    message: err.message
+                });
+            }
+        }
+
+        return params;
+    }
```
dmajda: Same notes apply here as for `TextInput#normalizeParams`.

lib/inputs/select-input.js
dmajda: I think I’d actually prefer to expanding objects which lack `value` or `label`. It’s more likely the user makes a mistake of forgetting these properties than that he/she wants to actually expand an object.

dmajda: Type checks for all Juttle values should use functions from `values.js` (`values.isString`, `values.isArray`, etc.). This is true here but in other places in the code too. There is no actual difference in behavior in the simple cases like numbers but it would still be good to do this as a matter of principle and to increase robustness should we later deal with Juttle-specific types such as moments.

dmajda:
Looks good overall. A couple of high-level notes:

  * I’d probably join all the commits except the last one into one (they are closely tied together).

    (I think this is the first time ever I’m suggesting joining commits instead of splitting them :smile:.)

  * Names of `normalizeValue` and `normalizeParams` are misleading in that these methods actually also validate. Personally, I’d split them into `validateValue`/`validateParams` (or `checkValue`/`checkParams`) which would throw errors and `normalizeValue`/`normalizeParams` which would do any changes. But I understand this could lead to some code duplication so maybe it’s not that great idea. Up to you.

    If you decide to keep validation and normalization together, then improving method names could help a little. But I don’t have any suggestions beside `processValue`/`processParams` which are fairly generic (but I think I’d still take generic over misleading).

  * What is the purpose of making inputs classes versus simple functions? The only thing they really do is that they check & normalize values/params and serialize the input back. There is no state involved which would make objects more natural solution.

  * Terminology: I think we’re using the term “options” for “things with dashes” both internally (in the code) and externally (e.g. in the docs). I think I’d prefer that over “params” for inputs too, unless you have some strong reason to deviate.

Issue #557, opened 2016-03-08
Be lenient with empty line in csv files


dmehra:
Our csv file reader doesn't like files with empty lines (anywhere: top, middle, last line). Other csv readers out there accept such files just fine. We should be lenient about that too, as the problem isn't easy to identify for the user, given a large input file.

Sample file `/tmp/csv.csv`:
```

a,b,c,d

A,B,C,D

A,B,C,D

```

Juttle says (quite verbosely):
```
juttle> read file -file '/tmp/csv.csv' -format 'csv'
Error: internal error RuntimeError: Invalid CSV data: Error: Row length does not match headers { [RuntimeError: internal error RuntimeError: Invalid CSV data: Error: Row length does not match headers]
  message: 'internal error RuntimeError: Invalid CSV data: Error: Row length does not match headers',
  code: 'RT-INTERNAL-ERROR',
  info: 
   { error: 'RuntimeError: Invalid CSV data: Error: Row length does not match headers',
     procName: 'read',
     location: { filename: 'main', start: [Object], end: [Object] } } }
└───────────┴───────────┘
juttle> 
```

But for example [csvlint](https://github.com/Clever/csvlint) says
```
./csvlint /tmp/csv.csv 
file is valid
```

As far as the [RFC](https://tools.ietf.org/html/rfc4180) goes, it doesn't say anything one way or another about empty lines, only that the "last record in the file may or may not have an ending line break" (we handle that already). But it's a very bare RFC.

dmajda:
Is this motivated by an actual problem (software or humans producing CSV file(s) with empty lines)?

> Other csv readers out there accept such files just fine.

Examples?

> As far as the [RFC](https://tools.ietf.org/html/rfc4180) goes, it doesn't say anything one way or another about empty lines, only that the "last record in the file may or may not have an ending line break" (we handle that already).

That RFC contains a grammar which pretty unambiguously specifies that there are no blank lines allowed — individual records can be separated only by one CRLF sequence:

```
file = [header CRLF] record *(CRLF record) [CRLF]
```

But the RFC is only informational (not on the standards track), so no big deal ignoring it.

dmajda:
I filed #562 for the ugly error message.

dmehra:
> Is this motivated by an actual problem (software or humans producing CSV file(s) with empty lines)?

Yes, I ran into this when processing a real CSV file from our internal confluence. I don't know what software or human had generated it.

> Other csv readers out there accept such files just fine.
> Examples?

[...]

PR #605, opened 2016-03-16
expose more csv parser options

### LINKED ISSUE #603: expose other csv parser options in various builtin adapters
Since the `fast-csv` parser supports a few options we should expose them allowing the end user to deal with less than optimal CSV formats. These include `-comments '#', -ignoreEmpty true and -allowIncompleteRows true`

rlgomes:
fixes #603

exposes `comment`, `ignoreEmptyLines` and `allowIncmpleteLines` options for
csv parser along with necessary tests


rlgomes:
@dmajda @dmehra @davidvgalbraith more csv options! let me know what you think of the names and what else comes to mind but I think this also handles the #557 quite well. I'll update documentation once we agree on option naming.

rlgomes:
Other options exposed by `fast-csv` parser: https://www.npmjs.com/package/fast-csv#parsing some can be a bit "confusing" so I left them out but I'm open to suggestions.

dmehra:
+1 
I just discovered the need to skip top lines of a csv file because they are comments (double-quoted strings), I see people do this with Python Pandas using `skiprows=N` in read_csv; however the ability to skip comments would solve the same issue for me.

dmajda:
Typo in the commit message of https://github.com/juttle/juttle/pull/605/commits/8cc8a29d88178038b27669b0925f87aaadeb7213: “allowIncmpleteLines”.

lib/adapters/parsers/csv.js
```diff
@@ -9,6 +9,9 @@ class CSVParser extends base {
     constructor(options) {
         super(options);
         this.separator = options.separator ? options.separator : ',';
+        this.comment = options.comment ? options.comment : null;
```
dmajda: I wonder whether `commentChar` wouldn’t be a better name (it makes it more clear what the value is).

dmajda:
Re https://github.com/juttle/juttle/pull/605/commits/c7bfb482d2cb894e3ac1988c567a02f0f5e6c872 — why `file` and `stdio` but not `http` and `http_server`?

lib/adapters/parsers/csv.js
rlgomes: Then we'd have to rename `separator` to `separatorChar` ? I was keeping consistent with that specific option. @dmajda

rlgomes:
@dmajda and I'll add the options to http & http_server and we can decide on the name of the option once that is done.

rlgomes:
@dmajda did all the updates including exposing the options within the file, stdio, http and http_sever adapter and renamed `-separator` to `-separatorChar` and so we should advertise on gitter this backward comptability breakage at least to the user that requested the `separator` option.

davidvgalbraith:
+1

dmajda:
I’d actually keep `-separator` as it is even at the cost of slight inconsistency. Unlike `-commentChar` it will be used relatively often (so it should be short) and the term is well established.

Thinking about this, maybe rename the `-commentChar` option to `-commentStart`? That would move us from describing the type or allowed value to describing the meaning, making it more consistent with `-separator`. The cost is a slight loss of precision as people could think this could be e.g. a longer string or a regexp. Up to you.

dmehra:
-commentSymbol ?

rlgomes:
Alright updated with all remaining suggestions and the option is now `characterSymbol` with the `separator` option kept as is. Will merge on green.

PR #606, opened 2016-03-16
Add nested expression tests for “.” and “[]” operators

dmajda:
Just to make sure this situation is covered.

Part of work on #419.


dmajda:
@bkutil

bkutil:
Cool! What happens if one of the fields in the nested chain doesn't exist?

dmajda:
You’ll get `null` as a result and then a type error further in the chaing because `.` or `[]` on `null` is incorrect.

I added this case to the tests.

bkutil:
+1

PR #607, opened 2016-03-16
Generalize assignment LHS handling

dmajda:
Assignment LHS was handled specially both in the semantic pass and in later compiler stages. This was  bad design which was largely unnecessary with the current compiler structure. This PR removes all traces of this special handling and makes it so that assignments are handled mostly like all other expressions.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

Issue #608, opened 2016-03-16
pace behaves strangely with a historic read


bkutil:
This is a bug originally tracked as PROD-8679.

```
emit -from :-10m: -to :+10m:
| put hostname='foo'+Number.toString(count()%10)
| reduce -every :m: hourly=10*Math.floor(count()/10) by hostname
| reduce hourly=10*Math.floor(count()/10) by hostname
| reduce hosts=count(hostname) by hourly
| put read_time=Date.time()
| pace -every :s:
| put pace_time=Date.time()
| view text -format "raw" -ticks true
```

> the code above ticks along at a second like it should. but if you change the `reduce -every` interval to :30m: pace ticks along at increasingly slow intervals of :10s: :20s: :30s: etc.

Actually, the problem is `pace`'s deceleration to 'realtime rate', which with live would become `:30m:`. The deceleration is there even with `:m:`, but for this interval, it introduces almost unobservable delay to the `pace`'s rate of 1s (about 1ms).

Original bug report proposed this behaviour to be optional. As this feature is no longer relevant and is unused, let's remove it instead.

PR #609, opened 2016-03-16
pace.js - remove decelerate towards realtime rate feature

bkutil:
This was used in Jut 1.0 times and is not that relevant anymore.

As the logic is complex, the behavior unintuitive (and there is also a
bug with unitialized `data_t0` when `-from` is not passed), drop the
feature.

Fix: #608


bkutil:
cc: @dmajda

dmajda:
+1

Issue #565, opened 2016-03-09
Ugly runtime errors handling


dmajda:
Runtime errors are handled ([here](https://github.com/juttle/juttle/blob/f5a4ad7d6c087dd67fc8b2fb13b3610417f0cc7f/bin/juttle#L177-L179) and [here](https://github.com/juttle/juttle/blob/f5a4ad7d6c087dd67fc8b2fb13b3610417f0cc7f/bin/juttle#L185-L187)) by printing the error message and the error itself to console. This leads to ugly error messages like the one reported in #562.

Runtime errors should be handled by the same mechanism like compile errors. In particular, Juttle errors should be displayed with source excerpt.

dmajda:
The last one for @mstemm.

PR #611, opened 2016-03-16
Use instanceOf JuttleError to identify Juttle Errors.

### LINKED ISSUE #563: Juttle CLI detects Juttle errors incorrectly
Juttle CLI [assumes](https://github.com/juttle/juttle/blob/f5a4ad7d6c087dd67fc8b2fb13b3610417f0cc7f/bin/juttle#L209-L212) that any error with location info is a Juttle error and any error without it is not. This assumption is invalid. Juttle errors are those descending from `JuttleError`, all other errors are internal errors (and should generally be treated as bugs). Most Juttle errors have location info, but some do not (see e.g. #519).

The CLI needs to be fixed to detect Juttle errors using `instanceof` and display them correctly whether they contain location info or not.

mstemm:
Instead of assuming that any error with a location is a Juttle Error,
use instanceof JuttleError instead.

This means we should only call show_in_context when the error has a
location, and just generically print the error otherwise.

This fixes #563.

@dmajda @rlgomes

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -43,6 +43,7 @@ var URLResolver = require('../lib/module-resolvers/url-resolver');
 var FileResolver = require('../lib/module-resolvers/file-resolver');
 var resolver_utils = require('../lib/module-resolvers/resolver-utils');
 var compiler = require('../lib/compiler');
+var errors = require('../lib/errors');
 
 var modes = _.without(compiler.stageNames, 'eval').concat('run');
 function usage() {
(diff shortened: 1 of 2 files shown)


dmajda:
This PR doesn’t solve #563 completely. Description of a Juttle error without a location is still displayed differently from a description of a Juttle error with a location:

```
$ bin/juttle -e 'emit -limit 1 + null'
RuntimeError: Invalid operand types for "+": number (1) and null.
```

I’d expect:

```
$ bin/juttle -e 'emit -limit 1 + null'
Error: Invalid operand types for "+": number (1) and null. (TYPE-ERROR)
```

The only difference in presentation of Juttle errors with and without a location should be the presence of a source excerpt.

mstemm:
Ok, I've moved display of errors into a common function so it can be used in both locations, as well as thinking ahead to #565. Please look again?

dmajda:
+1, please squash before merging

PR #614, opened 2016-03-17
fix stdio adapter to stream correctly

### LINKED ISSUE #613: read stdio is not streaming points through the flowgraph
Noticed this when running:

```
while true; do df; sleep 5; done | bin/juttle -e "read stdio -format 'columns' | put time = Date.time() | view text"
```

Basically the issue is that we didn't apply the same `Iterator` approach we had used in the file adapter to make sure we push out points every so often and don't simply buffer them all before returning them from the `read` function.

rlgomes:
fixes #613

quick fix to use the same Iterator approach to fix the `stdio` adapter
so it can stream data correctly.

### DIFF
--- a/lib/adapters/stdio/read.js
+++ b/lib/adapters/stdio/read.js
@@ -4,6 +4,77 @@ var _ = require('underscore');
 var AdapterRead = require('../adapter-read');
 var parsers = require('../parsers');
 var errors = require('../../errors');
+var JuttleMoment = require('../../runtime/types/juttle-moment');
+
+class Iterator {
+    constructor(adapter, from, to) {
+        this.logger = adapter.logger;
+
+        var stream = adapter.getStdin();
+        this.buffer = [];
+        this.eof = false;
+        this.error = null;
+        this.pending = null;
+
+        adapter.parser.parseStream(stream, (points) => {
+            points = adapter.parseTime(points, { timeField: adapter.timeField });
+            points = adapter.filterPoints(points, from, to);
+
+            this.logger.debug('parsed', points.length, 'points');
+            this.buffer = this.buffer.concat(points);
+
+            // go back to the event loop once before notifying out a response to
+            // avoid returning from read once for the last batch of points and
+            // then again once the parser promise resolves
+            setTimeout(() => { this.notify(); }, 0);
+        })
+        .then(() => {
+            this.eof = true;
+            this.notify();
+        })
+        .catch((err) => {
+            this.error = err;
+            this.notify();
+        });
+    }
+
+    read(from, to, limit, state) {
+        return new Promise((resolve, reject) => {
+            this.pending = {
+                resolve, reject
+            };
+
+            if (this.error || this.buffer.length !== 0 || this.eof) {
+                this.resolve();
+            }
+        });
+    }
+
+    notify() {
+        if (this.pending) {
+            this.resolve();
+        }
+    }
+
+    resolve() {
+        this.logger.debug('resolve: err:', this.error,
+                          'buffer:', this.buffer.length, 'eof:', this.eof);
+        var pending = this.pending;
+        this.pending = null;
+
+        if (this.error) {
+            pending.reject(this.error);
+        } else {
+            var points = this.buffer;
+            this.buffer = [];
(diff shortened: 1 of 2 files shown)


rlgomes:
@demmer @davidvgalbraith stdio reading while streaming. I am still going to think about some way to verify that an adapter is truly streaming so we don't run into this again as we did with the `file` adapter. I verified by simply running `while true; do df ; done | ./bin/juttle -e "read stdio -format 'grok' | view text"` and seeing no output before the fix and immediately seeing output with the fix in place. Do you guys think I should move `Iterator` code into some common location since its used now in at least 3 places: http, stdio and file adapter and I think it could generalized or do that in a follow up PR ?

davidvgalbraith:
Is there no way to hack out a unit test similar to your manual test?

rlgomes:
@davidvgalbraith let me see what I can come up with because this has bitten us a few times now.

rlgomes:
@davidvgalbraith what do you think of this test ? Is it a little too cryptic to follow how it validates that the adapter is in fact streaming ? I can do a follow up PR to apply similar approach to `file` and `http` and verify they are in fact streaming points as well.

lib/adapters/stdio/read.js
```diff
@@ -4,6 +4,78 @@ var _ = require('underscore');
...
 var parsers = require('../parsers');
 var errors = require('../../errors');
+var JuttleMoment = require('../../runtime/types/juttle-moment');
+
+class Iterator {
+    constructor(adapter, from, to) {
+        this.logger = adapter.logger;
+
+        var stream = adapter.getStdin();
+        this.adapter = adapter;
```
davidvgalbraith: It doesn't look like you use `this.adapter`.

lib/adapters/stdio/read.js
```diff
@@ -4,6 +4,78 @@ var _ = require('underscore');
...
+    }
+
+    read(from, to, limit, state) {
+        return new Promise((resolve, reject) => {
+            this.pending = {
+                resolve, reject
+            };
+
+            if (this.error || this.buffer.length !== 0 || this.eof) {
+                this.resolve();
```
davidvgalbraith: Where does `this.resolve` get set?

davidvgalbraith: Oh duuuh it's the class method right there

test/adapters/stdio/read.stdio.spec.js
```diff
@@ -318,4 +318,42 @@ describe('read stdio adapter tests', function() {
             ]);
         });
     });
+
+    it('can read stream data correctly', () => {
+        juttle_test_utils.set_stdin({
+            on: (event, callback) => {},
+
+            pipe: (destination, options) => {
```
davidvgalbraith: lol wut

rlgomes:
@davidvgalbraith @demmer now with a test that works although I had to figure out how to generate enough points to have the parser actually call `emit` which happened to be more than 1024 default we have hard coded here https://github.com/juttle/juttle/blob/master/lib/adapters/parsers/base.js#L13. Which brings up the question of we want the parsers to behave like this ? Its a bit tricky since you don't want the parser to emit on every single new point but at the same time you don't want it to buffer `1024` points before it emits if its only parsing 10 points per second because the inbound stream is slow and with the buffer not filling the "streaming" behavior of our pipeline is far from really streaming.  Thoughts on what would be best to do here ?

davidvgalbraith:
Can we `drain()` every 5 seconds or whatever, regardless of how many points we have buffered?

davidvgalbraith:
I like the new test!

rlgomes:
@davidvgalbraith thats why I was wondering if we could drain when we hit the limit or when a `setTimeout(drain, 0)` gets time to drain. Which is similar to waht we're doing in the Iterator code and why I'm wondering if it make sense to duplicate this logic ? @demmer you probably have a better idea on this one.

rlgomes:
@demmer @davidvgalbraith what shall we do with this ? close ? the streaming fix here does in fact fix the issue and the test albeit complicated tests the expectation correctly.

davidvgalbraith:
+1

PR #615, opened 2016-03-17
Document string and object functions

davidvgalbraith:
finishes https://github.com/juttle/juttle/issues/58

@dmehra @dmajda @bkutil

### DIFF
--- a/docs/modules/string.md
+++ b/docs/modules/string.md
@@ -53,7 +53,7 @@ These are the basic string coercion rules:
     "myvalue", use `-option 'myvalue'`.
 
 - Arguments for reducer calls are not subject to string coercion
-    
+
     For example, if your program includes `reduce count(host)` then host
     must be defined elsewhere in your program. If you want to reduce by
     counting the instances of each value of your data's host field, use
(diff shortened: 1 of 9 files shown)


docs/modules/string.md
```diff
@@ -103,11 +103,51 @@ regular expression, like this:
 
 ***
 
+## String.charAt
+
+Given a string and an index, returns the character at that index in the string.
```
bkutil: What is the return value for out-of-bounds index? Are negative indices supported?

docs/modules/string.md
```diff
@@ -124,6 +164,25 @@ _Example_
 
 ***
 
+## String.fromCharCode
```
bkutil: Not related to this PR, but shouldn't we use [fromCodePoint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint) and `codePointAt` instead, to handle higher code points as well? Both are available since at least node 4.2.6.

davidvgalbraith:
Bump! Our customer had to learn about `Array.push` by searching the source code today.

docs/modules/string.md
```diff
@@ -103,11 +103,51 @@ regular expression, like this:
...
+
+Parameter  |  Description  |  Required?
+---------- | ------------- | ---------:
+`string`   | The string to return a character from   |  Yes
+`index`   | The index in `string` of the character code to return|  Yes
+
+_Example_
+
+```
+{!docs/examples/modules/string_char_at.juttle!}
```
dmehra: This is string_char_code_at

dmehra:
Sorry, missed this, will fix the example name, merge and publish now

PR #616, opened 2016-03-17
Rewrite mutability checks in the semantic pass

dmajda:
This PR rewrites mutability checks in the semantic pass as symbol checks. This is the right thing conceptually, it leads to more straightforward code, it avoids repeated symbol lookups, and it paves the path for allowing nested member assignments.

Part of work on #419.

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -1332,12 +1332,6 @@ class SemanticPass {
                     location: node.location
                 });
             }
-            if (!this.scope.is_mutable(node.name, this.scope.type)) {
-                throw errors.compileError('VARIABLE-NOT-ASSIGNABLE', {
-                    name: node.name,
-                    location: node.location
-                });
-            }
         }
 
         if (this.expr_mode === 'call') {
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
First commit message typo: `scceed`

lib/compiler/semantic.js
```diff
@@ -1402,7 +1402,7 @@ class SemanticPass {
         });
 
         if (this.expr_mode === 'assign') {
-            if (!this.scope.is_mutable(node.object.name, this.scope.type)) {
+            if (!node.object.symbol || node.object.symbol.type !== 'var') {
```
bkutil: Would it make sense to introduce a helper with some descriptive name for this condition?

bkutil:
see comments, +1

lib/compiler/semantic.js
dmajda: Not at this point — this code will be subject to further tweaking and refactoring soon. But later probably yes.

dmajda:

PR #619, opened 2016-03-17
Revert reduce every emits marks

### LINKED ISSUE #594: Make -every not emit marks (again)
This is in question / maybe-bug / gotta figure this out category. When we recently made a change to pass the batch marks from an upstream `reduce -every` to downstream procs, some cases were not considered, that now showed up in practice, such as:

- if upstream reduce has `-every` but downsteam of it there is a reduce without `-every` that's meant to aggregate all the data, it now aggregates on the same `-every` interval as upstream, and requires introduction of `unbatch` into the program to get back to the intended behavior. I had to do this in [postgres-diskstats/throughput.juttle](https://github.com/juttle/juttle-engine/pull/87/files#diff-9a9a1b96627402f2d5161a5726c3212c) example. (Why the two-level reduce: because there's a lot of raw data in the backend, so I'm doing a `reduce -every :hour:` on read that gets optimized with the read, and then i fork the program to do something with hourly statistics, and also aggregate across.)

- similarly if upstream reduce has `-every` but downstream of it you want to sort across all data, `reduce -every :day: count() | sort count` to get top daily counts, that no longer works and needs to be `reduce -every :day: count() | unbatch | sort count`. 

Those are common use cases, and we have made them more complex for the user, who now has to be aware of `unbatch` even though they never used `batch`.

The original idea was to simplify the case where the user continues doing calculations for the same batch interval, such as `reduce -every :hour: ... | head 1` and means to get the head of every hour. That is still a case we want to support, but not at the expense of complicating the also-common cases above. 

Options:

(1) Drop the support for `-every` and instead use `batch` (and explicit `unbatch` where needed). See @go-oleg.

(2) Introduce parameter `-all true` (unless there's a better name) that acts as `unbatch` but applies to reduce, head, sort and anything that supports `-every`, so in the sort example above, the user would write `reduce -every :day: count() | sort -all true count`. I wish we didn't have to give it 'true' and could just say `sort -all count`. This was @dmehra's vote. 

(3) Similar to (2) but use `-every :forever:` which should be already supported logically (the parser barfs today). Suggested by @demmer but perhaps not seriously.

(4) Rename `unbatch` to something friendlier. @go-oleg said `unevery`. Maybe even friendlier. Don't know what.

(5) Go back to not passing the batch marks downstream. That will break some use case we wanted to make easy, but we (@dmehra, @go-oleg) failed to remember exactly what that was.

For your consideration, @demmer @dmajda and general interest @juttle/developers .

bkutil:
Revert 'reduce -every emits marks' work. Should fix #594.

Commits which are plain reverts reference the reverted ones, those which had to be updated do not.

### DIFF
--- a/lib/runtime/procs/reduce.js
+++ b/lib/runtime/procs/reduce.js
@@ -97,7 +97,6 @@ class reduce_base extends periodic_fanin {
         if (this.to && this.to.lte(time)) {
             var out = this.advance_epoch(this.to);
             this.emit(out);
-            this.emit_mark({ time: this.to });
         }
         return false;
     }
@@ -318,18 +317,11 @@ class reduce_every extends reduce_base {
         return 'reduce';
     }
     first_epoch(epoch) {
-        // emit initial mark (before any points arrive)
-        this.emit_mark({ time: epoch });
         this.after_first_epoch = true;
     }
     //
     // mark() and process() are from periodic. tick() from reduce_runner_funcs
     //
-
-    other_epoch(epoch) {
-        super.other_epoch(epoch);
-        this.emit_mark({ time: epoch });
-    }
 
     eof(from) {
         // Edge case - 0 points followed immediately by eof. To be consistent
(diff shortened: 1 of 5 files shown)


bkutil:
cc: @dmajda @demmer @dmehra

dmajda:
+1

PR #620, opened 2016-03-17
Make the “[]” operator work properly in lvalues

dmajda:
After some preparatory work, this PR adds tests that specify behavior of the `[]` operator in lvalues and adjusts the implementation to match. Specifically, it adds missing type checks and allows modification of expressions not associated with a symbol (which enables nested member assignment).
    
Note that member assignment in `put`/`reduce` still isn't allowed. Making it work requires more substantial changes which will be done later.

Part of work on #419.


dmajda:
@bkutil

bkutil:
Looks good +1

PR #622, opened 2016-03-17
table: rework progressive mode to wait before emitting the first points

### LINKED ISSUE #621: cli table view should wait a bit to calculate column widths
Currently the CLI table view figures out the set of columns to show and the widths based only on the first point that arrives. This is suboptimal, especially in simple cases where the output of the program is a single batch in which all the points arrive at virtually the same time, yet may have different columns and/or different lengths of values for each column.

Instead a better heuristic would be to delay for a certain amount of time before rendering the table, such that all the points that arrive in that timeframe are considered when computing the columns.

demmer:
Instead of computing the columns and their widths based only on the first
point that arrives, delay for a configurable amount for more points to
arrive.

As part of this change, remove the 1.5x padding on the widths and also
make things consistent so both progressive and non-progressive mode have
a minimum width of 10 characters.

Fixes #621


demmer:
@mstemm For the code review

@dmehra I know you've run into issues where the table doesn't do a good job of picking default column widths. Can you try this branch and see if it works better for your use cases.

mstemm:
Can you update the documentation at docs/sinks/view_table.md?

mstemm:
The other code changes look good.

dmehra:
I tried data in #33 and this branch gives nice output, not too wide, and doesn't cut off fields. For example
```
emit -points [ { name: 'Short'}, {name: 'Long long long long long long long long long long long long long long' } ] | put time = :now:
```
Works fine on this branch, but not on master.
+1

demmer:
@mstemm @dmehra I updated the description in the docs. This one turned out to be kinda tricky to describe, so please take a look and let me know if you have any better way of describing it.

mstemm:
I think the text is pretty accurate for a summary designed to fit in the table but maybe it's time to have a dedicated paragraph outside the table which you could just refer to.

dmehra:
Lgtm on the description, but it prompts a question, is -progressive true at all possible with multiple views? If you override the default `false` setting? Seems like it would be a mess of output.

demmer:
@dmehra if a user chooses to shoot themselves in the foot by explicitly passing `-progressive true` in a program with multiple views, then we don't stop them from seeing a garbled mess of interleaved tables, and in fact there's a unit test that verifies we do exactly that.

This PR doesn't change that behavior, nor should it IMO.

PR #623, opened 2016-03-17
Add adapter modules

### LINKED ISSUE #325: Add support for per-adapter modules
This came up in the context of the aws adapter, where the current implementation did automatic aggregations of raw per-item information into aggregates (number of instances), demographics (breakdowns of instances by instance type), etc.

Although useful, it ends up hiding the raw per-item information and isn't in line with the data format returned by our other adapters.

One solution to this would be to define a module that created subgraphs that performed the aggregations (i.e. ec2_demographics() would be a | reduce count=count() by demographic). Juttle users who wanted these automatic rollups could just use the module. However, we don't have a great way to pass around modules in a way they can easily be used by all juttle users.

To address this, how about we allow adapters to publish modules along with their read and write procs? They could be namespaced as Adapter.<adapter_name>.*, and loaded into the juttle runtime when the adapter is loaded. This would allow us to provide useful aggregation methods, constants, etc along with each adapter that could make the raw data easier to use.

mstemm:
This adds support for per-adapter modules.

Should I bump the juttle api version for this? Following semver we'd want it to be 0.6.0, but maybe we could cheat and make it 0.5.3 instead.

After this merges I'll update the aws adapter and adapter template to provide examples.

This fixes #325.

@demmer @dmajda

### DIFF
--- a/lib/module-resolvers/file-resolver.js
+++ b/lib/module-resolvers/file-resolver.js
@@ -36,6 +36,23 @@ class FileResolver {
         }
     }
 
+    _read_file(filename, name) {
+
+        if (! this._is_file(filename)) {
+            return false;
+        }
+
+        try {
+            return {
+                source: fs.readFileSync(filename, 'utf8'),
+                name: (name ? name : filename)
+            };
+        }
+        catch (e) {
+            return false;
+        }
+    }
+
     _canonicalize(filename) {
         // If the filename is actually a directory, add a
         // /index.juttle
@@ -78,15 +95,8 @@ class FileResolver {
         // absolute.
         let filename = path.resolve(dir, module_path);
         filename = this._canonicalize(filename);
-        try {
-            return {
-                source: fs.readFileSync(filename, 'utf8'),
-                name: filename
-            };
-        }
-        catch (e) {
-            return false;
-        }
+
+        return this._read_file(filename);
     }
 
     _search_system_import(module_path, module_name, importer_path) {
@@ -99,16 +109,10 @@ class FileResolver {
 
             filename = this._canonicalize(filename);
 
-            if (this._is_file(filename)) {
-                try {
-                    return {
-                        name: module_path,
-                        source: fs.readFileSync(filename, 'utf8')
-                    };
-                }
-                catch (e) {
-                    // continue looking for a module we can successfully load
-                }
+            let module = this._read_file(filename, module_path);
+
+            if (module) {
+                return module;
             }
         }
         return false;
--- a/lib/module-resolvers/file-resolver.js
+++ b/lib/module-resolvers/file-resolver.js
@@ -1,10 +1,14 @@
 'use strict';
 
 /* eslint-env node */
+var _ = require('underscore');
 var fs = require('fs');
 var path = require('path');
 var Promise = require('bluebird');
 var logger = require('../logger').getLogger('file-resolver');
+var adapters = require('../runtime/adapters');
+
+const ADAPTER_MODULE_SUBDIR = 'juttle';
 
 function search_paths() {
     var env = process.env;
@@ -24,6 +28,28 @@ class FileResolver {
 
         this._additional_search_paths = options.search_paths || search_paths();
 
+        // Also add the module root for all adapters.
+        this._adapter_roots = [];
+
+        for (let adapter of adapters.list()) {
+            // Only consider non built-in adapters that were loaded successfully.
+            if (! adapter.builtin && adapter.loaded) {
+                let install_path = this._canonicalize(adapter.path);
+                let module_root = [path.dirname(install_path), ADAPTER_MODULE_SUBDIR].join(path.sep);
+
+                // Only add the entry to adapter_roots if it exists and is a directory.
+                if (fs.existsSync(module_root)) {
+                    let stats = fs.statSync(module_root);
+                    if (stats.isDirectory()) {
+                        this._adapter_roots.push({
+                            adapter: adapter.adapter,
+                            module_root: module_root
+                        });
+                    }
+                }
+            }
+        }
+
         this.resolve = this._resolve.bind(this);
     }
 
@@ -37,7 +63,6 @@ class FileResolver {
     }
 
     _read_file(filename, name) {
-
         if (! this._is_file(filename)) {
             return false;
         }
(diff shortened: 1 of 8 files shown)


lib/module-resolvers/file-resolver.js
```diff
@@ -24,6 +27,22 @@ class FileResolver {
 
         this._additional_search_paths = options.search_paths || search_paths();
 
+        // Also add the module root for all adapters.
+        this._adapter_roots = [];
+
+        for(let adapter of adapters.list()) {
+            // Skip built-in adapters or adapters for which the
+            // install path is not absolute (i.e. not actually available)
+            if (adapter.module !== '(builtin)' &&
```
dmajda: Better check this with `adapter.builtin`.

lib/module-resolvers/file-resolver.js
```diff
@@ -24,6 +27,22 @@ class FileResolver {
...
+                path.isAbsolute(adapter.path)) {
```
dmajda: I think it would be better to include a flag indicating if the adapter was actually loaded (e.g. `adapter.loaded`) and use it here. Checking the path is definitely not a clean check adapter’s state.

lib/module-resolvers/file-resolver.js
```diff
@@ -24,6 +27,22 @@ class FileResolver {
...
+                let install_path = this._canonicalize(adapter.path);
+                this._adapter_roots.push({
+                    adapter: adapter.adapter,
+                    module_root: [path.dirname(install_path), ADAPTER_MODULE_SUBDIR].join(path.sep)
+                });
```
dmajda: Maybe add the path only when it exists and it is a directory? Otherwise we will unnecessarily check it on every module load.

lib/module-resolvers/file-resolver.js
```diff
@@ -36,6 +55,23 @@ class FileResolver {
...
+            return {
+                source: fs.readFileSync(filename, 'utf8'),
+                name: (name ? name : filename)
+            };
+        }
+        catch (e) {
+            return false;
+        }
+    }
+
```
dmajda: The `_read_file` extraction would be better in a separate commit (it is orthogonal with the other changes).

lib/module-resolvers/file-resolver.js
```diff
@@ -78,20 +114,30 @@ class FileResolver {
...
     _search_system_import(module_path, module_name, importer_path) {
         logger.debug(`searching system module_path=${module_path} module_name=${module_name} importer_path=${importer_path}`);
 
+        // First see if the module_path matches one of the adapter
+        // roots. If it does, prefix the adapter's module root, drop
+        // the adapter name from the module path, and try to read the
+        // file.
+        let parts = module_path.split(path.sep);
+        let first_subdir = parts.shift();
+        for(let root of this._adapter_roots) {
```
dmajda: Use `_.findWhere` instead of a `for` loop?

dmajda:
Beside the minor comments inline, I have two bigger notes:

  1. I think we should prefix adapter module paths with a common prefix, e.g. `adapters`. So instead of:

    ```juttle
    import "aws/aggregations" as Adapter_aws;
    ```

   one would do:

    ```juttle
    import "adapters/aws/aggregations" as Adapter_aws;
    ```

    My main concern is that I can imagine stdlib growing subdirectories at some point and it wouldn’t be good to thare that namespace with adapters (which would then be able to shadow stdlib stuff or vice versa).

  2. I don’t like the `module_root` directory name, it wouldn’t say anything to me if I saw adapter code for the first time. I’d prefer `juttle`, `juttle-modules`, or `modules` (in that order); there may be some other good names too.

@demmer I’d like to have your opinion on both points.

dmajda:
> Should I bump the juttle api version for this? Following semver we'd want it to be 0.6.0, but maybe we could cheat and make it 0.5.3 instead.

I think so (we are providing a new capability), I’d go for 0.6.0 (I don’t see a reason for “cheating”).

mstemm:
I like the top-level adapters directory, I'll do that.

@demmer didn't like module_root either but didn't have any good alternatives. I like most of yours better. He also wanted the name juttle in the name somewhere so I think I'll use juttle-modules.

mstemm:
My understanding of semver for versions < 1.0 is that minor number bumps are not considered compatible with lower minor numbers. So if we add this (really backwards-compatible) feature and increment the minor version we'd have to bump the adapter api version of all adapters.

demmer:
>> I don’t like the module_root directory name, it wouldn’t say anything to me if I saw adapter code for the first time. I’d prefer juttle, juttle-modules, or modules (in that order); there may be some other good names too.

> @demmer didn't like module_root either but didn't have any good alternatives. I like most of yours better. He also wanted the name juttle in the name somewhere so I think I'll use juttle-modules.

I can live with `juttle-modules` but I'd slightly prefer the simpler
`juttle` or `juttle-lib`.


On Fri, Mar 18, 2016 at 9:32 AM, Mark Stemm <notifications@github.com> wrote:
> I like the top-level adapters directory, I'll do that.
>
> @demmer didn't like module_root either but didn't have any good
> alternatives. I like most of yours better. He also wanted the name juttle in
> the name somewhere so I think I'll use juttle-modules.
>
> —
> You are receiving this because you were mentioned.



-- 
Michael Demmer
Jut - CTO / VP Engineering
demmer@jut.io

mstemm:
I don't like plain 'juttle' but it's at the top of both of your lists and I want this done, so I'll switch to juttle.

dmajda:
> My understanding of semver for versions < 1.0 is that minor number bumps are not considered compatible with lower minor numbers. So if we add this (really backwards-compatible) feature and increment the minor version we'd have to bump the adapter api version of all adapters.

One thing is what the semver spec itself says, which is basically [“all bets are off in < 1.0.0”](http://semver.org/#spec-item-4). From that standpoint, we shouldn’t use version ranges in adapters’ `package.json` at all.

Another thing is what the `^` operator does in Node.js (or, more precisely, in `semver.satisifes`). This is basically what you say and yes, we’d need to bump if we’d go to 0.6.0 (or 0.7.0 to correspond to the next release).

Reading the [`^` docs](https://github.com/npm/node-semver#caret-ranges-123-025-004), it looks like the idea is basically that the version semantics is shifted by one step in the < 1.0.0. This would correspond to your 0.5.3 suggestion.

Considering all this, I think you are right and we should just go with 0.5.3 to avoid the hassle of updating all adapters.

lib/module-resolvers/file-resolver.js
```diff
@@ -36,6 +36,23 @@ class FileResolver {
         }
     }
 
+    _read_file(filename, name) {
+
```
dmajda: Nit: Omit the empty line?

lib/module-resolvers/file-resolver.js
```diff
@@ -24,6 +28,29 @@ class FileResolver {
 
         this._additional_search_paths = options.search_paths || search_paths();
 
+        // Also add the module root for all adapters.
+        this._adapter_roots = [];
+
+        for(let adapter of adapters.list()) {
```
dmajda: Nit: Space after `for`.

lib/module-resolvers/file-resolver.js
```diff
@@ -24,6 +28,29 @@ class FileResolver {
...
+            // Only consider non built-in adapters that were loaded successfully.
+            if (! adapter.builtin &&
+                adapter.loaded) {
```
dmajda: Nit: I think the condition can just be on one line.

lib/module-resolvers/file-resolver.js
```diff
@@ -99,9 +126,40 @@ class FileResolver {
         return this._read_file(filename);
     }
 
+    _search_adapter_import(module_path, module_name, importer_path) {
+
+        // Remove the top 'adapters' directory
+        let parts = module_path.split(path.sep).slice(1);
```
dmajda: Hmm, this makes Juttles incompatible between Unix and Windows because one would have to use `\` as a separator on Windows. I guess we just need to hard-code `/` here.

lib/module-resolvers/file-resolver.js
```diff
@@ -99,9 +126,40 @@ class FileResolver {
...
+
+        return false;
+    }
+
     _search_system_import(module_path, module_name, importer_path) {
         logger.debug(`searching system module_path=${module_path} module_name=${module_name} importer_path=${importer_path}`);
 
+        // If the module_path starts with 'adapters' + pathsep, assume
+        // it's an adapter module load.
+        if (module_path.startsWith('adapters' + path.sep)) {
```
dmajda: Same here (hard-code `/`).

dmajda:
+1 with the comments addressed (and sorry I didn’t catch some of these during the first round)

PR #624, opened 2016-03-18
modules: support importing procs from native js modules

demmer:
This adds rudimentary support for extending the runtime with procs
implemented in javascript by supporting `import "/path/to/file.js"`
in which the file exports an object containing the proc implementation.

To make this work, the module loader looks at the file extension, and for
.js files it simply inlines the source and returns it.

Then the semantic pass evals the blob of code and puts all the symbols from
the exported `procs` into scope.

The flowgraph builder generates a new type of natice proc call node and also
passes through the native modules so they can be carried through to the
built flowgraph.

Finally, the module source is inlined into the compiled code so it gets
evaluated again as part of the program itself.

### DIFF
--- a/lib/compiler/build.js
+++ b/lib/compiler/build.js
@@ -57,7 +57,11 @@ class Build extends CodeGenerator {
         var self = this;
 
         _.each(ast.modules, function(ast) {
-            self.gen_ModuleDef(ast);
+            if (ast.type === 'ModuleDef') {
+                self.gen_ModuleDef(ast);
+            } else if (ast.type === 'NativeModuleDef') {
+                self.gen_NativeModuleDef(ast);
+            }
         } );
 
         this.gen_SubDef(ast, {path_info: {mod: '', sub: []}});
@@ -162,6 +166,23 @@ class Build extends CodeGenerator {
         this.emit(code);
         return node.uname;
     }
+    gen_native_proc_call(node) {
+        var code;
+        var proc = {
+            type: 'NativeModuleProc',
+            module: node.module_uname,
+            name: node.op.property.value,
+            uname: node.uname,
+            options: node.options
+        };
+        proc.pname = {type:'Eval', expr:'builder.alloc_pname()' };
+        code = 'var ' + proc.uname + ' = ';
+        this.build_options(proc);
+        code += this.build_code_ast(proc) + ';';
+        code += 'builder.add_node(' + proc.uname + ');\n';
+        this.emit(code);
+        return proc.uname;
+    }
     build_options(node) {
         var that = this;
         node.options = _.map(node.options, function(option) {
@@ -191,6 +212,7 @@ class Build extends CodeGenerator {
             case 'SequentialGraph':
             case 'ImportStatement':
             case 'ParallelGraph':
+            case 'NativeProcDef':
                 return this['gen_' + node.type](node, opts);
             default:
                 throw new Error('unrecognized statement type ' + node.type);
@@ -203,6 +225,18 @@ class Build extends CodeGenerator {
         // with unames
         //
         var self = this;
+        _.each(node.elements, function(stmt) {
+            self.gen_statement(stmt, {path_info: {mod: node.name, sub: []}});
+        } );
+    }
+    gen_NativeModuleDef(node) {
+        //
+        // just spit out the module elements in the main scope since
+        // the semantic pass flattened out all the module contents
+        // with unames
+        //
+        var self = this;
+        this.emit(`var ${node.uname} = builder.native_module("${node.uname}", "${node.name}", ${JSON.stringify(node.source)});\n`);
         _.each(node.elements, function(stmt) {
             self.gen_statement(stmt, {path_info: {mod: node.name, sub: []}});
         } );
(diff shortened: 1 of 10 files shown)


demmer:
@dmajda @go-oleg @mccanne This is a rough first cut but it works for simple cases :)

One major constraint is that the loaded files can't do a require() or anything like that -- all the code must be in the single file and can assume that `juttle` is in scope.

The other approach would be to `require()` it dynamically as part of the program and stash it in the runtime like we do for adapters.

go-oleg:
Sweeeet!

`require()` works fine for me. In fact, it looks like the module gets `eval()`'d twice...I added this to the top of it:

```javascript

_.each(['a','b'], (item) => { console.log(item); });
```

and saw
```
a
b
a
b
```

Did you mean that you can't `require()` with relative paths?

lib/compiler/build.js
```diff
@@ -191,6 +212,7 @@ class Build extends CodeGenerator {
             case 'SequentialGraph':
             case 'ImportStatement':
             case 'ParallelGraph':
+            case 'NativeProcDef':
```
dmajda: Looks like a mistake.

lib/compiler/build.js
```diff
@@ -207,6 +229,18 @@ class Build extends CodeGenerator {
...
+        //
+        // just spit out the module elements in the main scope since
+        // the semantic pass flattened out all the module contents
+        // with unames
+        //
+        var self = this;
+        this.emit(`var ${node.uname} = builder.native_module("${node.uname}", "${node.name}", ${JSON.stringify(node.source)});\n`);
+        _.each(node.elements, function(stmt) {
+            self.gen_statement(stmt, {path_info: {mod: node.name, sub: []}});
+        } );
```
dmajda: The loop over `node.elements` is not needed (native module doesn’t have any).

lib/compiler/build.js
```diff
@@ -524,7 +558,13 @@ class Build extends CodeGenerator {
         return this.gen_ReifierProc(node);
     }
     gen_FunctionProc(node) {
-        return this.gen_sub_call(node);
+        if (node.uname_sub) {
```
dmajda: This is fragile, I’d rather switch on `node.op.symbol.type`.

lib/compiler/build.js
```diff
@@ -162,6 +166,23 @@ class Build extends CodeGenerator {
         this.emit(code);
         return node.uname;
     }
+    gen_native_proc_call(node) {
+        var code;
+        var proc = {
+            type: 'NativeModuleProc',
+            module: node.module_uname,
```
dmajda: I’d name this property `module_uname`. (The uname-less name confused me when reading `GraphCompiler` code before this part. All (or almost all) variables/properties which contain unames have `uname` somewhere in their names and I think it’s a good convention to prevent confusion with source-level names, which we also pass around all the time.)

lib/compiler/build.js
```diff
@@ -162,6 +166,23 @@ class Build extends CodeGenerator {
...
+            name: node.op.property.value,
+            uname: node.uname,
+            options: node.options
+        };
+        proc.pname = {type:'Eval', expr:'builder.alloc_pname()' };
```
dmajda: Fold into the literal above?

lib/compiler/semantic.js
```diff
@@ -847,7 +892,7 @@ class SemanticPass {
 
         if (module_ast !== undefined) {
             if (!this.modules.hasOwnProperty(modulename)) {
-                this.sa_ModuleDef(module_ast);
+                this.sa_Module(module_ast);
```
dmajda: This should be `sa_module` or better `sa_module_def` to not look like an AST node handler.

lib/compiler/semantic.js
```diff
@@ -555,6 +578,28 @@ class SemanticPass {
...
     }
+    sa_NativeModuleDef(node) {
+        var code = '(function (juttle) {\n' +
+            'var module = {};\n' +
+            node.source + '\n' +
+            'return module.exports; })';
+        var module = eval(code)(runtime);
+        node.source = code;
+        var scope = new Scope(null, 'module');
+        node.elements = [];
```
dmajda: Setting `node.elements` should not be needed as there are no Juttle statements inside a native module.

lib/compiler/semantic.js
```diff
@@ -555,6 +578,28 @@ class SemanticPass {
...
+        node.uname = this.scope.alloc_var('module');
+        _.each(module.procs, (code, name) => {
+            this.enter_scope('sub');
+            var symbol = scope.define_native_proc(name, [], true);
+            symbol.module_uname = node.uname;
+            this.exit_scope();
```
dmajda: The body of the loop can be simplified to:

```js
scope.define_native_proc(name, [], true, node.uname);
```

with appropriately extended definition of `define_native_proc`.

lib/compiler/scope.js
```diff
@@ -110,6 +110,12 @@ class Scope {
         this.put(name, symbol);
         return symbol;
     }
+    define_native_proc(name, args, exported) {
+        var runtimeVar = this.alloc_var(name);
+        var symbol = { type: 'native_proc', exported: exported, uname: runtimeVar, args: args };
```
dmajda: It may make sense to do something like `type: 'proc', native: true`. In the future, we’ll probably also want native modules to export consts, functions, and reducers and we certainly won’t want to introduce new types like `native_const`, just a flag so that later stages can treat them differently if necessary.

dmajda:
> The other approach would be to `require()` it dynamically as part of the program and stash it in the runtime like we do for adapters.

I think that approach would make much more sense in light of the limitations and negative effects of double-eval.

(I still commented on the PR because most of its code will probably get used with any approach.)

Issue #618, opened 2016-03-17
Make the “interval” field speical


dmajda:
Make the `interval` field a special field like the `time` field. Specifically, don’t allow anything but a non-negative duration there.

bkutil:
I.e. the offending point should be dropped and a warning emitted?

dmajda:
Yes.

PR #625, opened 2016-03-18
oops-fanin.js - type check assignments to interval

bkutil:
Allow assignments to interval field, but ensure it is a positive
duration.

Fix: #618


bkutil:
Depends on #581 merge.

cc: @dmajda

dmajda:
+1

PR #629, opened 2016-03-20
read: quantize from/to for optimized reduce -every

davidvgalbraith:
If we're doing an optimized reduce -every live program, we should only make one query per
batch. We start by going from from to the end of the first batch, then each
subsequent query is made over a full batch.
Will fix https://github.com/juttle/juttle-elastic-adapter/issues/108
@demmer @VladVega

### DIFF
--- a/lib/runtime/procs/read.js
+++ b/lib/runtime/procs/read.js
@@ -126,6 +126,12 @@ class Read extends source {
             this.logger.debug('scheduling first live read at', firstRead.toJSON());
             this.program.scheduler.schedule(firstRead.unixms(), _.bind(this.run, this));
         }
+
+        if (this.params.optimization_info._reduce_every && this.to.isEnd()) {
+            var every = JuttleMoment.duration(this.params.optimization_info._reduce_every);
+            this.readTo = this.readFrom.quantize(every).add(every);
+            this.readEvery = every;
+        }
     }
 
     // Kick off a read iteration from the adapter.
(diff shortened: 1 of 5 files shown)


VladVega:
@davidvgalbraith would there need to be additional code in the adapters for this to be used?

lib/runtime/procs/read.js
```diff
@@ -126,6 +126,12 @@ class Read extends source {
             this.logger.debug('scheduling first live read at', firstRead.toJSON());
             this.program.scheduler.schedule(firstRead.unixms(), _.bind(this.run, this));
         }
+
+        if (this.params.optimization_info.reduce_every && this.to.isEnd()) {
```
demmer: Previously, we left it totally up to the adapter to put what it wanted into the `optimization_info` -- this would be a departure from that design since read is now poking into the contents.

What if instead we changed the contract so that in the case of optimized `reduce -every`, the adapter returns the `every` interval as part of the `defaultTimeOptions()`. Then we would change the logic at L108 above so that it sets the initial `readTo` to be `now - lag` quantized on `every` (which seems like sensible behavior anyway) which seems to me like it would accomplish the same goal but without needing to reach into the optimization info.

Somewhat related, I think it would also be sensible to disallow the combination of `read -every :x: | reduce -every :y:` in the case where the reduce is optimized.

davidvgalbraith:
Updated, with the latest change it'll magically work without changes to the adapters.

lib/runtime/procs/read.js
```diff
@@ -126,6 +126,12 @@ class Read extends source {
             this.logger.debug('scheduling first live read at', firstRead.toJSON());
             this.program.scheduler.schedule(firstRead.unixms(), _.bind(this.run, this));
         }
+
+        if (this.params.optimization_info._reduce_every && this.to.isEnd()) {
```
demmer: I'd still like a check that throws an error if the user explicitly passed an `-every` option to read.

davidvgalbraith:
Added such a check.

demmer:
looks good. +1

PR #630, opened 2016-03-21
Make the “.” operator work properly in lvalues

dmajda:
This PR adds tests that specify behavior of the `.` operator in lvalues and adjusts the implementation to match. Specifically, it extends the Juttle grammar and removes incorrect checks in the semantic pass.

Note that member assignment in `put`/`reduce` still isn't allowed. Making it work requires more substantial changes which will be done later.

Part of work on #419.


dmajda:
@bkutil

bkutil:
+1

PR #631, opened 2016-03-21
Implement nested field assignment

dmajda:
Implement nested field assignment in `put`/`reduce`. This means code like this is completely valid now and works (assuming the `a` field contains appropriate Juttle value):

```juttle
put a[0].b[0] = 1
```

Final part for work on #419.


dmajda:
@bkutil

bkutil:
+1

PR #633, opened 2016-03-21
view.js: emit event payload in a consistently named property

go-oleg:
For 'emit:mark', 'emit:tick', and 'emit:points' events, always put the payload for (e.g. mark, time, points) into a property called 'payload'.

The change in https://github.com/juttle/juttle/pull/567 requires making changes in a few places that receive marks and pass them on to somewhere else (https://github.com/juttle/juttle-service/issues/71 and https://github.com/juttle/juttle-client-library/issues/80). Since these places just pass-through the core payload of the event and only transform the surrounding event metadata, using a generic 'payload' allows us to future-proof against any changes we might make to the 'payload' contents. These intermediate steps shouldn't need to be changed every time - only the actual end-subscribers to the events (generally the views).

We can use 'data' instead of 'payload' if you guys think thats a more appropriate name.

@demmer @dmajda @mstemm

### DIFF
--- a/lib/runtime/procs/view.js
+++ b/lib/runtime/procs/view.js
@@ -27,17 +27,17 @@ class view extends sink {
     // include proc names including the error/warning.
 
     mark(mark) {
-        this.program.events.emit('view:mark', {channel: this.channel, mark: mark});
+        this.program.events.emit('view:mark', {channel: this.channel, data: mark});
     }
     tick(time) {
-        this.program.events.emit('view:tick', {channel: this.channel, time: time});
+        this.program.events.emit('view:tick', {channel: this.channel, data: time});
     }
     eof() {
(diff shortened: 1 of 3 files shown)


dmajda:
+1

I like `data` more than `payload`, but can live with both.

dmajda:
Just noticed in the commit message:

> For 'emit:mark', 'emit:tick', and 'emit:points' events

I think you wanted s/emit/view/.

demmer:
FWIW I'd prefer `data` as well but am ok with either. 

-m

go-oleg:
Changed to `data` (and changed the formal arg name in the consumers to `event` instead of `data` to avoid `data.data`).

And thx for catching the commit msg typo.

PR #639, opened 2016-03-21
views_sourceinfo: get info from 'emit' nodes

go-oleg:
Previously we were grabbing from/last/to info from 'read' nodes. Grab the info from 'emit' as well.

Removed `_jut_time_bounds` verification from some tests because it felt a bit redundant.

@dmajda @bkutil @demmer

### DIFF
--- a/test/compiler/flowgraph/views_sourceinfo.spec.js
+++ b/test/compiler/flowgraph/views_sourceinfo.spec.js
@@ -45,8 +45,13 @@ describe('Views get info on source time bounds', function() {
     var last = JuttleMoment.duration(1, 'hour');
 
     test('read stochastic -source "cdn" -from :' + from.valueOf() + ': | view view ', [[{from: from, to: null, last: null}]]);
+    test('emit -from :' + from.valueOf() + ': | view view ', [[{from: from, to: null, last: null}]]);
+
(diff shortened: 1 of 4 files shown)


test/compiler/flowgraph/views_sourceinfo.spec.js
```diff
@@ -54,4 +59,6 @@ describe('Views get info on source time bounds', function() {
     test('(read stochastic -source "cdn" -from :' + from.valueOf() + ':; read stochastic -last :' + last.valueOf() + ': -source "cdn" ) | view view ',
          [[{from: from, to: null, last: null}, {from: null, to: null, last: last}]]);
 
+    test('(read stochastic -source "cdn" -last :' + last.valueOf() + ':; emit -last :' + last.valueOf() + ': ) | view view ',
+         [[{from: null, to: null, last: last}, {from: null, to: null, last: last}]]);
```
dmajda: I’d say this test is slightly redundant and I’d remove it. These test seem to exercise behavior with different graph topologies (not specific sources) and we already established `emit` works in the tests above.

dmajda:
> Removed `_jut_time_bounds` verification from some tests because it felt a bit redundant.

Could you split this into a separate commit?

dmajda:
+1

PR #641, opened 2016-03-21
Standard adapter config

VladVega:
Implemented config standardization talked about here: https://github.com/juttle/juttle/pull/555

This will not be merged before this next release. We'll wait until the next one because we'll need to bump the adapter API anyway for the next one anyway.


lib/runtime/adapters.js
```diff
@@ -58,24 +58,49 @@ function adapterModulePath(type, options) {
...
+        }
+        instances = conf.instances;
+    } else {
+        instances = _.isArray(conf) ? conf : [_.omit(conf, 'path')];
+    }
+    
+    if (instances.length > 1) {
+        _.each(instances, (instance) => {
+            if (!instance.id) {
+                throw new Error(`One or more configuation instances for adapter ${type} is missing an id.`);
```
demmer: s/configuation/configuration/

lib/runtime/adapters.js
```diff
@@ -58,24 +58,49 @@ function adapterModulePath(type, options) {
     return modulePath;
 }
 
+function getInstances(type, conf) {
+    var instances;
+    
+    if (conf.instances) {
+        if (!_.isArray(conf.instances)) {
+            throw new Error(`instances key in configuation for adapter ${type} must be an array.`);
```
demmer: s/configuation/configuration/

test/runtime/adapter-config.spec.js
```diff
@@ -0,0 +1,139 @@
...
+var check_juttle = juttle_test_utils.check_juttle;
+var path = require('path');
+var expect = require('chai').expect;
+var _ = require('underscore');
+var adapters = require('../../lib/runtime/adapters');
+
+
+function configure(name, conf) {
+    conf = _.clone(conf);
+    conf.path = path.resolve(__dirname, '../runtime/test-adapter-conf');
```
demmer: For completeness, I think it would be good to test that the adapter loads as expected when the path isn't supplied, and I find it kinda hackish that we test the case of assigning a `path` property to an array since that just seems dirty.

How about if we change the adapters implementation somewhat so that the internal `adapterModulePath` could be overridden by the unit test.

Then you could test the various configuration options here both with the implicit path (derived from your overridden module path handler) and with the explicit path in the cases where the configuration is an object.

test/runtime/adapter-config.spec.js
```diff
@@ -0,0 +1,139 @@
...
+            program: `read ${name}`
+        })
+        .then(function(result) {
+            throw new Error('unexpected success');
+        })
+        .catch(function(err) {
+            expect(err.message).equal(`instances key in configuation for adapter ${name} must be an array.`);
+        });
+    });
+});
```
demmer: I think it'd also be good to have some tests to verify what happens when `path` points to a nonexistent directory or an otherwise invalid path, coupled with the change above.

docs/adapters/configuration.md
```diff
@@ -0,0 +1,110 @@
+## Goal
```
demmer: As a design discussion this is good but I think we should update docs/reference/cli.md to reflect what we ended up settling on and not include this doc in the final set of changes.

test/runtime/adapter-config.spec.js
VladVega: added.

test/runtime/adapter-config.spec.js
VladVega: I thought about how to do this and implemented it. I guess another way would be to use `sinon`. Can you take a look and see if it's an acceptable solution?

docs/adapters/configuration.md
VladVega: Cool - I moved the design doc to the wiki: https://github.com/juttle/juttle/wiki/adapter-config-design

VladVega: Also I put in some information in `docs/reference/cli.md`

test/runtime/adapter-config.spec.js
VladVega: Also I was thinking one better would be just to have:
```
var AdapterHelper = {
  register: function,
  configure:function
};
```
and then export that. That's better right?

docs/reference/cli.md
```diff
@@ -245,6 +245,49 @@ By default, the runtime will try to load the adapter from an npm module matching
...
+                    "address": "localhost",
+                    "port": 9200
+                },
+                {
+                    "id": "prod",
+                    "address": "prodHost",
+                    "port": 9200
+                }
+            ],
+            "path": "path/to/elastic/module"
```
demmer: Maybe mention that "path" is optional.

VladVega:
@demmer Putting this here in case it gets lost in the outdated diff:
I implemented a strategy for the internal `adapterModulePath` to be overridden by the unit test.
I guess another way this could be done is with `sinon`?

Also I was thinking maybe the `adapters.js` file should look like this instead :
```
var AdapterHelper = {
  register: function,
  configure:function
};
```
That's better right?

PR #644, opened 2016-03-22
Fold file resolver into test resolver.

mstemm:
Modify the resolver used by juttle-test-utils, which takes a
collection of modules and falls back to stdlib otherwise, to use the
file resolver as a backup instead of looking under the stdlib
directory itself.

This allows import statements within unit tests to use short names for
stdlib modules if desired. It also allows other test programs (like
the aws unit tests) to also utilize adapter modules.

@rlgomes and @dmajda this is the change we discussed at https://github.com/juttle/juttle/pull/591#issuecomment-198086588.

### DIFF
--- a/test/runtime/specs/juttle-spec/stdlib/predict.spec.md
+++ b/test/runtime/specs/juttle-spec/stdlib/predict.spec.md
@@ -2,7 +2,7 @@
 
 ## predict with features disabled returns previous value
 ### Juttle
-    import "predict.juttle" as predict;
+    import "predict" as predict;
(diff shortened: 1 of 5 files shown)


rlgomes:
+1 but to prove it works why not modify these guys https://github.com/juttle/juttle/tree/master/test/runtime/specs/juttle-spec/stdlib to simply import without the `.juttle` extension ?

dmajda:
+1

mstemm:
Modified the stdlib unit tests to drop the .juttle extension. The tests still passed, so going ahead and merging.

PR #647, opened 2016-03-22
SemanticPass: Fix reducer args coercion

### LINKED ISSUE #646: Reducer args are coerced incorrectly
Variables passed as reducer args are coerced into fields unless there is a const or var in scope with the same name. All other entities such as functions or reducers are ignored. This is inconsistent with by-lists and sort-lists where coercion in such cases leads to a `CANNOT-USE-AS-VARIABLE` error:

```
$ bin/juttle -e 'function f() { }; emit -from :0: -limit 1 | reduce min(f)'
┌────────────┐
│ min        │
├────────────┤
│ Infinity   │
└────────────┘
$ bin/juttle -e 'function f() { }; emit -from :0: -limit 1 | reduce by f'
<input>:1:55: 
   1:function f() { }; emit -from :0: -limit 1 | reduce by f
                                                           ^
Error: Cannot use a function as a variable (CANNOT-USE-AS-VARIABLE)

```

Coercion of reducer args should work exactly the same as in by-lists and sort-lists.

dmajda:
Before this PR, variables passed as reducer args were coerced into fields unless there was a const or var in scope with the same name. All other entities such as functions or reducers were ignored. This was inconsistent with by-lists and sort-lists where coercion in such cases lead to a `CANNOT-USE-AS-VARIABLE` error.

This PR makes coercion of reducer args work exactly the same as in by-lists and sort-lists (by utilizing the same code path).

Fixes #646.

### DIFF
--- a/lib/compiler/semantic.js
+++ b/lib/compiler/semantic.js
@@ -988,17 +988,10 @@ class SemanticPass {
     sa_reducer_arg(arg) {
         // coerce a naked name to a string (ie a field name) if it
         // isn't visible as a variable.
-        if (arg.type === 'Variable'
-            && this.scope.lookup_variable(arg.name) === undefined) {
-            //XXX this shouldn't have to be special cased and we should
-            // just be passing down the opts argument and letting this
-            // happen in sa_Variable XXX that is, unless we want only
-            // single variable arguments to be coerced...? and other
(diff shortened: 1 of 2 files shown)


dmajda:
@bkutil

bkutil:
+1

PR #653, opened 2016-03-23
add top and bottom subs to select module in stdlib

### LINKED ISSUE #388: Implement “top” and “bottom” subs
Add `top` and `bottom` subs to stdlib. Something like:

```juttle
top -n 100 -by downloads 

sub top(n, by) { 
    sort *by | head n 
    // gets optimized to sort -limit n -no_warn_on_drop *by 
} 

bottom -n 100 -by downloads 

sub bottom(n, by) { 
    sort *by | tail n 
    // gets optimized to sort -limit n -no_warn_on_drop -drop_head *by 
} 
```

(This issue was tracked as PROD-9042 internally.)

rlgomes:
fixes #388

added `top` & `bottom` subs along with necessary tests and documentation

### DIFF
--- a/docs/stdlib/select.md
+++ b/docs/stdlib/select.md
@@ -66,3 +66,39 @@ Options      |         Description           | Required?
 
 Only the point with the exact median value for the `field` name
 provided will be forwarded.
+
+## top - sub
+
+```
+... | select.top -n number -by field | ...
+```
+
+Options  |                              Description                              | Required?
+-------- | --------------------------------------------------------------------- | --------- :
+`n`      | how many of the points with the highest values of `by` to keep        | Yes
+`by`     | name of the field to sort by before figuring out the top `n` to keep  | Yes
+`limit`  | the limit value to pass to sort if you happen to exceed the default limit | No, default: see [sort](../processors/sort.md) documentation
+
+Sort the inbound stream by the `by` field and return the `n` points that have
+highest value of the `by` field.
+
+**Note** Since `top` uses `sort` to sort by the `by` field we drop the `time` field
+from the individual points.
+
+## bottom - sub
+
+```
+... | select.bottom -n number -by field | ...
+```
+
+Options  |                                Description                                | Required?
+-------- | ------------------------------------------------------------------------- | --------- :
+`n`      | how many of the points with the lowest values of `by` to keep             | Yes
+`by`     | name of the field to sort by before figuring out the bottom `n` to keep   | Yes
+`limit`  | the limit value to pass to sort if you happen to exceed the default limit | No, default: see [sort](../processors/sort.md) documentation
+
+Sort the inbound stream by the `by` field and return the `n` points that have
+lowest value of the `by` field.
+
+**Note** Since `top` uses `sort` to sort by the `by` field we drop the `time` field
+from the individual points.
--- a/lib/stdlib/select.juttle
+++ b/lib/stdlib/select.juttle
@@ -59,3 +59,30 @@ export sub percentile(field, p, by=[]) {
 export sub median(field, by=[]) {
     percentile -field field -p 0.5 -by by
 }
+
+//
+// forward the top N points ordered by the `by` field provided
+//
+// Parameters:
+//
+//   * n: number of points to forward
+//   * by: field to sort by
+//   * limit: the limit to pass to sort
+//
+export sub top(n, by, limit=null) {
+    sort -limit limit by -desc | head n
+} 
+
+//
+// forward the bottom N points ordered by the `by` field provided
+//
+// Parameters:
+//
+//   * n: number of points to forward
+//   * by: field to sort by
+//   * limit: the limit to pass to sort
+//
+export sub bottom(n, by, limit=null) {
+    sort -limit limit by | head n
+} 
+
--- a/test/runtime/specs/juttle-spec/stdlib/select.spec.md
+++ b/test/runtime/specs/juttle-spec/stdlib/select.spec.md
@@ -179,3 +179,133 @@
     {parity:1, x: 15}
     {parity:0, x: 36}
     {parity:1, x: 35}
+
+## top selects the 5 largest values
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 10 -from Date.new(0)
+    | put x = 10-count()
+    | select.top -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 9}
+    {x: 8}
+    {x: 7}
+    {x: 6}
+    {x: 5}
+
+## bottom selects the 5 smallest values
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 10 -from Date.new(0)
+    | put x = 10-count()
+    | select.bottom -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 0}
+    {x: 1}
+    {x: 2}
+    {x: 3}
+    {x: 4}
+
+## top selects the largest value per batch
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 9 -from Date.new(0)
+    | put x = count()
+    | batch -every :3s:
+    | select.top -n 1 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    { x: 3 }
+    { x: 6 }
+    { x: 9 }
+
+## bottom selects the smallest value per batch
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 9 -from Date.new(0)
+    | put x = count()
+    | batch -every :3s:
+    | select.bottom -n 1 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    { x: 1 }
+    { x: 4 }
+    { x: 7 }
+
+## top selects the 5 largest values from 3 element stream
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 3 -from Date.new(0)
+    | put x = 3-count()
+    | select.top -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 2}
+    {x: 1}
+    {x: 0}
+
+## bottom selects the 5 smallest values from 3 element stream
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 3 -from Date.new(0)
+    | put x = 3-count()
+    | select.bottom -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 0}
+    {x: 1}
+    {x: 2}
+
+## top can correctly modify the underlying sort limit
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 6 -from Date.new(0)
+    | put x = 6-count()
+    | select.top -n 5 -by 'x' -limit 5
+    | keep x
+    | view result
+
+### Warnings
+ 
+  * sort limit of 5 exceeded, dropping points
+
+### Output
+    {x: 5}
+    {x: 4}
+    {x: 3}
+    {x: 2}
+    {x: 1}
+
+## bottom can correctly modify the underlying sort limit
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 6 -from Date.new(0)
+    | put x = 6-count()
+    | select.bottom -n 5 -by 'x' -limit 5
+    | keep x
+    | view result
+
+### Warnings
+ 
+  * sort limit of 5 exceeded, dropping points
+
+### Output
+    {x: 1}
+    {x: 2}
+    {x: 3}
+    {x: 4}
+    {x: 5}


rlgomes:
@dmehra @dmajda quick and easy although I switched the `tail`/`head` usage from the original issue as top should be the highest values and bottom should be the lowest values (unless I completely mixed up my tops and bottoms then I can easily switch back).

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+```
+
+Options  |                              Description                              | Required?
+-------- | --------------------------------------------------------------------- | --------- :
+`n`      | how many of the fields with the highest values of `by` to keep        | Yes
+`by`     | name of the field to sort by before figuring out the top `n` to keep  | Yes
+
+Let the `n` points that have the highest value of `by` pass through. 
+
+**Note** since we use sort to sort by the `by` field we end up not emitting
```
dmehra: Double negation here? We emit points without time field or we don't do so? Maybe just say that we drop the time field from the forwarded points (regrettably so...this sort-drops-time business has tripped me up repeatedly, especially when using subs where I forget sort is in there, and then nothing works downstream. @dmajda can we revisit this?)

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+```
+
+Options  |                                Description                              | Required?
+-------- | ----------------------------------------------------------------------- | --------- :
+`n`      | how many of the fields with the lowest values of `by` to keep           | Yes
+`by`     | name of the field to sort by before figuring out the bottom `n` to keep | Yes
+
+Let the `n` points that have the lowest value of `by` pass through. 
+
+**Note** since we use sort to sort by the `by` field we end up not emitting
```
dmehra: Same as above, emitting points without time field, not `not emitting` I think

lib/stdlib/select.juttle
```diff
@@ -59,3 +59,28 @@ export sub percentile(field, p, by=[]) {
...
     percentile -field field -p 0.5 -by by
 }
+
+//
+// forward the top N points ordered by the `by` field provided
+//
+// Parameters:
+//
+//   * n: number of points to forward
+//   * by: (opt) field to sort by
```
dmehra: This comment implies `by` is optional but it's not, is it? I don't see a default

lib/stdlib/select.juttle
```diff
@@ -59,3 +59,28 @@ export sub percentile(field, p, by=[]) {
...
+//
+export sub top(n, by) { 
+    sort by | tail n
```
dmehra: This would be clearer with explicit `sort -asc`, which I know is default behavior, but if we spell it out then having `tail` not `head` makes sense

docs/stdlib/select.md
dmajda: @dmehra Revisiting the decision to drop time in `sort` in non-batched streams means either:

  1. Revisiting the requirement for time to be non-decreasing in points.
  2. Changing the semantics of `sort` to take times from points, sort them, and assign them back one by one to points sorted by specified criteria.

I’m not opposed to either, it just needs thinking through. Obviously, 2 is more localized change, but the association between points and their times would be lost (which is not intuitive and may prove hurt some use cases). It also wouldn’t make sense to do it if we want to do 1 at tome point. As for 1, Henri wanted to do it for quite some time, we just need to think through all the consequences.

Probably worth filing an issue. Also, summoning @bkutil.

test/runtime/specs/juttle-spec/stdlib/select.spec.md
```diff
@@ -179,3 +179,63 @@
...
+    | select.top -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 5}
+    {x: 6}
+    {x: 7}
+    {x: 8}
+    {x: 9}
```
dmajda: Interesting, I’d expect the sequence to be reversed. My mental model for `top` is “sort so that the biggest value is first and then cut-off *n* items”. Think hi-score in games, various [rating tables](http://www.goratings.org/), etc.

test/runtime/specs/juttle-spec/stdlib/select.spec.md
```diff
@@ -179,3 +179,63 @@
...
+    | select.bottom -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 0}
+    {x: 1}
+    {x: 2}
+    {x: 3}
+    {x: 4}
```
dmajda: Similarly, I’d expect this sequence to be reversed.

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+
+## top - sub
+
+```
+... | select.top -n number -by field | ...
+```
+
+Options  |                              Description                              | Required?
+-------- | --------------------------------------------------------------------- | --------- :
+`n`      | how many of the fields with the highest values of `by` to keep        | Yes
```
dmajda: We are keeping points, not fields I think.

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+`by`     | name of the field to sort by before figuring out the top `n` to keep  | Yes
+
+Let the `n` points that have the highest value of `by` pass through.
```
dmajda: This should mention sorting. As it is, the text implies not touching the points at all beside filtering.

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+
+## bottom - sub
+
+```
+... | select.bottom -n number -by field | ...
+```
+
+Options  |                                Description                              | Required?
+-------- | ----------------------------------------------------------------------- | --------- :
+`n`      | how many of the fields with the lowest values of `by` to keep           | Yes
```
dmajda: Same as above (keeping points, not fields).

docs/stdlib/select.md
```diff
@@ -66,3 +66,35 @@ Options      |         Description           | Required?
...
+`by`     | name of the field to sort by before figuring out the bottom `n` to keep | Yes
+
+Let the `n` points that have the lowest value of `by` pass through.
```
dmajda: Same as above (implies no sorting).

test/runtime/specs/juttle-spec/stdlib/select.spec.md
```diff
@@ -179,3 +179,63 @@
...
+
+## top selects the 5 largest values from an incomplete stream
```
dmajda: What do you mean by incomplete stream?

test/runtime/specs/juttle-spec/stdlib/select.spec.md
```diff
@@ -179,3 +179,63 @@
...
+    | select.top -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
+    {x: 0}
+    {x: 1}
+    {x: 2}
+
+## bottom selects the 5 smallest values from an incomplete stream
```
dmajda: Same as above (incomplete stream?).

rlgomes:
@dmehra @dmajda addressed all comments.

test/runtime/specs/juttle-spec/stdlib/select.spec.md
```diff
@@ -179,3 +179,63 @@
...
+## bottom selects the 5 smallest values from 3 element stream
+### Juttle
+    import "select.juttle" as select;
+    emit -limit 3 -from Date.new(0)
+    | put x = 3-count()
+    | select.bottom -n 5 -by 'x'
+    | keep x
+    | view result
+
+### Output
```
dmehra: Intuitively I want to go the other way for `bottom`, so smallest value is first, otherwise I think I'd just have to `select.bottom | sort -asc` every time. Are we going by standard expectations as in other libraries?

dmehra:
Does `sort` drop time if executed in batch context? Or does it assign end-of-batch time? I'll file a separate issue for time handling in sort, this question is specific to this PR so we correctly document what happens if used in `batch x | select.top` manner.

test/runtime/specs/juttle-spec/stdlib/select.spec.md
rlgomes: @dmehra find me a library that has top/bottom to mimic. 

As for the specific behavior I'd say yes `give me the top 10 most downloaded` ranks them from highest to lowest (currenty behavior) `give me the bottom 10 download speed` (ie slowest downlaods) would rank them from smallest to largest. I'll make the change now

rlgomes:
@dmehra sort always drops time why or how would it even care that the previous node was a batch ?

rlgomes:
@dmehra I stand corrected ```rlgomes@x230> ./bin/juttle -e "emit -limit 10 -from :0: | put value=count() | sort value -desc"
┌──────────┐
│ value    │
├──────────┤
│ 10       │
├──────────┤
│ 9        │
├──────────┤
│ 8        │
├──────────┤
│ 7        │
├──────────┤
│ 6        │
├──────────┤
│ 5        │
├──────────┤
│ 4        │
├──────────┤
│ 3        │
├──────────┤
│ 2        │
├──────────┤
│ 1        │
└──────────┘
2016-03-23 09:10:51 $?=0 pwd=/home/rlgomes/jut/workspace/juttle branch=master duration=.519s                                                                            
rlgomes@x230> ./bin/juttle -e "emit -limit 10 -from :0: | put value=count() | batch -every :10s: | sort value -desc"
┌──────────────────────────┬──────────┐
│ time                     │ value    │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 10       │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 9        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 8        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 7        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 6        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 5        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 4        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 3        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 2        │
├──────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z │ 1        │
└──────────────────────────┴──────────┘
``` the timestamp is magically propagated

**Comment by dmehra** [comment]
One of @dmajda's comments made me think this would be the case. I think we got competing guidelines there: sort drops time, but batch assigns the batch interval time, and looks like batch wins. This makes it a bit trickier to describe in the docs what will happen as the result of using `select.top`(or bottom) because it's different depending on batched/unbatched context.

**Comment by dmehra** [comment]
Also, perhaps worth a testcase where we do `emit ... | batch N | select.top`. Not sure if this sort of thing is considered covered by other unit tests, but here since the resulting behavior differs, at least the tests would show how.

**Comment by dmehra** [comment]
Oy. Looked at sort doc, noticed
```
Note: Sorting is limited to 10,000 data points by default. To sort more points, use the -limit flag to set a higher value.
```
That will be a problem, as the sub doesn't take a limit parameter, making it impossible to get around the issue if you run into it. You will see a warning (N of them): `Warning: sort limit exceeded, dropping points` but then what - go fork the sub? Seems like we must support limit as an optional parameter.
As a side note, `10,000` in the doc is a lie, it defaults to 100,000, i'll go fix that.

**Comment by rlgomes** [comment]
@dmehra added -limit, documented it and also added tests for batched data.

**Comment by demmer** [comment]
> I think we got competing guidelines there: sort drops time, but batch assigns the batch interval time, and looks like batch wins. 

This isn't entirely accurate. Sort drops the time from the individual points in all contexts. However, when run in batched context, the emitted points contain the batch timestamp -- currently epsilon before end-of-batch... but for consistency I think this maybe be start-of-batch + interval once #123 lands? Not so sure about this one...

For example:

```
juttle> emit -limit 10 -from :0: | put count=count() | batch :2s: | sort count -desc
┌────────────────────────────────────┬──────────┐
│ time                               │ count    │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:02.000Z           │ 2        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:02.000Z           │ 1        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:04.000Z           │ 4        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:04.000Z           │ 3        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:06.000Z           │ 6        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:06.000Z           │ 5        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:08.000Z           │ 8        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:08.000Z           │ 7        │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z           │ 10       │
├────────────────────────────────────┼──────────┤
│ 1970-01-01T00:00:10.000Z           │ 9        │
└────────────────────────────────────┴──────────┘```

docs/stdlib/select.md
```diff
@@ -66,3 +66,39 @@ Options      |         Description           | Required?
...
+-------- | --------------------------------------------------------------------- | --------- :
+`n`      | how many of the points with the highest values of `by` to keep        | Yes
+`by`     | name of the field to sort by before figuring out the top `n` to keep  | Yes
+`limit`  | the limit value to pass to sort if you happen to exceed the default limit | No, default: see [sort](../processors/sort.md) documentation
+
+Sort the inbound stream by the `by` field and return the `n` points that have
+highest value of the `by` field.
+
+**Note** since we use `sort` to sort by the `by` field we drop the `time` field
+from the stream output.
```
dmehra: let's say "drop the `time` field from individual points" i guess. No matter what we say here, the user is likely to be surprised when running this in batched vs unbatched context. A more detailed explanation probably belongs in sort's doc article.

lib/stdlib/select.juttle
```diff
@@ -59,3 +59,30 @@ export sub percentile(field, p, by=[]) {
...
+// forward the bottom N points ordered by the `by` field provided
+//
+// Parameters:
+//
+//   * n: number of points to forward
+//   * by: field to sort by
+//   * limit: the limit to pass to sort
+//
+export sub bottom(n, by, limit=null) {
+    sort -limit limit by | head n
```
dmehra: i'd still do `sort -limit limit by -asc` here, for clarity of how it's sorted.

rlgomes: meant to say it earlier what is this `-asc` you speak of ? http://juttle.github.io/juttle/processors/sort/#sort as its not documented anywhere.

test/runtime/specs/juttle-spec/stdlib/select.spec.md
dmehra: So spark has top() which is commonly used to get ordering like 9,8,7,6,5 above (although it has a way to specify ordering, too). I haven't found anything with bottom() - no equivalent in spark or pandas, sql is build-your-own with `order by`, so i think it's what we can agree on. I like 0,1,2,3,4 order for bottom() in practice, but can also see the viewpoint of "we order the whole list descending, and then top() is the top slice of that, and bottom() is the bottom slice of that, without further reordering".

lib/stdlib/select.juttle
dmehra: It sure isn't documented... but works:
```
juttle> emit -from :0: -limit 100 | put value = count() | sort value -desc | head 1
┌──────────┐
│ value    │
├──────────┤
│ 100      │
└──────────┘
juttle> emit -from :0: -limit 100 | put value = count() | sort value -asc | head 1
┌──────────┐
│ value    │
├──────────┤
│ 1        │
└──────────┘
```

rlgomes: Humm yeah its not quite like that ```
rlgomes@x230> ./bin/juttle -e "emit -from :0: -limit 3 | put value = count() | sort value  -bananas"
┌──────────┐
│ value    │
├──────────┤
│ 1        │
├──────────┤
│ 2        │
├──────────┤
│ 3        │
└──────────┘
```

Comment (dmehra) [inline lib/stdlib/select.juttle]: lol `-asc` is the option we should have had. Carry on with the PR...

File: test/runtime/specs/juttle-spec/stdlib/select.spec.md (hunk 13 above)
Comment (dmajda) [inline test/runtime/specs/juttle-spec/stdlib/select.spec.md]: On a second thought I think I like your intuition about `bottom` more than what I originally proposed, @dmehra. It’s like what `uniq -c | sort -nr | head ...` in Unix gives you (this is something I do all the time).

File: docs/stdlib/select.md [hunk 16]
```diff
@@ -66,3 +66,39 @@ Options      |         Description           | Required?
...
+-------- | --------------------------------------------------------------------- | --------- :
+`n`      | how many of the points with the highest values of `by` to keep        | Yes
+`by`     | name of the field to sort by before figuring out the top `n` to keep  | Yes
+`limit`  | the limit value to pass to sort if you happen to exceed the default limit | No, default: see [sort](../processors/sort.md) documentation
+
+Sort the inbound stream by the `by` field and return the `n` points that have
+highest value of the `by` field.
+
+**Note** since we use `sort` to sort by the `by` field we drop the `time` field
+from the individual points.
```
dmajda: Do we use “we” elsewhere in the docs? I think this should be more like “Since `top` uses `sort` to sort by the `by` field it drops the `time` field from the individual points.”

docs/stdlib/select.md
```diff
@@ -66,3 +66,39 @@ Options      |         Description           | Required?
...
+-------- | ------------------------------------------------------------------------- | --------- :
+`n`      | how many of the points with the lowest values of `by` to keep             | Yes
+`by`     | name of the field to sort by before figuring out the bottom `n` to keep   | Yes
+`limit`  | the limit value to pass to sort if you happen to exceed the default limit | No, default: see [sort](../processors/sort.md) documentation
+
+Sort the inbound stream by the `by` field and return the `n` points that have
+lowest value of the `by` field.
+
+**Note** since we use `sort` to sort by the `by` field we drop the `time` field
+from the individual points.
```
dmajda: Same here (“we”).

dmajda:
+1

PR #654, opened 2016-03-23
Bump adapter API version to 0.7.0

dmajda:
This is needed because #384 introduces an incompatible change.


dmajda:
Per https://github.com/juttle/juttle/pull/384#issuecomment-199910073, merging without review in order to get 0.7.0 out.

@demmer FYI

PR #655, opened 2016-03-23
Report sort limit correctly

dmehra:
Ideally, the warning you get when running a program with `sort` on large input, `Warning: sort limit exceeded, dropping points` would also include the value of the limit - asking @rlgomes to add that. Here, fixing the doc which was wrong.


dmajda:
+1

PR #657, opened 2016-03-23
Handle non juttle errors in read adapter.

mstemm:
@demmer this is what we were discussing this am.

### DIFF
--- a/test/adapters/file/read.spec.js
+++ b/test/adapters/file/read.spec.js
@@ -87,7 +87,10 @@ describe('read file adapter tests', function () {
                 return run_read_file_juttle('bogus', {format: format})
                 .then(function(result) {
                     expect(result.errors.length).equal(1);
-                    expect(result.errors[0]).match(/ENOENT/);
+                    expect(result.error_objs[0].name).to.equal('RuntimeError');
+                    expect(result.error_objs[0].code).to.equal('INTERNAL-ERROR');
+                    expect(result.error_objs[0].message).to.include('ENOENT: no such file or directory');
+                    expect(result.errors[0]).to.include('ENOENT: no such file or directory');
                 });
             });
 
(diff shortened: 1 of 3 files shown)


test/adapters/file/read.spec.js
```diff
@@ -87,7 +87,9 @@ describe('read file adapter tests', function () {
                 return run_read_file_juttle('bogus', {format: format})
                 .then(function(result) {
                     expect(result.errors.length).equal(1);
-                    expect(result.errors[0]).match(/ENOENT/);
+                    expect(result.error_objs[0].name).to.equal('RuntimeError');
+                    expect(result.error_objs[0].code).to.equal('INTERNAL-ERROR');
+                    expect(result.error_objs[0].message).to.include('ENOENT: no such file or directory');
```
demmer: Seems like you should also check that result.errors[0] includes the same message.

demmer:
+1

dmajda:
I don’t think this check/handler is a good idea. Exception wrapping should happen on the adapter level and any unhandled errors *should* blow up (because they are bugs).

dmajda:
Also, `git bisect` just told me that wrapping was done in the `file` adapter until 4cf87fa49769b3584e08e4ba444100dc5ae92f72. So I’d say we should just restore it there and revert this PR.

@demmer @mstemm

mstemm:
I think I'd still lean towards this approach. All errors from calls to read() were being caught and passed along via `this.trigger('error', ...)` independent of this change. This change simply ensures that the errors being emitted are of type JuttleError.

I do agree that errors from bugs should not be caught, but I think it may be difficult to ensure that every adapter catches all its exceptions properly and only lets through the ones related to bugs. If people start writing their own adapters, this may not even be possible for us to ensure.

PR #658, opened 2016-03-23
stdlib fields module

### LINKED ISSUE #387: Implement “tail” and “prior” reducers
Implement `tail` and `prior` reducers and put them into stdlib.

Description of `tail`:

```juttle
// Return an array of last N values of a field
//
// Parameters:
//
//   * field: ("value") field to tail
//   * N: (1) number of recent values to retain
```

Description of `prior`:

```juttle
// Return the Nth most recent value of a field
//
// Parameters:
//
//   * field: ("value") name of value field
//   * N: (1) lag
//   * undef: (null) value to return before N values have been seen
```

(This issue was tracked as PROD-9172 internally.)

rlgomes:
fixes #387

created fields.tail and fields.prior reducers

### DIFF
--- a/docs/stdlib/fields.md
+++ b/docs/stdlib/fields.md
@@ -19,6 +19,19 @@ Options    |                     Description                 | Required?
 
 The result is a field containing the N last values of the field specified.
 
+## head - reducer
+
+```
+... | reduce hosts = fields.head(hostname, 5) | ...
+```
+
+Options    |                     Description                 | Required?
+---------- | ----------------------------------------------- | --------- :
+`field`    | the field name to store the values of in a list | Yes
+`length`   | the number of initial values to retain          | No, default: `1`
+
+The result is a field containing the N first values of the field specified.
+
 ## prior - reducer
 
 ```
--- a/docs/stdlib/fields.md
+++ b/docs/stdlib/fields.md
@@ -1,0 +1,32 @@
+---
+title: Fields Module | Juttle Language Reference
+---
+
+# Fields
+
+Reducers to manipulate fields within your data.
+
+## tail - reducer
+
+```
+... | reduce hosts = fields.tail(hostname, 5) | ...
+```
+
+Options    |                     Description                 | Required?
+---------- | ----------------------------------------------- | --------- :
+`field`    | the field name to store the values of in a list | Yes
+`length`   | the number of recent values to retain           | No, default: `1`
+
+The result is a field containing the N last values of the field specified.
+
+## prior - reducer
+
+```
+... | reduce penultimate = fields.tail(hostname, 2) | ...
+```
+
+Options  |                       Description                   | Required?
+-------- | --------------------------------------------------- | --------- :
+`field`  | the field name to store the `nth` last value for    | Yes
+`nth`    | index (from the end) of the value to keep track of  | No, default: `1`
+`undef`  | value to return before nth value has been seen      | No, default: `null`
--- a/lib/stdlib/fields.juttle
+++ b/lib/stdlib/fields.juttle
@@ -1,0 +1,54 @@
+// Manipulate fields within a stream with these field reducers
+
+// Return an array of last N values of a field
+//
+// Parameters:
+//
+//   * field: ("value") field to tail
+//   * length: (1) number of recent values to retain
+export reducer tail(field, length=1) {
+    var data = [];
+
+    function update() {
+        if (*field != null) { 
+            Array.push(data, *field);
+
+            if (Array.length(data) > length) {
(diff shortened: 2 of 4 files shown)


rlgomes:
@dmehra @dmajda `tail` and `prior` reducers! also added `head` since its "as useful" as `tail`.

go-oleg:
Theres some similarities between `head` and `tail` and the built in `first` and `last` (the latter could be extended to support a second `n` parameter). The one major difference is the built in ones currently return a scalar whereas the ones introduced here return arrays. Thoughts on whether its possible to bring the two together?

Also, I vote for `head` and `tail` to be in the core language w/o requiring `import "fields" as fields";`.

docs/stdlib/fields.md
```diff
@@ -0,0 +1,45 @@
...
+---------- | ----------------------------------------------- | --------- :
+`field`    | the field name to store the values of in a list | Yes
+`length`   | the number of initial values to retain          | No, default: `1`
+
+The result is a field containing the N first values of the field specified.
+
+## prior - reducer
+
+```
+... | reduce penultimate = fields.tail(hostname, 2) | ...
```
go-oleg: `s/tail/prior/`

docs/stdlib/fields.md
```diff
@@ -0,0 +1,45 @@
...
+## tail - reducer
+
+```
+... | reduce hosts = fields.tail(hostname, 5) | ...
+```
+
+Options    |                     Description                 | Required?
+---------- | ----------------------------------------------- | --------- :
+`field`    | the field name to store the values of in a list | Yes
+`length`   | the number of recent values to retain           | No, default: `1`
```
go-oleg: I think `n` is a better name for this option.
- Does a better job matching '...containing the N last values...'.
- Thats what [R uses](https://stat.ethz.ch/R-manual/R-devel/library/utils/html/head.html) (didn't check other languages).

lib/stdlib/fields.juttle
```diff
@@ -0,0 +1,74 @@
...
+//
+// Parameters:
+//
+//   * field: ("value") field to tail
+//   * length: (1) number of recent values to retain
+export reducer tail(field, length=1) {
+    var data = [];
+
+    function update() {
+        if (*field != null) {
```
go-oleg: Don't have a strong preference, but my first guess would've been that `tail` _does not_ ignore `null` values. Fine with it either way.

rlgomes: Do we have a different way of detecting the field isn't in the data vs the field is but set to null ? Because the problem here is points that don't have the field desired shouldn't count in the reduction right ?

docs/stdlib/fields.md
rlgomes: Ok I can go back to N everywhere... have just hated 1 letter variable names.

rlgomes:
head/tail being similar to first/last is only for n=1 and then there's the issue of one returning the value vs the head/tail reducers returning lists. I'm just filling in these stdlib bugs without much thought on consolidating with existing reducers (unless its a complete dup of an existing one) so I'd let @dmajda @dmehra decide if we'd want something other than head/tail.

lib/stdlib/fields.juttle
go-oleg: No, we don't have a way to tell the two apart. To me, whether they should or shouldn't count is unclear...but if its clear to others I'm game.

rlgomes: I'll wait for @dmajda @dmehra to vote on this one as I find it confusing if you have a mix of points and specifically state `values=tail(fieldx, 5)` but end up with values equal to [null, null, null] at the end of execution and the fieldx never never had any nulls itself. The other idea is to make it an option the reducer to `skipNulls`. The fact we can't distinguish between not there and set to null is a bit concerning.

dmajda:
This is another issue I pulled mechanically from JIRA. A gist referenced from there mentions the use case of alerting for `tail` — one could give e.g. few last field values in the stream as a context (think last few log messages). Not sure what the relevance of this use case is today.

Do we have any other obvious use cases for this? If not, I’d be inclined to just close this and maybe implement at some later point if a real demand appears.

> Don't have a strong preference, but my first guess would've been that `tail` _does not_ ignore `null` values. Fine with it either way.

I think the answer to question whether or not ignore `null`s lies in the use case(s). But my instinct is the same as yours, so I’d do the simple thing (not ignore `null`s) and add this as an option later if we need it.

> Do we have a different way of detecting the field isn't in the data vs the field is but set to null ? 

No and it bothers me too. I’ll file an issue.

PR #660, opened 2016-03-24
Show stack for internal errors.

### LINKED ISSUE #564: Display stack trace for internal errors
Right now, CLI handles internal errors (those for which [`is_juttle_error`](https://github.com/juttle/juttle/blob/f5a4ad7d6c087dd67fc8b2fb13b3610417f0cc7f/bin/juttle#L209-L212) returns `false`) by just displaying an error message. Since these errors are really bugs, it would be nice to also display a stack trace (like Node.js does by default). This would help a lot with diagnostics and fixing.

Before implementing this, #563 should be fixed first (so we don’t get stack traces for Juttle errors without location info).

mstemm:
Instead of just showing the message for internal errors, show the full
stack. This also includes the error class and message, so it's
strictly more information than we were showing.

This fixes #564.

@dmajda @go-oleg @rlgomes

### DIFF
--- a/lib/cli/errors.js
+++ b/lib/cli/errors.js
@@ -82,7 +82,7 @@ var handle_error = function(e, context) {
             console.error(error_string(e));
         }
     } else {
-        console.error('Error: ' + e.message);
+        console.error(e.stack);
     }
 };
 


dmajda:
+1

mstemm:
I'll wait to merge this until juttle 0.7.1 is released just to avoid creating a 0.7.0 branch for the 1 fix https://github.com/juttle/juttle/pull/657.

demmer:
Why not just include this in 0.7.1?

mstemm:
We can, this didn't seem important enough to merit inclusion and we already had the CHANGELOG written.

PR #661, opened 2016-03-24
remove duplicate implementation of stdev

rlgomes:
found upon inspection of the stdlib and since we have a built-in stdev
implementation this is completely unnecessary


rlgomes:
@demmer @dmajda

demmer:
+1

PR #663, opened 2016-03-25
add a view mode to the cli to show either all text or raw

demmer:
In order to help debug a juttle program in the CLI that contains
visual sinks, add a --view-mode option to the cli which can be either "text" in
which case all views are rendered as text), or "raw" in which case it just
prints out the raw underling data that would go to the views over the
websocket.

@go-oleg @mstemm

### DIFF
--- a/bin/juttle
+++ b/bin/juttle
@@ -47,7 +47,7 @@ var errors = require('../lib/errors');
 
 var modes = _.without(compiler.stageNames, 'eval').concat('run');
 function usage() {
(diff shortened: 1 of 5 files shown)


mstemm:
I'd suggest a simpler raw mode that just uses the text view for all sinks instead of this raw viewer. In some cases users might want the exact sequence of points, marks, ticks, eof, etc. but I think in some cases you just want the output.

demmer:
@mstemm Good idea -- I updated the PR to now support both `raw` and `text` output.

mstemm:
lgtm. Can you also update docs/reference/cli.md with the new option?

PR #664, opened 2016-03-25
cli: send logging output to stderr instead of console

### LINKED ISSUE #637: cli logger output should use stderr
In the CLI, we currently send both the view output and the logger output to stdout.

It would be better if we used stdout for the view output and stderr for the logger output (including both explicit logs and runtime warnings / errors) so that a user could redirect the logs to a different location.

demmer:
Use a custom log4js appender to send the logging output to stderr
instead of through console.log.

Fixes #637

### DIFF
--- a/lib/cli/log-setup.js
+++ b/lib/cli/log-setup.js
@@ -27,7 +27,7 @@ function logSetup() {
 
     log4js.configure({
         levels: levels,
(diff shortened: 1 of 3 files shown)


demmer:
@mstemm @go-oleg

mstemm:
lgtm.

